PR #14, opened 2020-03-29
Update README.md

Zrrrpy:
Resolves #10  

### Description
_case_ is a reserved word in Ruby so updated terminology for foster youth from _case_ to _casacase_

### Type of change

* Documentation update


compwron:
:)

PR #37, opened 2020-04-03
Policies minimally working

Thrillberg:
This is a small proof of concept of how to distinguish between admins
and volunteers by role. It needs tests.


Thrillberg:
Note: I did not include feature tests because I'm certain the front end will change a lot.

PR #50, opened 2020-04-04
upload mockup files for wiki

Zrrrpy:
PR for uploading mockup files so I can link to them in wiki

### Description
mockup files for wiki (they need to be hosted on git so they can be linked)
   
Include anything else we should know about. -->

### Type of change
* Documentation update


compwron:
Closing without merging because it's only a way to upload photos for wiki use

Issue #1, opened 2020-03-28
test issue


compwron:
test issue

compwron:
![admin_home](https://user-images.githubusercontent.com/578159/77831426-5f331f00-70ec-11ea-99a0-3f329fa6d021.jpg)
![volunteer_home_and_new_note](https://user-images.githubusercontent.com/578159/77831433-62c6a600-70ec-11ea-9f5a-7ab91a155e9b.jpg)
![admin_new_or_edit_volunteer](https://user-images.githubusercontent.com/578159/77831437-635f3c80-70ec-11ea-8b5f-7739f5c94345.jpg)
![admin_new_edit_supervisor](https://user-images.githubusercontent.com/578159/77831439-63f7d300-70ec-11ea-995e-9320b79ecdb3.jpg)

Thrillberg:
See the above in the wiki: https://github.com/rubyforgood/casa/wiki

compwron:
![IMG_20200328_152243](https://user-images.githubusercontent.com/578159/77836169-26f20780-7111-11ea-91d9-f479e09f3496.jpg)
![IMG_20200328_152241](https://user-images.githubusercontent.com/578159/77836172-2c4f5200-7111-11ea-8905-9af66b4f3054.jpg)
![IMG_20200328_152239](https://user-images.githubusercontent.com/578159/77836173-2d807f00-7111-11ea-9fce-32fab607eb11.jpg)
![IMG_20200328_152230](https://user-images.githubusercontent.com/578159/77836175-2e191580-7111-11ea-86ac-cf0ca8c809d2.jpg)
![IMG_20200328_152222](https://user-images.githubusercontent.com/578159/77836178-2f4a4280-7111-11ea-94b0-fd9b531afa47.jpg)
![overview](https://user-images.githubusercontent.com/578159/77836179-2fe2d900-7111-11ea-9a2b-e261d67b3963.jpg)

compwron:
I'm using this closed issue as a place to host images that I want to embed in the wiki :)

compwron:
![vert_logo-full_color](https://user-images.githubusercontent.com/578159/165240278-c2c0ac30-c86f-4b67-9da6-e6a5e4ab4c37.png)
this if for use in teh readme because where else to put it?

Issue #27, opened 2020-04-03
Volunteer view - case_contacts


Zrrrpy:
Related to #3 
- [x] button to create `new_case_contact`
  - button leads to `new_case_contact` path (#28 
- [x] infinite previous case contacts
  - should pull all information from previous case_contact including:
      - date, case_contact type, time duration
  - sorted by newest at the top

***_only_ `case_contacts` from active `casa_case_assignments` should display in the volunteer view.

Thrillberg:
I believe this is closed but happy to re-open if appropriate!

PR #51, opened 2020-04-04
Volunteer view case contacts 27

lisavogtsf:
Got through first checkbox of two for #27 

Need some clarification on case contacts? Or just keep it isolated.

Resolves #001 

### Type of change

* Breaking change (fix or feature that would cause existing functionality to not 
work as expected)
* This change requires a documentation update
* Documentation update

### DIFF
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -16,6 +16,14 @@ class CaseContact < ApplicationRecord
     other
   ].freeze
   enum contact_type: CONTACT_TYPES.zip(CONTACT_TYPES).to_h
+
+  def humanized_type
+    "#{contact_type.humanize.titleize}#{ " - " + other_type_text if use_other_type_text?}"
+  end
+
+  def use_other_type_text?
+    contact_type == 'other'
+  end
 end
 
 # == Schema Information
--- a/app/policies/case_contact_policy.rb
+++ b/app/policies/case_contact_policy.rb
@@ -38,4 +38,25 @@ class CaseContactPolicy # rubocop:todo Style/Documentation
   def destroy?
     _is_creator_or_casa_admin?
   end
+
+  class Scope
+    attr_reader :user, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      case @user.role
+      when 'casa_admin'
+        # scope.in_casa_administered_by(@user)
+        scope.all
+      when 'volunteer'
+        []
+      else
+        raise "unrecognized role"
+      end
+    end
+  end
 end
--- a/app/policies/case_contact_policy.rb
+++ b/app/policies/case_contact_policy.rb
@@ -53,7 +53,7 @@ class CaseContactPolicy # rubocop:todo Style/Documentation
         # scope.in_casa_administered_by(@user)
         scope.all
       when 'volunteer'
-        []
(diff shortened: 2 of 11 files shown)


datadaveshin:
Here is the updated version of the case contacts table
Screen Shot 2020-04-05 at 12 37 49 PM

app/models/case_contact.rb
```diff
@@ -16,6 +16,14 @@ class CaseContact < ApplicationRecord
     other
   ].freeze
   enum contact_type: CONTACT_TYPES.zip(CONTACT_TYPES).to_h
+
+  def humanized_type
+    "#{contact_type.humanize.titleize}#{ " - " + other_type_text if use_other_type_text?}"
```
Thrillberg, line 21: I love this! 😃

app/policies/case_contact_policy.rb
```diff
@@ -38,4 +38,25 @@ def edit?
...
+      @scope = scope
+    end
+
+    def resolve
+      case @user.role
+      when 'casa_admin'
+        # scope.in_casa_administered_by(@user)
+        scope.all
+      when 'volunteer'
+        scope.where(casa_case: CasaCase.actively_assigned_to(@user), creator: @user)
```
Thrillberg, line 56: This looks perfect to me!

app/views/dashboard/show.html.erb
```diff
@@ -56,4 +56,20 @@
...
     </table>
   </div>
+  <%= link_to "New Case Contact",  new_case_contact_path %>
+
+  <table class="case-contacts-table">
+    <tr>
+      <th>Date</th>
+      <th>Duration</th>
+      <th>Type</th>
+    </tr>
```
Thrillberg, line 66: I'm wondering if in this table we want to list the case number of the `CasaCase`. Judging by the `actively_assigned_to` scope, I feel like it's combining all `CaseContacts`, potentially mixing them between multiple `CasaCase`s. Or am I not reading that right? (quite probable!) I _think_ while it is in reality rare, there is the possibility that a `Volunteer` is assigned to multiple active `CasaCase`s at once.

spec/features/volunteer_adds_a_case_contact_spec.rb
```diff
@@ -6,7 +6,8 @@
 
     sign_in volunteer
 
-    visit new_case_contact_path
+    visit "/"
+    click_link_or_button "New Case Contact"
```
Thrillberg, line 10: Nice! Seems good to exercise the navigation here.

Thrillberg:
Let me know if y'all want me to merge this now and if so, if you'd rather resolve the merge conflicts or me. 😄

app/views/dashboard/show.html.erb
h-m-m: Oh, this is a good point

Thrillberg: Cool! Let me know when ready for final review!

h-m-m: How do you feel about moving this out into its own issue?

Thrillberg: I feel great about that! Shall I go ahead and rebase and merge this puppy? I can also write up the other issue.

app/models/case_contact.rb
h-m-m: I was thinking this would be a good candidate to refactor out into a view presenter object. It feels early to do that, though. I don't think there's enough complexity here to justify that yet

PR #55, opened 2020-04-05
Fix Case/Volunteer relationship in db

### LINKED ISSUE #53: Reconcile case_assignment and volunteer_id
We're not sure why volunteer_id needs to be in the join table, seems duplicative but there good be things we're missing!

Main issues this affects are: #27 and #32

Thrillberg:
Fixes #53

@h-m-m Here's my stab at the issue we discussed last night about the relationship between CasaCases and Volunteers. Tests are broken so I know something is wrong with my implementation. Do you mind taking a look and seeing if you can figure it out? I'm also happy to pair on it, if you want!

### DIFF
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -5,7 +5,7 @@ class CasaCase < ApplicationRecord
   has_many(
     :volunteers,
     through: :case_assignments,
-    source: :user,
+    source: :volunteer,
     class_name: "User"
   )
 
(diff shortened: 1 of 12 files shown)


Thrillberg:
Tests are now passing and I _think_ this is the setup we want. @h-m-m or @compwron can either of you confirm?

h-m-m:
This looks good to me.

I like that you left out any conditionals based on `is_active`. Even ignoring the question as to whether a case can have more than one active assignment or not, I feel like enforcing that at the level of ActiveRecord DSL statements is a mistake I've made before.

PR #63, opened 2020-04-05
28 new case contact form

### LINKED ISSUE #28: new_case_contact form
Related to #27 and #3 

Source: `new_case_contact` button

Fields to display on form:
- type (radio button, populated by hard coded options)
- hours (drop down menu that accepts only numbers but allows 15 minute increments. Ex: .15, 1.15, etc)
  - date - (pre-filled with html placeholder: mm/dd/yyyy and default to today's date) (post MVP: include a calendar popup, likely want to refer to a library for this)
 - ^^^ will be related to new post MVP issue: research libraries for calendar pop up options.
- submit button [related to new issue to come re: submit button backend stuff]
   - submit redirects to dashboard view

duaimei:
Resolves #28

### Type of change

* Breaking change (fix or feature that would cause existing functionality to not 
work as expected)
* This change requires a documentation update
* Documentation update


Thrillberg:
@johncarlocerna and @duaimei Is this ready for review? I think we just merged something into `master` that this needs to update against. Do you want me to handle it or would y'all rather? I can review in the meantime... 👀

app/controllers/case_contacts_controller.rb
```diff
@@ -10,19 +10,30 @@ def index
   # GET /case_contacts/1
   # GET /case_contacts/1.json
   def show
+    @case_contact_number = CasaCase.find(@case_contact.casa_case_id).case_number
```
Thrillberg, line 13: In the show view, it is traditional to have something like this:
```ruby
@case_contact = CaseContact.find(params[:id])
```
From the `CaseContact`, we can then derive the `case_number` in the view (or elsewhere), which I'll indicate in a note on the view as well. 😄

app/views/case_contacts/show.html.erb
```diff
@@ -7,7 +7,7 @@
 
 <p>
   <strong>Casa case:</strong>
-  <%= @case_contact.casa_case_id %>
+  <%= @case_contact_number %>
```
Thrillberg, line 10: This can be rewritten as:
```
<%= @case_contact.casa_case.case_number %>
```
If `@case_contact` is defined in the controller action.

app/controllers/case_contacts_controller.rb
```diff
@@ -10,19 +10,30 @@ def index
...
   end
 
   # GET /case_contacts/new
   def new
     @case_contact = CaseContact.new
+    @casa_cases = current_user.casa_cases
```
Thrillberg, line 19: I like this! 😃

app/controllers/case_contacts_controller.rb
```diff
@@ -10,19 +10,30 @@ def index
...
   def create
-    @case_contact = CaseContact.new(case_contact_params)
+    info =
+      {
+        creator_id: current_user.id,
+        casa_case_id: params[:casa_case_id],
+        contact_type: params[:contact_type],
+        duration_minutes: params[:duration_minutes],
+        occurred_at: params[:case_contact][:occurred_at]
+      }
```
Thrillberg, line 35: To construct the arguments for our new `CaseContact`, can we use the `case_contact_params` (which is an instance of `CaseContactParameters`)? I'm curious if there was a reason another approach was used here. I'm also happy to pair on it, if that seems better! 😄

spec/requests/case_contacts_spec.rb
```diff
@@ -32,39 +32,7 @@
...
-        }.to change(CaseContact, :count).by(1)
-      end
-
-      it "redirects to the created case_contact" do
-        sign_in volunteer
-
-        post case_contacts_url, params: { case_contact: valid_attributes }
-        expect(response).to redirect_to(case_contact_url(CaseContact.last))
-      end
-    end
```
Thrillberg, line 66: It looks like we are using these routes. Is there a reason these specs should be removed?

app/views/case_contacts/show.html.erb
duaimei: This is the shape of @case_contact_number:

{"id":19,
"creator_id":5,
"casa_case_id":4,
"contact_type":"youth",
"other_type_text":null,
"duration_minutes":15,
"occurred_at":"2020-04-05T00:00:00.000Z",
"created_at":"2020-04-05T23:36:26.440Z",
"updated_at":"2020-04-05T23:36:26.440Z"
}

we do not have the casa_case object. Which is why we have stuff in the controller to get the information.

app/controllers/case_contacts_controller.rb
duaimei: We have to find the information because the form only brings back the information in this shape:

{"id":19,
"creator_id":5,
"casa_case_id":4,
"contact_type":"youth",
"other_type_text":null,
"duration_minutes":15,
"occurred_at":"2020-04-05T00:00:00.000Z",
"created_at":"2020-04-05T23:36:26.440Z",
"updated_at":"2020-04-05T23:36:26.440Z"
}

app/controllers/case_contacts_controller.rb
duaimei: Thank you. This is done because we want to only display the casa cases that belong to the volunteer.

app/controllers/case_contacts_controller.rb
duaimei: I spent a long time trying to use case_contact_params last night. The method in the values/ folder changed the params so much that when the new CaseContact was created as an empty object.

spec/requests/case_contacts_spec.rb
duaimei: These are removed because we now expect for there to be a "current_user". Which is tested in the features.

PR #66, opened 2020-04-05
ARD for multiple user tables

compwron:
:)


doc/architecture-decisions/0003-multiple-user-tables.md
```diff
@@ -0,0 +1,21 @@
...
+
+Date: 2020-04-05
+
+## Status
+
+Accepted
+
+## Context
+
+This is planned to be a multi-tenant system. There will be multiple CASA orgs in the system, so every case, case_contact, volunteer, supervisor, casa_admin etc must have a casa_org_id, because no one is allowed to belong to multiple CASAs. Volunteer, supervisor, and casa_admin are all roles for a "User" db object. We want to not have to do null handling around all_casa_admin rows, since they will by definition have access to all CASA orgs. We have used the built-in Devise ability to have multiple user tables, as recommended to us by our Rails expert Betsy. This is to prevent needing null handling around casa_id for User records since all_casa_admin users will not have casa_id populated.
```
Zrrrpy, line 11: change "We want to not have to do null handling....." to "We don't want to do null handling"

lisavogtsf, line 11: For clarity I would introduce the concept of `all_casa_admin` before you start talking about not wanting to do null handling. I got a little whiplash trying to understand that concept. 

suggest replacing: "We want to not have to do null handling around all_casa_admin rows, since they will by definition have access to all CASA orgs."


with something more like this:
 "In addition to those existing roles, we want to create a new kind of user: all_casa_admin. We need to handle the case of super users who have access to multiple casa_orgs, so they would be difficult to handle in the existing User table--with null handling around their casa_org_id field."

PR #81, opened 2020-04-07
72 other contact type behavior

### LINKED ISSUE #72: "other" casa contact type behavior
relates to #28 

As a volunteer that is logging a new Case Contact event, I should not see the "other contact type" text box unless "other" is chosen as a Case Contact Type.

As a volunteer, I expect that if "other" is chosen, then the "other contact type" text field is required.

As a volunteer, I expect for the "other contact type" to no longer be visible if I choose "other" and then choose another option in the "case contact type" select.

duaimei:
Resolves #72

### Description
As a volunteer that is logging a new Case Contact event, I should not see the "other contact type" text box unless "other" is chosen as a Case Contact Type.

As a volunteer, I expect that if "other" is chosen, then the "other contact type" text field is required. 

As a volunteer, I expect for the "other contact type" to no longer be visible if I choose "other" and then choose another option in the "case contact type" select.

### Reviewer start
The reviewer should start in the file:
app/javascript/packs/new_casa_contact.js

This uses JQuery to detect when the select has changed, and to execute the desired behavior based on the change.

app/javascript/packs/application.js

I had to install JQuery in order to have the $ syntax.

### manual testing
There is a limitation to the cucumber tests. They don't load Javascript. Therefore going onto the case_contacts/new page and making sure that the information is saved as expected is a good way to manually test.


app/javascript/packs/new_casa_contact.js
```diff
@@ -0,0 +1,15 @@
+import $ from 'jquery'
+$('document').ready(() => {
+  // inside the app/views/casa_contact/_form, we are trying to only show a field if 'other' is chosen.
+  $("#contact_type").change((e) => {
+    if (e.target.value === 'other') {
+      $(".other-contact-type").removeAttr('hidden')
+      $(".other-contact-type input").attr('required', 'required')
+    } else {
+      $(".other-contact-type").attr('hidden', true)
+      $(".other-contact-type input").removeAttr('required')
```
Thrillberg, line 10: It's great that you watched out for making the field required! 👍 It makes me wonder if we should also handle this in the model validations. At the moment, we have no model validations, as far as I know. What do you think?

spec/features/volunteer_adds_a_case_contact_spec.rb
```diff
@@ -1,31 +1,46 @@
-require "rails_helper"
+require 'rails_helper'
 
-RSpec.feature "volunteer adds a case contact", type: :feature do
-  scenario "is successful" do
+RSpec.describe 'volunteer adds a case contact', type: :feature do
```
Thrillberg, line 3: Is there a reason you changed `feature` to `describe`? I don't think it makes a material difference but I'm just curious. 😄

spec/features/volunteer_adds_a_case_contact_spec.rb
```diff
@@ -1,31 +1,46 @@
...
+    select volunteer_casa_case_1.case_number, from: 'casa_case_id'
+    select 'School', from: 'contact_type'
+    select '60 minutes', from: 'duration_minutes'
+    fill_in 'case_contact_occurred_at', with: '04/04/2020'
 
-    # find(:css, "#case_contact_occurred_at".)
-    fill_in "case_contact_occurred_at", with: "04/04/2020"
-
-    click_on "Submit"
+    click_on 'Submit'
```
Thrillberg, line 17: I like the refactoring you did on this test! Thanks!

spec/features/volunteer_adds_a_case_contact_spec.rb
```diff
@@ -1,31 +1,46 @@
...
+    expect(page).not_to have_selector '#case_contact_other_type_text'
+    select '90 minutes', from: 'duration_minutes'
+    fill_in 'case_contact_occurred_at', with: '04/04/2020'
+    select 'Other', from: 'contact_type'
+
+    click_on 'Submit'
+
+    expect(CaseContact.first.casa_case_id).to eq volunteer_casa_case_1.id
+    expect(CaseContact.first.contact_type).to eq 'other'
+    expect(CaseContact.first.other_type_text).to eq nil
```
Thrillberg, line 43: Is it allowable for the `other_type_text` to be `nil` when `other` is selected? Perhaps this test is a case in point of why a model validation might be a good idea. But I'm surprised the test runner is letting this pass, tbh! 😃

compwron:
I will try to figure out what's up with Travis today. Thanks for the cool PR!

compwron:
Very cool! :)

PR #86, opened 2020-04-09
Resolves [#8] - Adds favicons

### LINKED ISSUE #8: Add favicon
See below for files!

[Some info on adding a favicon to a site on Ruby here.](https://discoposse.com/2018/09/04/setting-your-favicon-in-ruby-on-rails-apps/)

jimnanney:
Resolves #008

### Description
Adds favicons for android, ms, apple, ie

### Type of change

no impact

### How Has This Been Tested?

Manually tested in browser


app/assets/images/manifest.json
```diff
@@ -0,0 +1,41 @@
+{
+ "name": "App",
```
tundal45, line 2: Can we change this to a more specific name? I don't know how this manifest file is used but I would imagine it supports saving websites as apps on the phone.

jimnanney: Doh, I did not mean to add that file.  The manifest is for mulitple android sizes and would need the asset pipeline path in there.

Let me clean that up.

PR #87, opened 2020-04-11
Display volunteer's active cases in dashboard

Thrillberg:
This commit also disallows the volunteer from creating new cases
themselves. Instead, only admins should (currently) have that ability.

### Description
Display volunteer's active cases in dashboard

This commit also disallows the volunteer from creating new cases
themselves. Instead, only admins should (currently) have that ability.

### Type of change

<!-- Please outline how each level of user permissions will be affected by this change. 
(If user permissions will not be impacted by this change, please say "no impact") 

This removes volunteers' ability to create cases and allows admins to create cases.

### How Has This Been Tested?

Automated tests and confirmed in browser.


compwron:
Yay!

PR #128, opened 2020-04-19
Try to make the travis build pass with less stuff?

compwron:
I think we should update the PR template to be more friendly


colinsoleim:
Looks good to me, thanks for cleaning up my comments and my .travis.yml changes!

PR #149, opened 2020-04-20
Removes duplicated notices and adds notice styles

### LINKED ISSUE #130: success message should only display once after new case contact creation
part of epic #3

volunteer

On the page a volunteer views after successfully creating a new `case_contact`.

**Description** 
After a volunteer creates a new case contact, the success message should only display once. (It currently displays twice. See screenshot below.)

**Screenshots of current behavior**
volunteer_bug

colinsoleim:
Resolves #130

### Description

- Removes duplicates flash notices and adds basic styles using colors from the style guide.

### Type of change

No impact

### How Has This Been Tested?

Local testing and all current Rspec tests pass.

### Screenshots

Screen Shot 2020-04-20 at 2 31 57 PM

### DIFF
--- a/app/assets/stylesheets/navbar.scss
+++ b/app/assets/stylesheets/navbar.scss
@@ -11,3 +11,10 @@
     font-size: 2rem;
   }
 }
+
+.notice {
+  text-align: center;
+  padding: 8px;
+  margin: 12px 0;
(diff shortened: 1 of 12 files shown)


compwron:
Looks great! 

One thing- I would rather that non-error messages not be red, because red is normally an error color. We can refine this more later.

PR #150, opened 2020-04-20
Adds All Case Contacts Report

### LINKED ISSUE #122: Reporting: simplest report- export all case contacts
Admin

**Description**
On the Admin view, add a Reports section and one report, "All case contacts"
The export will be a CSV

Each row should have:
`case_contact.id`, `casa case number`, `case_contact.duration_minutes`, `case_contact.occurred_at`, creator email, creator name, creator supervisor's name (if any), and one column each with every contact type, set to yes/no. (don't hardcode this list, generate this out of all existing contact types, listed alphabetically"
Order rows by case contact id

Include headings in the export. So an example export could be:
```
case contact id, casa case number, duration min, occurred at, creator email, creator name, creator supervisor name, type: attorney, type: bio_parent, type: court, type: foster_parent, type: other, type: other_family, type: school, type: social_worker, type: supervisor, type: therapist, type: youth
1, 111, 30, 2020-01-01, volunteer1@example.com, Jane Smith, Jackie Smith, false, true, true, false, false, false, true, false, false, false, true
2, 111, 15, 2020-01-02, volunteer1@example.com, Jane Smith, Jackie Smith, false, true, true, false, false, false, true, false, false, false, true
3, 222, 15, 2020-01-02, volunteer2@example.com, Tim Fjeld, Jackie Smith, false, true, true, false, false, false, true, false, false, false, true
```

Also feel free to add any other columns that seem relevant. Admins will filter this report in excel for what they actually want to see

**Related issues, if any**

colinsoleim:
Resolves #122

### Description

This PR:
- Adds new reports controller to list and download reports
- Changes header nav to a dropdown which gives edit profile and reports links

Alternate Implementations:
I wasn't sure of the best way to show a link to the reports page. I ended up replacing the "Edit Profile" button with a dropdown that includes a link to "Edit Profile" and the new reports page.

Two other possibilities I considered were adding buttons on the dashboard next to "New Case Contact" labeled something like "Export Case Contacts", but it sounded like we might have different kinds of reports moving forward that wouldn't fit so easily into single buttons on the dashboard. I also considered adding a fourth section to the dashboard page listing reports similar to the reports#index in the current PR, but I decided against that because I was worried the dashboard would become too busy if we kept adding sections. 

I'd be happy to change this around if one of those other implementations fits better with what the stakeholder might want. The current reports#index doesn't look great at the moment because it's so empty.

### Type of change

Only admins can access the reports page to download reports

### How Has This Been Tested?

Local testing, existing rspec tests pass, added several new rspec tests for reports controller

### Screenshots

Screen Shot 2020-04-20 at 5 44 15 PM

### DIFF
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -37,6 +37,36 @@ class CaseContact < ApplicationRecord
   def use_other_type_text?
     contact_type == 'other'
   end
+
+  def self.to_csv
+    attributes = report_headers
+
+    CSV.generate(headers: true) do |csv|
+      csv << attributes.map(&:titleize)
+
+      all.decorate.each do |case_contact|
+        csv << [
+          case_contact&.id,
+          case_contact&.casa_case&.case_number,
+          case_contact&.duration_minutes,
+          case_contact&.occurred_at,
+          case_contact&.creator&.email,
+          'N/A', # case_contact&.creator&.name, Add back in after user has name field
+          case_contact&.creator&.supervisor&.email,
+          case_contact&.contact_type
+        ]
+      end
+    end
+  end
+
+  def self.report_headers
+    headers = %w[case_contact_id casa_case_number duration occurred_at
+                 creator_email creator_name creator_supervisor_name contact_type]
+
+    # TODO: Issue 119 -- Enable multiple contact types for a case_contact
+    # headers.concat(CaseContact::CONTACT_TYPES.map { |t| "contact_type: #{t}" })
+    headers
+  end
 end
 
 # == Schema Information
--- a/app/views/dashboard/show.html.erb
+++ b/app/views/dashboard/show.html.erb
@@ -1,9 +1,7 @@
 <% if policy(:dashboard).see_volunteers_section? %>
   <div class="row">
-    <div class="col-sm-4 col-md-3">
+    <div class="col-sm-12 dashboard-table-header">
       <h1>Volunteers</h1>
-    </div>
-    <div class="col-sm-8 col-md-9">
       <%= link_to "New Volunteer", new_volunteer_path, class: "btn btn-primary" %>
     </div>
   </div>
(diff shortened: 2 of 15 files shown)


app/views/dashboard/show.html.erb
```diff
@@ -71,8 +67,11 @@
   </table>
   <br>
   <div class="row">
-    <div class="col-sm-6 col-md-4">
+    <div class="col-sm-12 dashboard-table-header">
       <h1>Case Contacts</h1>
+      <% if policy(:dashboard).create_case_contacts? %>
+        <%= link_to "New Case Contact", new_case_contact_path, class: "btn btn-primary" %>
```
colinsoleim, line 73: Oh I meant to bring this up separately, but we completely removed the "New Case Contact" button for both admins and volunteers in a previous PR. Is that intended, or should we only hide the "New Case Contact" button on this page for admins and keep it visible for volunteers?

compwron: Hide for admins, keep for volunteers 😅

app/models/case_contact.rb
```diff
@@ -37,6 +37,36 @@ def humanized_type
...
+      csv << attributes.map(&:titleize)
+
+      all.decorate.each do |case_contact|
+        csv << [
+          case_contact&.id,
+          case_contact&.casa_case&.case_number,
+          case_contact&.duration_minutes,
+          case_contact&.occurred_at,
+          case_contact&.creator&.email,
+          'N/A', # case_contact&.creator&.name, Add back in after user has name field
```
compwron, line 54: :)

app/models/case_contact.rb
```diff
@@ -37,6 +37,36 @@ def humanized_type
...
+        ]
+      end
+    end
+  end
+
+  def self.report_headers
+    headers = %w[case_contact_id casa_case_number duration occurred_at
+                 creator_email creator_name creator_supervisor_name contact_type]
+
+    # TODO: Issue 119 -- Enable multiple contact types for a case_contact
```
compwron, line 66: Excellent :)

app/views/dashboard/show.html.erb
colinsoleim: Ok great, that's what I was hoping. The policy check  in the code above should be what we want then!

PR #151, opened 2020-04-21
Adds Bulk Create Option For Case Contacts form

### LINKED ISSUE #118: Volunteer with multiple cases can a add case contact to multiple cases at once
Volunteer

**Description**
If a volunteer has multiple cases, have an option to make an update to all of them at once so they don’t have to do it multiple times. Maybe make this a checkbox which for single-case volunteers is always checked, and for multi-case volunteers, any subset of their cases can be checked at one time. 

This will be used, for example, if a volunteer is spending time with two or more youth at the same time- for example, two siblings, or a child and the child's cousin whose case is assigned to the same volunteer. 

This will result in making two case_contact database records, not one database record for two cases. 

Make sure that it is not valid or allowed to save a case contact which is not associated to at least one case. 

We DO want this working before we onboard all the volunteers (before launch)

Screen Shot 2020-04-18 at 6 23 04 PM

**Related issues, if any**

colinsoleim:
Resolves #118

### Description
Changes casa case id from dropdown to checkbox of all assigned cases

I went ahead and did a lot of RSpec updates in this PR as well. I implemented all the currently skipped Rspec tests. I also went ahead and deleted the case_assignments and supervisor_volunteers controllers since those seem to be join tables that don't need full controllers at the moment.

### Type of change

No impact.

### How Has This Been Tested?

Adds new rspec tests, existing tests pass, and local testing

### Screenshots
Screen Shot 2020-04-21 at 11 17 39 AM


.travis.yml
```diff
@@ -19,7 +20,7 @@ install:
   - bundle install
 script:
   - bundle exec rails webpacker:compile
-  - bundle exec rails spec
+  - RUBYOPT='-W:no-deprecated -W:no-experimental' bundle exec rails spec
```
compwron, line 23: Why this?

app/views/dashboard/show.html.erb
```diff
@@ -88,7 +88,7 @@
     <tbody>
       <% @case_contacts.each do |contact| %>
         <tr>
-          <td><%= contact.occurred_at.strftime('%B %e, %Y') %></td>
+          <td data-sort="<%= contact.occurred_at.strftime("%Y%m%d%H%M%s") %>"><%= contact.occurred_at.strftime('%B %e, %Y') %></td>
```
compwron, line 91: Nice :)

compwron:
Wait why did this get closed help

colinsoleim:
> Wait why did this get closed help

Oh weird, I'm not sure why it was closed. Github had some status issues earlier in the day, so maybe it was a bug related to that?

.travis.yml
colinsoleim: I added this to hide all of the ruby 2.7 errors that fill the page when rspec gets run. There were other deprecation messages that were not readable because of how many ruby 2.7 deprecations were showing up, so I thought it'd be a good idea to hide the deprecations we can't change for now.

Let me know if you'd prefer that I revert it!

compwron: This is great thanks. We should remember to upgrade and take it out someday but this is definitely an improvement :)

compwron:
Wonderful :)

Issue #142, opened 2020-04-20
Auto-run new migrations in heroku on deploy


compwron:
Right now, if any PR including a migration is merged, heroku autodeploy does not run the migration. Then staging probably stops working because there is code that depends on a missing column, and @compwron is currently the only person with heroku permissions on staging to run 
```
heroku login
heroku run rake db:migrate
```
This is not ideal. Automate this. 

Notes:
tundal45
it looks like they support a feature called releases that gives us this ability once we add a Procfile to the repo

compwron:
this still needs doing

tundal45:
Yes, it's on my list to do this weekend but if other folks are interested in taking this on before that, feel free!

PR #163, opened 2020-04-23
Adds display name to all users

### LINKED ISSUE #116: Add Name to user
part of epic #59 

all

_db_: in the `create_table`: `users`
_frontend_: When any user is being created. Within each user dashboard (admin, supervisor, volunteer) when any user is editing their own information.

**Description**
1. The user db object should have a `display_name` field. It is called display_name instead of full_name because it is what the user will be called rather than their legal name.
1. Do not split this field into first and last name because we need to support volunteers who have many pieces to their name. 
1. If you haven't yet read https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/ please do so before working on this issue. In fact, maybe re-read it anyway, it's healthy to remember :)
1. Any user should be able to edit their own name at any time and to any length (maximum... ~200 characters for now is fine). 
1. No user should be able to edit another user's name. 

Some sample names to consider (not real current volunteer names)
1. `'Hubert Blaine Wolfeschlegelsteinhausenbergerdor
ff Sr'.length # 52`
1. `Carlos María Eduardo García de la Cal Fernández Leal Luna Delgado Galván Sanz'.length # 77`
1. `'Svetlana Viktorevna Rakhmaninova'.length # 32`

Names can change when someone gets married or divorced, when they transition genders, when they take a "stage name", in order to separate their identity from unpleasant or overly-famous people with a similar name, etc. Many software systems handle this poorly and many people are used to being annoyed and sad about how their names are handled by software. Let's not be an additional problem; we want our volunteers to not be annoyed at CASA because of us in any way. :)

More resources about names:
1. https://en.wikipedia.org/wiki/Spanish_naming_customs 
1. http://yuba.stanford.edu/~molinero/html/surname.html 
1. http://blog.tesol.org/7-naming-customs-from-around-the-world/
1. https://www.fbiic.gov/public/2008/nov/Naming_practice_guide_UK_2006.pdf

**Related issues, if any**
#133

### LINKED ISSUE #133: swap volunteer_id for display_name under "creator"
relates to epic #3

volunteer

On the page a volunteer views after successfully creating a new `case_contact`. 

**Description**
Instead of `volunteer_id` displaying next to **Creator**, `display_name` should display.

**Screenshots** 
volunteer_bug

**Relates to issues...**
dependent on #116

colinsoleim:
Resolves #116 
Resolves #133 

### Description

- Adds display_name migration and users controller to handle new profile routes/views
- Adds column picker modal to fit volunteer table on screen with new columns. After adding the name and recent contacts made columns, the volunteers table was going past 1200px in width, so I added a modal where admins can pick/choose what columns they want to see.

As general cleanup, I updated the simple coverage report to track all files in the lib and app folders. After changing that, I noticed there were a number of policy files that weren't being used, so I deleted those.

### Type of change

The profile page is set up only for current_user so no permissions are needed.

### How Has This Been Tested?

New rspec tests added, existing tests pass

### Screenshots
Screen Shot 2020-04-23 at 7 24 05 PM
Screen Shot 2020-04-23 at 7 23 56 PM
Screen Shot 2020-04-23 at 7 23 03 PM


app/models/volunteer.rb
```diff
@@ -0,0 +1,14 @@
+# CaseContact Model
```
compwron, line 1: This comment is not accurate, this is the Volunteer model. Not a big deal :)

app/views/case_contacts/show.html.erb
```diff
@@ -20,7 +20,7 @@
 
 <p>
   <strong>Occurred at:</strong>
-  <%= @case_contact.occurred_at %>
+  <%= @case_contact.occurred_at.strftime('%B %e, %Y') %>
```
compwron, line 23: :)

app/models/volunteer.rb
colinsoleim: Oh oops, too much copy and paste, thanks for catching!

app/views/dashboard/_volunteer_filters.html.erb
```diff
@@ -31,6 +31,43 @@
...
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title" id="exampleModalLabel">Pick Displayed Columns</h5>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        <div>
+					Select columns:
```
compwron, line 52: This spacing feels odd, since spacing doesn't show up in the html. Not a big deal

app/views/dashboard/_volunteer_filters.html.erb
```diff
@@ -31,6 +31,43 @@
...
+
+          <% Volunteer::TABLE_COLUMNS.each_with_index do |column, index| %>
```
compwron, line 54: cool!

spec/spec_helper.rb
```diff
@@ -2,7 +2,9 @@
 require 'capybara/rspec'
 require 'pundit/rspec'
 require 'pry'
-SimpleCov.start
+SimpleCov.start do
+  track_files "{app,lib}/**/*.rb"
```
compwron, line 6: :)

app/models/volunteer.rb
```diff
@@ -1,4 +1,3 @@
-# CaseContact Model
```
compwron: :)

compwron:
Because https://github.com/rubyforgood/casa/issues/142 is still in progress I need to remember to run heroku db migrations by hand after merging this to unbreak staging.

PR #169, opened 2020-04-25
Convert bugsnag API key to ENV variable

colinsoleim:
### Description
To prevent development errors popping up in the Slack channel and for security reasons, I changed the hardcoded API key to an ENV variable.

### Type of change

No Impact

### How Has This Been Tested?

Local testing

### DIFF
--- a/config/initializers/bugsnag.rb
+++ b/config/initializers/bugsnag.rb
@@ -1,3 +1,3 @@
 Bugsnag.configure do |config|
-  config.api_key = 'c2c2efde084a0f8e7f772324577a110b'
+  config.api_key = ENV['BUGSNAG_API_KEY']
 end


eclectic-coding:
Just a thought, but if the API key is now in an ENV file, as it should be, the actually API key has already been committed to the open source log for anyone to [view](https://github.com/rubyforgood/casa/commit/e9216fd96b69d7bf0f97de2bd614863d6b917e7a).  How can we safely address this security concern? Just a thought.

compwron:
> Just a thought, but if the API key is now in an ENV file, as it should be, the actually API key has already been committed to the open source log for anyone to [view](https://github.com/rubyforgood/casa/commit/e9216fd96b69d7bf0f97de2bd614863d6b917e7a). How can we safely address this security concern? Just a thought.

It's just an error tracking key to an account without a credit card, not a big deal. I'll re-issue the key when I set the env var in heroku

PR #250, opened 2020-05-11
Security Meta Test

### LINKED ISSUE #207: All-endpoint security meta-test
all

**Description**
Add a rspec test fo some kind that iterates over all endpoints(not hardcoded- needs to support adding/removing endpoints) and tests that they all require auth. (except login)
https://github.com/rubyforgood/casa/pull/198/files

Use something like this
https://michaelsoolee.com/rails-list-routes/
https://gist.github.com/bantic/5688232
`Rails.application.routes.routes.map { |r| {alias: r.name, path: r.path.spec.to_s, controller: r.defaults[:controller], action: r.defaults[:action]}}`
`Rails.application.routes.routes.to_a.each do`

armahillo:

Resolves #207 

### Checklist

* [x] I have performed a self-review of my own code
* [x] I added comments, particularly in hard-to-understand areas (n/a, I think?)
* [x] (n/a) I updated the `/docs`
* [x] I added tests that prove my fix is effective or that my feature works
* [x] `bundle exec rake` passes locally
* [x] ~My PR title includes "WIP" or is [draft PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) if work is in progress~

### What changed, and why?
@compwron requested that there be an omnibus check for authorization, app-wide. I've created a single request spec that analyzes the routes and attempts to access each route via its appropriate verb. The expectation is that they will all require redirection to `.../sign_in` because they all require authorization.

I check this by looking for a 302 and verifying that the Location header ends with `sign_in`.

The setup specifically omits Rails and Devise routes, but since both of those are third party code, I feel ok not testing them.

It won't affect user permissions specifically, but I *did* add a `authenticate_user!` before action to a few controllers (check the diff) because they were gaping open. (The spec identified it!). I don't know what the permissions should be for those, so I just dropped a note with it. Someone from the core team can tackle it.

### How is this tested?
I wrote a spec to do this.

### Screenshots please :)
n/a


app/controllers/case_assignments_controller.rb
```diff
@@ -1,4 +1,6 @@
 class CaseAssignmentsController < ApplicationController
+  # NOTE: I don't know what auth levels to use here, but am pretty certain it needs SOMETHING.
+  before_action :authenticate_user!
```
compwron, line 3: This is taken care of by `must_be_admin` on the line below

app/controllers/casa_orgs_controller.rb
```diff
@@ -1,4 +1,6 @@
 class CasaOrgsController < ApplicationController
+  # NOTE: I don't know what auth levels to use here, but am pretty certain it needs SOMETHING.
+  before_action :authenticate_user!
```
compwron, line 3: This will probably eventually need to be `authenticate_all_casa_admin!` but we're not using it for now so this is fine (I'm not sure whether Devise `authenticate_all_casa_admin!` exists/works yet)

app/controllers/volunteers_controller.rb
```diff
@@ -1,4 +1,6 @@
 class VolunteersController < ApplicationController
+  # NOTE: I don't know what auth levels to use here, but am pretty certain it needs SOMETHING.
+  before_action :authenticate_user!
```
compwron, line 3: This is taken care of by `must_be_admin` on the line below

spec/requests/casa_orgs_spec.rb
```diff
@@ -13,6 +13,11 @@
 # sticking to rails and rspec-rails APIs to keep things simple and stable.
 
 RSpec.describe '/casa_orgs', type: :request do
+  before do
+    skip "None of these currently do authorization, and I think that might be necessary. \
```
compwron, line 17: I amok with skipping these for now because we aren't working with mutliple casa orgs for MVP. 
Eventually we will want to use the Devise test helpers like in admin_adds_new_case_spec.rb
```
sign_in admin # devise helper to sign in as the user saved in the variable "admin"
```

spec/requests/security_meta_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      it "Redirects to the sign_in page with #{route.name}" do
+        send(route.verb.downcase.to_sym, route.path)
+        expect(response.status).to eq(302)
+        expect(response.header['Location']).to be_end_with("sign_in")
+      end
+    end
+  end
+
+end
```
compwron, line 53: You can fix the line ending here by running `bundle exec standardrb --fix` 
It's not required in CI yet so if it's a pain don't worry about it :)

app/controllers/case_assignments_controller.rb
armahillo: Are you certain? It was failing the spec beforehand, but passed afterwards.

spec/requests/casa_orgs_spec.rb
armahillo: I figured that was the case but wasn't sure of the specifics, so I didn't want to implement it incorrectly.

app/controllers/case_assignments_controller.rb
compwron: Hmm well it *should* be taken care of by `must_be_admin` but it sounds like we have a bug! Thanks for the catch :)

PR #279, opened 2020-05-20
Something in this branch broke prod CSS

compwron:
Something in this branch broke prod CSS

Figure out what it was and bring back the rest of it...


colinsoleim:
@compwron I wonder if it's related to sprockets-rails gem since that gem interacts with the asset pipeline. Is there a reason we reverted from version 3.x to 2.x?

compwron:
Yeah I'm suspicious of that too. the reason we reverted is that the variable `$red` wasn't working, and the revert made it work... even though the variable that was next to it looks the same and *did* work.

PR #314, opened 2020-06-03
Remove blank padding from date display to fix flaky test

colinsoleim:
### What changed, and why?

The existing PRs are failing because of a flaky test that is trying to find specific values in the `Last Contact Made` column of the volunteer table. Even though the `strftime` calls are identical in the test and the view output, it looks like spacing is being automatically fixed in the HTML on the view side, but the same correction isn't being made in the tests.

This PR changes the relevant `strftime` calls to remove any padding in both the view and the tests.


compwron:
Thanks! :) :)

PR #343, opened 2020-06-14
Allow case_assignments to be unassigned

### LINKED ISSUE #20: supervisor can remove (unassign) volunteer from case
supervisor cannot deactivate volunteer account (cannot lock volunteer out of login)

maebeale:
Resolves #20 (Also related to number 230)

### What changed, and why?
- Added unassign endpoint
- Includes redirect to volunteer if unassigning from Volunteer edit
- Added Status column to index

### Notes
- Would really prefer not to have to add get and patch routes. Any advice here?

- Admin permissions: Admins can unassign (remove) a Volunteer from a CasaCase

Help, please!

### Screenshots please :)
Screen Shot 2020-06-14 at 2 48 18 PM

### DIFF
--- a/app/views/casa_cases/_volunteer_assignment.html.erb
+++ b/app/views/casa_cases/_volunteer_assignment.html.erb
@@ -8,6 +8,7 @@
           <tr>
             <th>Volunteer Name</th>
             <th>Volunteer Email</th>
+            <th>Status</th>
             <th>Actions</th>
           </tr>
         </thead>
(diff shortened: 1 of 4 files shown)


app/views/casa_cases/_volunteer_assignment.html.erb
```diff
@@ -25,7 +25,7 @@
                 <% end %>
               </td>
               <td>
-                <% if assignment.is_active? %>
+                <% if assignment.is_active? && current_user.role == "casa_admin" || assignment.volunteer.supervisor == current_user %>
```
compwron, line 28: `volunteer.supervisor` should be `volunteer&.supervisor` because not all volunteers will have supervisors

maebeale: @compwron i'm pretty sure that the `&` in this case is only verifying that the `volunteer` is present, so if the `volunteer` doesn't have a `supervisor` `volunteer.supervisor` will be `nil`

maebeale: pretty sure the `&` would be needed if not all `assignments` have `volunteers` (but the `assignment` has a `belongs_to` that validates `volunteer` is present)

compwron: makes sense :)

PR #350, opened 2020-06-14
Validate the user has an active role

### LINKED ISSUE #179: [BUG] Handle inactive user login attempt
**What's the bug?**
Logging as an inactive user causes an error. 

Desired behavior is NOT successful login.

Needs design, but we want an unsuccessful login and a message noting that the account is inactive

Need to add a test to ensure this doesn't happen in the future.

**How to make it happen**
1. As a logged out user go to the homepage/login page
2. Login with valid inactive user email and password from seeds.rb file
3. Click "log in" button

**Screenshots**
Screen Shot 2020-04-25 at 1 48 46 PM

**Extra info:**
Please talk to maintainers/organizers about exactly what we want to display on unsuccessful but non erroring login for inactive user.

alindeman:
If the user is inactive, display a localized error message in the flash message and do not allow them to login, nor see any pages that require authentication.

Resolves #179 

### What changed, and why?
Devise has a [mechanism for determining if a user is active and refusing their login if not](https://rdoc.info/github/plataformatec/devise/master/Devise/Models/Authenticatable). This is checked at login and on every request. This PR hooks it up to our user model and inactive role model.

If a user has their role changed to inactive, they will no longer be able to login or view any content that requires authentication.

I wrote a couple of model specs. I'm open to writing a feature spec if folks think it would be valuable, as pretty much all of the underlying logic is provided by Devise.

### Screenshots please :)
inactive user message

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -44,6 +44,18 @@ class User < ApplicationRecord
 
     raw_token
   end
+
+  # Called by Devise during initial authentication and on each request to
+  # validate the user is active. For our purposes, the user is active if they
+  # do not have the inactive role.
+  def active_for_authentication?
+    super && !inactive?
+  end
+
+  # Called by Devise to generate an error message when a user is not active.
+  def inactive_message
+    inactive? ? :inactive : super
+  end
 end
 
 # == Schema Information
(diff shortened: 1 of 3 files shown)


compwron:
very cool! ✨🙌🏼

PR #381, opened 2020-06-21
add message for miles driven

### LINKED ISSUE #355: Validate no miles driven recorded for non-driving case contact types
If a volunteer tries to save a case contact with miles_driven more than 0 for any contact type other than in_person, show a little pop-up asking if they're sure (but let them do it- don't disallow the save if they click that they're sure)

Screen Shot 2020-06-14 at 4 19 49 PM

renatamarques97:
Resolves #355

### What changed, and why?
pop-up added when changing miles driven to more than 0

### Screenshots please :)
Screen Shot 2020-06-21 at 19 38 13


app/javascript/packs/case_contact.js
```diff
@@ -0,0 +1,8 @@
+window.onload = function() {
+  var milesInput = document.getElementById("case_contact_miles_driven");
+  milesInput.onchange = function(){
+    if(milesInput.value > 0){
+      alert('Are you sure?');
```
compwron, line 5: Thanks for working on this! This issue also needs a check for the contact type. The pop-up should only show up if the contact_type is not "in person". Please also make the message more clear about what the problem is. "Contact type is currently X, remember to double-check that is correct before submitting"

PR #395, opened 2020-06-28
volunteers can edit case contacts

scarymercedes:
### What github issue is this PR for, if any?
Resolves #379: Volunteers can edit case contacts

### What changed, and why?

- Added a link to the edit page on the volunteer contact page
- Added logic to the edit form to cycle through the list of contact types and pre-check the checkboxes for applicable values

### How will this affect user permissions?
- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

### How is this tested? (please write tests!) 💖💪
Added case_contacts/edit is listing all the contact methods from the model to verify that the edit view is returning all the contact types for the case_contact model

### Screenshots please :)
volunteer-edit-contact

### Feelings gif (optional)
💪

PR #398, opened 2020-06-29
volunteers can edit case contacts

### LINKED ISSUE #379: volunteers can edit (their own) case contacts
part of epic #3

volunteer

In the volunteer dashboard under "Case Contacts"

**Description**
In the Case Contacts part of the volunteer dashboard, their should be a column called "Actions" under which there is hyperlinked text "Edit" that takes you to an edit case contact page (example of what this looks like in the Admin dashboard under "Volunteers"). 

Volunteers should be able to edit _all_ fields in their existing case_contacts. Volunteers can only see and edit their own case_contacts.

scarymercedes:
Resolves #379: Volunteers can edit case contacts

### What changed, and why?

- Added a link to the edit page on the volunteer contact page
- Added logic to the edit form to cycle through the list of contact types and pre-check the checkboxes for applicable values

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

Added case_contacts/edit is listing all the contact methods from the model to verify that the edit view is returning all the contact types for the case_contact model

### Screenshots please :)
volunteer-edit-contact

### Feelings gif (optional)
💪

### DIFF
--- a/app/views/case_contacts/_form.html.erb
+++ b/app/views/case_contacts/_form.html.erb
@@ -36,6 +36,9 @@
                value="<%= contact_type %>"
                name="case_contact[contact_types][]"
                id="case_contact[contact_types][<%= index %>]"
+               <% if case_contact.contact_types.to_s.include?(contact_type) %>
+               checked=true
+               <% end %>
         >
         <label class="form-check-label" for="case_contact[contact_types][<%= index %>]">
           <%= contact_type.titleize %>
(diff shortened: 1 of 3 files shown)


scarymercedes:
I decided to open a new PR using a branch of this repo instead of my fork. (The old PR was https://github.com/rubyforgood/casa/pull/395).

Turns out I definitely broke the "volunteer adds a case contact" feature in the previous PR. As part of this, in the case_contact view form I started checking over all selected contact types and seeing if they were included in the list of possible contact types (to pre-check the boxes). What I didn't count on was having no contact type selected (which is the case when you create a new one), which would return a null value. There's no "included in" property of a null value, like there is of a string, so "is selected values included in list of possible values" turned into a problem.

I fixed it by putting a to_s when I'm getting the selected contact types in the form, so I'm getting an empty string instead of a null value. That stopped breaking the "new case contact" page and its test is passing again.

Issue #378, opened 2020-06-21
Supervisor should be able to edit their own email&name from the edit supervisor page


compwron:
**What type of user is this for? volunteer/supervisor/admin/all**
supervisor

**Description**
Supervisor should be able to edit their **own** email&name from the edit supervisor page

**Screenshots of current behavior, if any**
<img width="1276" alt="Screen Shot 2020-06-21 at 2 58 24 PM" src="https://user-images.githubusercontent.com/578159/85236134-e6ccb900-b3cf-11ea-8593-0691598e2e9a.png">

cliftonmcintosh:
@compwron This is addressed in #416. I saw this while addressing #377 and thought it made sense to make the update for both together.

cliftonmcintosh:
I will assign this to myself and move it into In Progress.

PR #416, opened 2020-07-18
Issue 377 remove submit button on supervisor

### LINKED ISSUE #377: Remove Submit button on Supervisor Edit Supervisor page
Supervisor

**Description**
Log in as a supervisor and go to the "edit" page of another supervisor. Because there are no editable fields controlled by the Submit button, the button should be removed. 

Screen Shot 2020-06-21 at 2 58 14 PM

cliftonmcintosh:
Resolves #377 and #378 
Addressed #378 by accident before realizing the issue even existed. Can break https://github.com/rubyforgood/casa/pull/416/commits/8444ab2736df16c57db63418fd6472265c634505 (which address #378) out into a separate PR if desired.

### What changed, and why?
Conditionally display the Submit button in the supervisor edit form. Only display it if the current user can update supervisor email or update supervisor name

- Volunteer permissions: No effect
- Supervisor permissions: No effect
- Admin permissions: No effect

Tests were added to verify whether or not the Submit button is present.

#### Concerns about added tests

1. Just checking for a button with "Submit" on it seems a little brittle. I'm not very familiar with the tests that verify HTML. Is there a better check? I'm happy to update
2. I also wondered if these should be broken into two separate test files, one for admin views supervisor edit and the other for supervisor view supervisor edit. It felt funny to have tests that only did one thing -- check for a button. I'm happy to break them out into separate files. I'm worried about scope creep if we start adding other tests for the supervisor view as part of this PR. I'm not sure I have the time this weekend to build out tests for the supervisor view.

### Existing bug
(Issue #378)
A supervisor cannot edit self. That is because the policy `UserPolicy#update_supervisor_email?` and `UserPolicy#update_supervisor_name?` do not behave as expected. The check `record == user` returns false even when the user is on their own edit page. That is because `record` equals `:user`. It does not equal the object of the current logged-in user. Here is an example from the console with a user, "larrykirlin@example.org", logged in and on that user's own supervisor edit view.

```
From: /home/cmcintosh/code/rubyforgood/casa/app/policies/user_policy.rb:12 UserPolicy#update_supervisor_email?:

    10: def update_supervisor_email?
    11:   binding.pry
 => 12:   user.casa_admin? || record == user
    13: end

[1] pry(#<UserPolicy>)> record
=> :user
[2] pry(#<UserPolicy>)> user
=> #<User id: 109, email: "larrykirlin@example.org", created_at: "2020-07-12 20:50:44", updated_at: "2020-07-12 20:50:44", role: "supervisor", casa_org_id: 1, display_name: "Larry Kirlin">
[3] pry(#<UserPolicy>)> record == user
=> false
```
As part of this PR's changes, the use of these functions is updated in the supervisor edit template to allow a supervisor to edit self. See https://github.com/rubyforgood/casa/commit/8444ab2736df16c57db63418fd6472265c634505.

#### Before this change

The "Submit" button is visible when a supervisor is viewing the page for another supervisor
before-removing-submit-button

#### After this change

The "Submit" button is not visible when a supervisor is viewing the page for another supervisor
after-removing-submit-button

The "Submit" button is still visible when an admin is viewing the page for a supervisor
admin-sees-button

#### Bug that doesn't allow a supervisor to edit self
Covered in issue #378 

Here are screenshots showing the existing bug that doesn't allow a supervisor to edit self

Existing functionality has fields for self un-editable but Submit button visible
supervisor-cannot-edit-self-in-master

After this change, the fields for self are editable and the Submit button is still visible
supervisor-can-edit-self

### DIFF
--- a/app/views/supervisors/edit.html.erb
+++ b/app/views/supervisors/edit.html.erb
@@ -36,7 +36,9 @@
       </div>
 
       <div class="actions">
-        <%= form.submit "Submit", class: "btn btn-primary" %>
+        <% if policy(:user).update_supervisor_email? || policy(:user).update_supervisor_name? %>
+          <%= form.submit "Submit", class: "btn btn-primary" %>
+        <% end %>
       </div>
     <% end %>
   </div>
--- a/app/views/supervisors/edit.html.erb
+++ b/app/views/supervisors/edit.html.erb
@@ -19,7 +19,7 @@
 
       <div class="field form-group">
         <%= form.label :email %>
-        <% if policy(:user).update_supervisor_email? %>
+        <% if policy(@supervisor).update_supervisor_email? %>
           <%= form.text_field :email, class: "form-control" %>
         <% else %>
           <input class="form-control" type="text" placeholder="<%= @supervisor.email %>" readonly>
@@ -28,7 +28,7 @@
 
       <div class="field form-group">
         <%= form.label :display_name %>
-        <% if policy(:user).update_supervisor_name? %>
+        <% if policy(@supervisor).update_supervisor_name? %>
           <%= form.text_field :display_name, class: "form-control" %>
         <% else %>
           <input class="form-control" type="text" placeholder="<%= @supervisor.display_name %>" readonly>
@@ -36,7 +36,7 @@
       </div>
 
       <div class="actions">
-        <% if policy(:user).update_supervisor_email? || policy(:user).update_supervisor_name? %>
+        <% if policy(@supervisor).update_supervisor_email? || policy(@supervisor).update_supervisor_name? %>
           <%= form.submit "Submit", class: "btn btn-primary" %>
         <% end %>
       </div>
--- a/spec/features/user_views_supervisor_edit_spec.rb
+++ b/spec/features/user_views_supervisor_edit_spec.rb
@@ -1,0 +1,27 @@
+require "rails_helper"
+
+RSpec.describe "view supervisor edit", type: :feature do
+  context "when the current user is a supervisor" do
+    it "displays a submit button" do
+      current_supervisor = create(:user, :supervisor)
+      other_supervisor = create(:user, :supervisor)
+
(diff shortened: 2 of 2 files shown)


cliftonmcintosh:
Hold on. I believe I've solved the problem with a supervisor unable to edit self and will add a commit.

app/views/supervisors/edit.html.erb
```diff
@@ -19,7 +19,7 @@
 
       <div class="field form-group">
         <%= form.label :email %>
-        <% if policy(:user).update_supervisor_email? %>
+        <% if policy(@supervisor).update_supervisor_email? %>
```
cliftonmcintosh, line 22: This change addresses #378

app/views/supervisors/edit.html.erb
```diff
@@ -28,15 +28,17 @@
 
       <div class="field form-group">
         <%= form.label :display_name %>
-        <% if policy(:user).update_supervisor_name? %>
+        <% if policy(@supervisor).update_supervisor_name? %>
```
cliftonmcintosh, line 31: This change addresses #378

app/views/supervisors/edit.html.erb
```diff
@@ -28,15 +28,17 @@
...
           <%= form.text_field :display_name, class: "form-control" %>
         <% else %>
           <input class="form-control" type="text" placeholder="<%= @supervisor.display_name %>" readonly>
         <% end %>
       </div>
 
       <div class="actions">
-        <%= form.submit "Submit", class: "btn btn-primary" %>
+        <% if policy(@supervisor).update_supervisor_email? || policy(@supervisor).update_supervisor_name? %>
```
cliftonmcintosh, line 39: Behind the scenes, these two checks are actually identical. `UserPolicy#update_supervisor_name?` actually just calls `UserPolicy#update_supervisor_email?`, but it makes sense to include them both here in this check since there is no guarantee that they will be the same thing in the future.

PR #439, opened 2020-07-26
Allow users to change their passwords.

### LINKED ISSUE #189: add Reset Password to Edit Profile
part of epics #3, #105, #4

all (but most pressingly, volunteers)

On the <a href="https://casa-r4g-staging.herokuapp.com/users/edit">edit profile</a> page for all users.

**Description**
Add "Reset password" field to Edit Profile page.

**Related issues**
#125

seanmarcia:
Add in some error helpers. Font Awesome. And a bit more.

This resolves #189 

At the core this resolves #189 allowing folks to change their password. We already have the ability for folks to reset their password if they forgot it so the work done, primarily, is for logged in folks.

A summary of stuff with pictures:
-Changed the forgot password button link on the sign-in page to a button to make it more obvious it exists:

image

becomes: 

image

Next I changed the edit profile page around a bit:

edit_profit

becomes:

new_profile

When you click the change password button it opens up the following:

change_password

I also added in a helper method with styling for the error messages so they are not plain text and more visible. 
So, rather than doing something like this in our views:
```
      <% if @user.errors.any? %>	
        <div id="error_explanation">	
          <h2><%= pluralize(user.errors.count, "error") %> prohibited this user from being saved:</h2>	

          <ul>	
            <% user.errors.full_messages.each do |message| %>	
              <li><%= message %></li>	
            <% end %>	
          </ul>	
        </div>	
      <% end %>
```
We can just do `<%= errors_for(@user) %>`

Error messages now look like this: 

error messages

I also updated the flash messages how they are shown in the `application.html.erb` and made them dismissible. I also added in `font-awesome-rails` to the gemfile to put in an X into the alerts. (Plus it is handy to use elsewhere)

I also put in a tiny javascript function that will auto close alerts from pages after a couple of seconds.

The new alerts look like this:

alerts

It is still a WIP because I have a couple questions:

How is the team testing stuff that overlaps with gems? I'm happy to add in tests for the changing password functionality but it kind of feels like I'm testing stuff that overlaps with devise and didn't want to add cruft to the app.

I also added in Font Awesome which you all may not be cool with.

I'm also not sure where that tiny js function should live -- it is currently in `application.js`.

Let me know what you want me to do with all this and I'll move it forward.

### DIFF
--- a/app/assets/stylesheets/shared/flashes.scss
+++ b/app/assets/stylesheets/shared/flashes.scss
@@ -3,3 +3,7 @@
   padding: 8px;
   margin: 12px 0;
 }
+
+.alert-dismissible .close {
+  padding: 0.5rem 0.75rem;
+}
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem "bootstrap", "~> 4.5.0" # frontend styling library
 gem "devise" # for authentication
 gem "draper" # adds decorators for cleaner presentation logic
 gem "faker" # creates realistic seed data, valuable for staging and demos
+gem "font-awesome-rails"
 gem "jbuilder", "~> 2.7" # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem "jquery-datatables-rails", "~> 3.4.0"
 gem "jquery-rails" # Add jquery to asset pipeline
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -124,6 +124,8 @@ GEM
       i18n (>= 1.6, < 2)
     ffi (1.12.2)
     ffi (1.12.2-x64-mingw32)
+    font-awesome-rails (4.7.0.5)
+      railties (>= 3.2, < 6.1)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
     http-cookie (1.0.3)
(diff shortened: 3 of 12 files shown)


compwron:
@seanmarcia This is great! 💖
> How is the team testing stuff that overlaps with gems? I'm happy to add in tests for the changing password functionality but it kind of feels like I'm testing stuff that overlaps with devise and didn't want to add cruft to the app.
There's no particular approach right now. I agree that Devise doesn't need testing. It would be nice to have a test around "go to page X and then Devise does Y" so probably a capybara test would be best. 

> I also added in Font Awesome which you all may not be cool with.
I like it!

> I'm also not sure where that tiny js function should live -- it is currently in application.js.
That's an ok place for now. We could add a file in `js/src`

Approved! ✨🙌🏼💖

app/assets/stylesheets/shared/flashes.scss
```diff
@@ -3,3 +3,7 @@
   padding: 8px;
   margin: 12px 0;
 }
+
+.alert-dismissible .close {
+  padding: 0.5rem 0.75rem;
+}
```
compwron, line 9: nit: missing linebreak

seanmarcia: Fixed :)

seanmarcia:
Also added in a couple capybara system specs.

PR #453, opened 2020-07-27
Enforce server-side that at least one case must be selected

### LINKED ISSUE #268: Do not allow a volunteer without an assigned case to submit a case contact
unknown

**Description**
When logged in as a user with no assigned case, it is possible to visit the `/case_contacts/new` page and submit an empty form, which results in an error because casa_case_id is nil. 

We *do* want to allow a volunteer without a case to view this page, so that they can see what they will do when they have a case but we don't want the form to be submittable without a case_id specified, so that they don't feel that the system is broken or unreliable. 

**Desired fix:** 
If there is no case associated with the user who is visiting the `/case_contacts/new` page, show a red error saying "no case available"

Since in this case this page was reached by changing the url directly, fixing this bug does not include hiding any "Create case contact" buttons/links on the UI.

Screen Shot 2020-06-14 at 1 52 27 PM
Screen Shot 2020-06-14 at 1 52 22 PM
Screen Shot 2020-06-14 at 1 52 11 PM

alindeman:
Resolves #268

### What changed, and why?
There's already client-side validation to make sure a user selects at least one case, but this should also be present server-side in case a browser doesn't enforce it properly.

Additionally, if a user is not assigned any cases at all, there will be no checkboxes to select, so the client-side validation may be inadvertently bypassed.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

Via local development environment and a new request spec.

I waffled on introducing a higher level feature spec, but decided against it because it seemed like an edge case. I'd be happy to take in feedback about this decision though.

### Screenshots please :)
Screenshot 2020-07-26 at 21 54 53

### DIFF
--- a/app/controllers/case_contacts_controller.rb
+++ b/app/controllers/case_contacts_controller.rb
@@ -25,17 +25,27 @@ class CaseContactsController < ApplicationController
   end
 
   def create
-    @casa_cases = policy_scope(CasaCase).where(id: params.dig(:case_contact, :casa_case_id))
+    # These variables are used to re-render the form (render :new) if there are
+    # validation errors so that the user does not lose inputs to fields that
+    # they did previously enter.
+    @casa_cases = policy_scope(CasaCase)
+    @case_contact = CaseContact.new(create_case_contact_params)
+
+    selected_cases = @casa_cases.where(id: params.dig(:case_contact, :casa_case_id))
+    if selected_cases.empty?
+      flash[:alert] = "At least one case must be selected"
+      render :new
+      return
+    end
 
     # Create a case contact for every case that was checked
-    case_contacts = @casa_cases.map do |casa_case|
+    case_contacts = selected_cases.map do |casa_case|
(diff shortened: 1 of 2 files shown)


app/controllers/case_contacts_controller.rb
```diff
@@ -25,17 +25,27 @@ def new
   end
 
   def create
-    @casa_cases = policy_scope(CasaCase).where(id: params.dig(:case_contact, :casa_case_id))
+    # These variables are used to re-render the form (render :new) if there are
```
compwron, line 28: Great comment! :)

spec/requests/case_contacts_spec.rb
```diff
@@ -63,6 +63,19 @@
         expect(response).to be_successful
       end
     end
+
+    context "with no cases selected" do
+      it "presents the user with a relevant error message" do
```
compwron, line 68: Yay tests! Nice test :)

PR #454, opened 2020-07-27
Allow admins to import volunteers, supervisors, and cases.

### LINKED ISSUE #18: casa_admin import many volunteers
Make a UI where an admin can
1.  import a csv or paste a list of volunteer emails and names to create volunteers in the system. 
2. same for casa cases, with case_number and optional tranistion-aged youth field
3. import list of supervisor-volunteer linkages, supervisor name to volunteer name
4. import list of volunteer-case linkages, volunteer name to case_number

seanmarcia:
Resolves #18 

### What changed, and why?

This PR does a bit more than the issue of just importing volunteers. I also allows the importation of supervisors and casa cases.

I added in the `devise_invitable` gem to make sending emails to new volunteers and supervisors easy. Also added in the mailcatcher gem with some configurations in `development.rb` to enable the testing of sending the emails.

The general process for importing csv's is for folks to download the example csv, edit it, and then upload it.

I also notice that while the dropdown that shows the reports link checks if the person is a casa_admin, the `reports_controller` didn't validate that the user on the page was a casa admin. I updated the controller to use the `before_action :must_be_admin` as well as updated the user in the reports request spec to be a casa_admin.

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions: Only admins can do this.

I added in a new imports route that is tested with a request spec and I create a `FileImporter` class that is tested in /spec/models directory. (Probably not the best spot for it.)
I also added in a fixtures directory with example csv's for the three different import types (volunteers, supervisors, casa_cases.)

### Screenshots please :)

menu_item

imports

### DIFF
--- a/app/views/imports/_supervisors.html.erb
+++ b/app/views/imports/_supervisors.html.erb
@@ -1,0 +1,21 @@
+<h3><i class="fa fa-download fa-2x" aria-hidden="true"></i> 1. Download example Supervisor CSV file</h3>
+<%= link_to "/supervisors.csv", format: :csv, class: "btn btn-xs btn-success" do %>
+  <i class="fa fa-download"></i> Download
+  example supervisor csv
+<% end %>
+<br>
+<br>
+<h3><i class="fa fa-file-text-o fa-2x" aria-hidden="true"></i> 2. Upload your CSV file</h3>
+<%= form_with(url: import_supervisors_imports_path, local: :true) do |f| %>
+  <ul>
+    <li>Click the choose file button and navigate to the saved file and select it.</li>
+    <li><strong>Do not</strong> change any of the values in the first line of the example csv file.</li>
+    <li>Then click the "Import Supervisors CSV" button to import your supervisors.</li>
+
+  </ul>
+  <%= f.file_field :file, accept: 'text/csv', class: 'form-control-file', style: "margin: auto;" %>
+  <br>
+  <%= button_tag :class => "btn btn-md btn-success" do %>
+    <i class="fa fa-upload"></i> Import Supervisors CSV
+  <% end %>
+<% end %>
--- a/config/locales/devise_invitable.en.yml
+++ b/config/locales/devise_invitable.en.yml
@@ -1,0 +1,31 @@
+en:
+  devise:
+    failure:
+      invited: "You have a pending invitation, accept it to finish creating your account."
+    invitations:
+      send_instructions: "An invitation email has been sent to %{email}."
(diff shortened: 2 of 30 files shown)


seanmarcia:
While I think this is good enough to go right now and begin testing I am planning on refactor to:

Remove the custom routes `import_volunteers`, `import_supervisors`, `import_cases` and be restful and use the create action with a trait to determine which import to run. Probably encapsulate each of the three imports in a service.

I think some of these imports should also be wrapped in an `activeRecord::Base.transaction` -- particularly the casa_case import. Casa cases don't appear to be attached to a casa_org so if we fail to associate them with a user we should probably roll back the transaction otherwise the case may end up floating in the ether. I don't currently have enough context on casa cases (and the rest) to

compwron:
My preferred solution to avoid free-floating cases is, before multi-tenancy, we should add `casa_org_id` to `casa_case`

app/views/imports/_supervisors.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    <li><strong>Do not</strong> change any of the values in the first line of the example csv file.</li>
+    <li>Then click the "Import Supervisors CSV" button to import your supervisors.</li>
+
+  </ul>
+  <%= f.file_field :file, accept: 'text/csv', class: 'form-control-file', style: "margin: auto;" %>
+  <br>
+  <%= button_tag :class => "btn btn-md btn-success" do %>
+    <i class="fa fa-upload"></i> Import Supervisors CSV
+  <% end %>
+<% end %>
```
compwron, line 21: Missing trailing linebreaks

config/locales/devise_invitable.en.yml
```diff
@@ -0,0 +1,31 @@
+en:
```
compwron, line 1: Yay, locales! :)

spec/models/file_importer_spec.rb
```diff
@@ -0,0 +1,79 @@
+require "rails_helper"
```
compwron, line 1: Yassssss, tests :)

spec/requests/reports_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe "/reports", type: :request do
   describe "GET /index" do
     it "renders a successful response" do
-      sign_in create(:user, :volunteer)
+      sign_in create(:user, :casa_admin)
```
compwron, line 6: 😅 Good fix :)

PR #465, opened 2020-07-29
Removes mailgun and switches to sendinblue for emails.

seanmarcia:
Signed up for a [sendingblue](https://www.sendinblue.com/) account that gives 300 free emails a day. (password shared to admins)

This will fix the issue with mailgun only allowing 5 differ recipients.


compwron:
Update staging also https://github.com/rubyforgood/casa/pull/472

PR #473, opened 2020-07-30
Formatting errors fix from copy paste.

seanmarcia:
Sorry, there were a couple of typos and spaces inserted from copy and pasting. This fixes htem.


compwron:
:)
Is there a test we can write that checks this kind of thing- maybe something with the fake mail gem? You don't have to add it, I'm just looking for advice- I haven't done this in a while 😅

seanmarcia:
Hey! Honestly, I don't know. This kind of stuff (emails) are stuff that,
for me, tends to get set and then never change. Then, each time I have to
do it, I have to look it up.

PR #479, opened 2020-08-01
Move auth logic into pundit policyfile

bklang:
### What github issue is this PR for, if any?
Resolves #441

### What changed, and why?

Check the Volunteer policy for activating and deactivating users, instead of relying on view logic.

This also adds logic to the User model to consult the `VolunteerPolicy` for all `volunteer` and `inactive` users.

### How will this affect user permissions?
- Volunteer permissions: None
- Supervisor permissions: No effective change, permissions check moved to policy
- Admin permissions: No effective change, permissions check moved to policy

### How is this tested? (please write tests!) 💖💪

Existing specs and manual testing in the browser.

### Screenshots please :)

No visible changes

### Feelings gif (optional)
What gif best describes your feeling working on this issue? https://giphy.com/
How to embed:
`alt text`

PR #480, opened 2020-08-01
[WIP] Allow reactivating volunteers from Casa Case Edit screen

### LINKED ISSUE #443: Add "reactivate volunteer" button to Case view
Screen Shot 2020-07-26 at 9 46 35 AM
A supervisor or admin should be able to reactivate a volunteer from this page https://casa-r4g-staging.herokuapp.com/casa_cases/890/edit 
This is possible now that https://github.com/rubyforgood/casa/pull/436 has been done :)

bklang:

Resolves #443

Note that this PR depends on #479 

### What changed, and why?

Allow reactivating volunteers from Casa Case Edit screen.

- Volunteer permissions: N/A
- Supervisor permissions: Checks to ensure volunteer is supervised by currently-logged-in supervisor
- Admin permissions: Can activate any volunteer

WIP

### Screenshots please :)

Screen Shot 2020-08-01 at 14 22 19

### DIFF
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -38,7 +38,11 @@ class VolunteersController < ApplicationController
   def activate
     @volunteer = authorize User.find(params[:id])
     if @volunteer.update(role: "volunteer")
-      redirect_to edit_volunteer_path(@volunteer), notice: "Volunteer was activated."
+      if params[:redirect_to_path] == "casa_case" && casa_case = CasaCase.find(params[:casa_case_id])
+        redirect_to edit_casa_case_path(casa_case), notice: "Volunteer was activated."
+      else
+        redirect_to edit_volunteer_path(@volunteer), notice: "Volunteer was activated."
+      end
     else
       render :edit
     end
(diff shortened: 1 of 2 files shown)


bklang:
@compwron While testing this feature, I noticed that a re-activated volunteer shows up on the case as "Unassigned" (see screenshot below). Is that your intent? Or should we take the opportunity while re-activating the volunteer to also re-assign them to this case? The latter feels more intuitively correct, if the intent of the user is to re-activate the volunteer so they can continue working on thise case, but I defer to your thoughts here.

With the current implementation in this PR, the *only* action taken is to activate the volunteer, which looks like this:
Screen Shot 2020-08-01 at 14 24 09

compwron:
> @compwron While testing this feature, I noticed that a re-activated volunteer shows up on the case as "Unassigned" (see screenshot below). Is that your intent? Or should we take the opportunity while re-activating the volunteer to also re-assign them to this case? The latter feels more intuitively correct, if the intent of the user is to re-activate the volunteer so they can continue working on thise case, but I defer to your thoughts here.
> 
> With the current implementation in this PR, the _only_ action taken is to activate the volunteer, which looks like this:
> Screen Shot 2020-08-01 at 14 24 09

@bklang it is correct that a newly reactivated volunteer is not assigned to any case, because in CASA world more reactivated volunteers are being assigned to a new case (think a volunteer who took time away from CASA because their previous youth aged out of CASA, or to go to grad school, will come back and be assigned to a new youth, because the previous youth is no longer with CASA, or has been working with a new volunteer etc)

compwron:
This looks good (although lacking tests) Because of our timeline, I'm going to merge it :)

PR #481, opened 2020-08-01
Applies STI to the User model

armahillo:
Mobbed by: @seanmarcia @benabik @jimnanney and @armahillo 

n/a -- code cleanup (extensive!)

### What changed, and why?
We did 2 major passes

1) We applied STI to the User model, adding a `type` column and creating subclasses under User. This will make it easier to separate the roles and their responsibilities in teh future, as well as better figure out who should see what on the views. We also added an `:active` column, to take over the responsibility for `role: :inactive`

2) We removed all references to `:role` and updated behaviors accordingly. We tried to maintain parity with how things behaved as much as possible. 

The tests all pass. We did some manual testing of the app, but I would strongly encourage thorough manual testing be done before this is merged. Please feel free to tag me or one of the ppl above with any bugs that come up related to this.

- Volunteer permissions: shouldn't change
- Supervisor permissions: shouldn't change
- Admin permissions: shouldn't change

Ran all AT until they ran clean

### Feelings gif (optional)
How to embed:
Man on front of train knocking a railroad tie off of railroad tracks by tossing another railroad tie at it


db/migrate/20200801170524_add_type_to_user.rb
```diff
@@ -0,0 +1,24 @@
...
+  self.table_name = 'users'
+end
+
+class AddTypeToUser < ActiveRecord::Migration[6.0]
+  def up
+    add_column :users, :type, :string
+    add_column :users, :active, :boolean, default: true
+
+    # Set supervisor
+    User.where(role: "supervisor").update_all(type: "Supervisor")
```
seanmarcia, line 11: If this goes in before we "go live" we can probably kill these updates and just reset the database.

compwron: I would like to keep migrations which demonstrate how to safely migrate even when we have prod data to preserve

compwron:
Thank you for the great refactorings!  💖✨💪🙌🏼 💖 @seanmarcia @Benabik @jimnanney @armahillo

PR #503, opened 2020-08-09
Switch back to sendinblue which works its just slooooow

compwron:
This reverts commit 1863663194226d5a63703d668e8ade66aa48038e.

Backstory:
Currently we are banned from sendgrid for unknown reasons. I and Sean have both opened tickets to try to get this fixed. 
We have supervisor training coming up on wednesday and we would really like them to have accounts by then. 
Log into https://my.sendinblue.com/dashboard# by choosing "log in with google" and logging in as casa@ rubyforgood. Sean or I have the password. (you should not need to do this, but just in case)
See emails sent at https://app-smtp.sendinblue.com/statistics
Keep in mind that email delivery is *very* slow. On 4 August 2020 it took about 4-6 hours. On 6 August 2020 it took about 30 minutes. I guess that's what we get for free :/ 

You can see where the env variables used here are configured at: https://dashboard.heroku.com/apps/casa-r4g-staging/settings

Screen Shot 2020-08-09 at 10 55 12 AM
Screen Shot 2020-08-09 at 10 56 00 AM

Screen Shot 2020-08-09 at 10 48 44 AM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,6 @@ gem "pg", ">= 0.18", "< 2.0" # Use postgresql as the database for Active Record
 gem "puma", "~> 4.3" # Use Puma as the app server
 gem "pundit" # for authorization management - based on user.role field
 gem "sass-rails", ">= 6" # Use SCSS for stylesheets
-gem 'sendgrid-ruby' # Emails
 gem "skylight" # automated performance testing https://www.skylight.io/
 gem "sprockets-rails" # TODO remove (deprecated) Provides Sprockets implementation for Rails Asset Pipeline.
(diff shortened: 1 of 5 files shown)


compwron:
Note: Just now, after deploying this to staging I ran `check_email_works.rb` and it appears to work. Now we waaaaaaaaait for emails. Then deploy to prod and create supervisor accounts :)

compwron:
It worked! Not too slow, either. 
Screen Shot 2020-08-09 at 11 12 53 AM

compwron:
Testing:
I tried to login with 123456 and it said "you have a pending invite" 
Then I reset my password and got this email 
Screen Shot 2020-08-09 at 11 15 32 AM
I changed my password to the same password, 123456, and login was successful. We should have better password requirements in prod.

PR #551, opened 2020-08-17
add start and end date to list of assignments

### LINKED ISSUE #538: display start and end date of current & past volunteer assignments
**Part of epics** #4 and #105 

supervisors and admins

On Editing CASA Case pages where past volunteer assignments display, accessible by supervisors and admins.

**Description**
Two new columns should be added to the table that displays information about Assigned Volunteers on the Editing CASA Case page:
- Start Date
- End Date

The _Start Date_ column should display the creation date of this volunteer/case assignment, a.k.a. the date this volunteer was assigned to the case.

The _End Date_ column should display the termination date of this volunteer/case assignment, a.k.a. the date the volunteer was unassigned from the case.

cliftonmcintosh:
Resolves #538

### What changed, and why?
add start date and end date columns to the table that shows volunteer assignments

- Volunteer permissions: no effect
- Supervisor permissions: no effect
- Admin permissions: no effect

~~No tests! AFAICT, there are no tests for this view written yet. If we want them written as part of this PR, please say so, and we can figure out how to set up tests for this view.~~

I had originally missed the existing tests for the case assignments feature. Tests for the presence of the dates have now been added.

### Screenshots
sample-start-end-dates

### DIFF
--- a/app/views/casa_cases/_volunteer_assignment.html.erb
+++ b/app/views/casa_cases/_volunteer_assignment.html.erb
@@ -9,6 +9,8 @@
             <th>Volunteer Name</th>
             <th>Volunteer Email</th>
             <th>Status</th>
+            <th>Start Date</th>
+            <th>End Date</th>
             <th>Actions</th>
           </tr>
         </thead>
@@ -22,6 +24,12 @@
                   <span class='badge badge-success text-uppercase'>Assigned</span>
                 <% else %>
                   <span class="badge badge-danger text-uppercase"><%= assignment.volunteer.active_volunteer ? "Unassigned" : "Deactivated volunteer" %></span>
+                <% end %>
+              </td>
+              <td><%= assignment.created_at.strftime('%B %e, %Y') %></td>
+              <td>
+                <% unless assignment.is_active? %>
+                  <%= assignment.updated_at.strftime('%B %e, %Y') %>
                 <% end %>
               </td>
               <td>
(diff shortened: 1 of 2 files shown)


compwron:
Looks good! 
Yes, please do add tests in this PR :)

cliftonmcintosh:
Tests added. I added ids to the table columns for the case assignments to facilitate testing. If there is a different, preferred way to access those columns, please let me know.

app/views/casa_cases/_volunteer_assignment.html.erb
```diff
@@ -9,22 +9,30 @@
...
             <th>Actions</th>
           </tr>
         </thead>
         <tbody>
           <% @casa_case.case_assignments.each do |assignment| %>
             <tr>
-              <td><%= link_to assignment&.volunteer&.display_name, edit_volunteer_path(assignment.volunteer) %></td>
-              <td><%= assignment&.volunteer&.email %></td>
-              <td>
+              <td id="volunteer-name"><%= link_to assignment&.volunteer&.display_name, edit_volunteer_path(assignment.volunteer) %></td>
```
compwron, line 20: ids should be unique on the page. I believe that when there is more than one volunteer shown, this i will not be unique and we will see a chrome console error. Can you check whether this is true, and if so fix it? Maybe something like id=`volunteer-name-${idx}` where idx is brought in by the each loop https://apidock.com/ruby/Enumerator/each_with_index

cliftonmcintosh:
@compwron 

There were no errors in the console with Chrome. However, you are correct that ids should be unique to a page. I have opened #556 to change the ids to data attrbutes.

PR #554, opened 2020-08-17
Add new contact types

### LINKED ISSUE #549: Add contact types
We learned during supervisor training that we want more contact types:
1.  Medical Professional
1. DSS Worker
1. Other Support Worker

Garbar:
Resolves #549 

### What changed, and why?
Add new contact types:
- Medical Professional
- DSS Worker
- Other Support Worker

### DIFF
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -25,6 +25,9 @@ class CaseContact < ApplicationRecord
     supervisor
     therapist
     youth
+    medical_professional
+    dss_worker
+    other_support_worker
   ].freeze
 
   IN_PERSON = "in-person"


app/models/case_contact.rb
```diff
@@ -25,6 +25,9 @@ class CaseContact < ApplicationRecord
     supervisor
     therapist
     youth
+    medical_professional
+    dss_worker
```
cliftonmcintosh, line 29: When the decorator is used to display the dss_worker type, what will it look like? Will it be "DSS Professional" or "Dss Professional"? If it is "Dss Professional", will that be acceptable?

https://github.com/rubyforgood/casa/blob/master/app/decorators/case_contact_decorator.rb#L43-L47

compwron: We do want "DSS Worker", we should add a UI prettifier somehow also

PR #560, opened 2020-08-18
Supervisor dashboard performance

seanmarcia:
First PR for speeding up the dashboard -- this focuses on the supervisor table and cuts about a second from the load time. There is still the volunteer table and some tweaking on the supervisors but I wanted to get this up for @armahillo to see.

### DIFF
--- a/app/views/dashboard/_supervisors_table.html.erb
+++ b/app/views/dashboard/_supervisors_table.html.erb
@@ -23,15 +23,12 @@
         <td id="name">
           <%= link_to(supervisor.display_name, edit_supervisor_path(supervisor)) %>
         </td>
-        <td id="volunteer-assignments"> <%= supervisor.volunteers.count %></td>
+        <td id="volunteer-assignments"> <%= supervisor.volunteers.size %></td>
         <td id="serving-transition-aged-youth">
-          <%= supervisor.volunteers.select { |v| v.serving_transition_aged_youth? }.count %>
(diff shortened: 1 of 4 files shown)


app/views/dashboard/_supervisors_table.html.erb
```diff
@@ -23,15 +23,12 @@
...
         <td id="serving-transition-aged-youth">
-          <%= supervisor.volunteers.select { |v| v.serving_transition_aged_youth? }.count %>
+          <%= supervisor.volunteers.includes(:casa_cases).where(casa_cases: {transition_aged_youth: true}).size %>
         </td>
         <td id="no-contact">
-          <%=
-            days_since_contact = 14
-            supervisor.volunteers.select { |v| v.recent_contacts_made(days_since_contact) < 1 }.count
-          %>
+          <%= supervisor.volunteers.includes(:case_contacts).where(case_contacts: {contact_made: true}).where.not(case_contacts: { occurred_at: 14.days.ago..Date.today}).size %>
```
compwron, line 31: Very nice! I'd like to eventually move query logic like this and the above into the controller and make the views less smart

compwron:
Are there any tests we can write so that n+1 queries don't come back in, or do we just have to fix it all and then uncomment `bullet.raise = true` in `config/test.rb` ?

compwron:
I want to merge this now so we can get the benefits, and we can add more on later :)

PR #623, opened 2020-08-30
Move feature specs to system

### LINKED ISSUE #614: Refactor test suite
* Refactor test suite:  Run `--profile` and seeing which tests are the slowest. 
* Move all feature tests to system tests in order to take advantage of screenshots for failed tests.

littleforest:
Resolves #614

### What changed, and why?
* Moved feature tests to system specs in order to more easily be able to debug feature tests by taking advantage of screenshots and watching tests run in the browser.
* Set up `rack_test` driver to run one test without js that was having trouble closing the modal.
* Added id field to the "Pick Displayed Columns" form to improve accessibility (and also so boxes could be checked in system spec, but couldn't get modal to close in timely fashion, so ultimately not needed for that.)
* Other minor refactoring along the way.

### Feelings gif (optional)
alt text

This PR is not a huge win, pretty much a wash in terms of spec suite running time. I personally like having all the FE specs in `spec/system` as this is the latest from Rails, and I like the consistency of just having system specs and not feature specs as it gets confusing where to put front-end specs.  I also like the ability to get the screenshots when the spec fails for debugging purposes. But absolutely OK to not merge this if it doesn't seem worth it.


spec/system/admin_views_dashboard_spec.rb
```diff
@@ -46,17 +48,16 @@
...
 
     visit root_path
 
-    expect(page).to have_text(volunteer.most_recent_contact.occurred_at.strftime("%B %-e, %Y"))
-    expect(page).to have_text("20") # miles driven
+    expect(page).to have_text("August 29, 2020")
 
     within "#volunteers" do
-      click_on volunteer.most_recent_contact.occurred_at.strftime("%B %-e, %Y")
+      click_on "August 29, 2020"
```
compwron, line 60: This seems date-based and likely to fail in the future

littleforest: I used the time travel code: `travel_to Time.zone.local(2020,8,29,4,5,6)`, so the date will be fixed. There had been a comment on a similar spec that it had been failing locally due to timezone issues, but then I saw [this commit](731f7283) and it seems like maybe the time zone issue had already been fixed. So I can also revert this.

compwron: Ah I see it now. Sorry about that, my mistake

PR #624, opened 2020-08-30
#603 - DRYs up File Importer

### LINKED ISSUE #603: DRY up file_importer.rb

armahillo:
I extracted subclasses for both User and Case importings, and left the
basic importing logic in the base class. Adding additional imports later
on should be less painful and involve replicating one of the subclasses.

The specs themselves were updated and refactored as well. I added
additional specs to cover the additional behaviors.

Resolves #603

### What changed, and why?
The issue requested that the `FileImporter` class be DRYd up, but wasn't more specific about how. I modularized a lot of the functionalities, but kept it contained to a small area.

n/a

Added additional specs

### Screenshots please :)
n/a

### Feelings gif (optional)
How to embed:
`alt text`


compwron:
@dominiquecuevas this will impact your work, you will probably want to rebase and consider pairing with @armahillo on it :)

keithrbennett:
Given that the importer objects have state that should apply to only one import call, should we maybe make the calling interface a static method that itself instantiates and calls the instance, and even make the constructors private (by using the answer at https://stackoverflow.com/questions/1567400/how-to-make-the-class-constructor-private-in-ruby)?

app/lib/importers/file_importer.rb
```diff
@@ -0,0 +1,40 @@
...
+      {type: :error, message: "#{successful_import_message}, "\
+        "the following #{type} were not imported: #{failed_imports.join(", ")}."}
+    end
+  end
+
+  def gather_users(comma_separated_emails)
+    comma_separated_emails.split(",")
+      .map { |email| User.find_by(email: email.strip) }
+      .compact
+  end
```
keithrbennett, line 39: Would a `.sort` be worth the cost here? It could be helpful for finding email addresses in log files, in testing, etc.

app/controllers/imports_controller.rb
```diff
@@ -18,11 +18,11 @@ def create
...
-      FileImporter.new(file, org).import_volunteers
+      UserImporter.new(file, org).import_volunteers
     when "supervisor"
-      FileImporter.new(file, org).import_supervisors
+      UserImporter.new(file, org).import_supervisors
     when "casa_case"
-      FileImporter.new(file, org).import_cases
+      CaseImporter.new(file, org).import_cases
     else
       {type: :error, message: "Something went wrong with the import, did you attach a csv file?"}
```
keithrbennett, line 27: The error is that a bad `import_type` has been passed, right? How about something like this for an error message?:

```
valid_import_types_string = %w{volunteer supervisor casa_case}.inspect
message = "Bad import type '#{import_type}'. Must be one of #{valid_import_types_string}."
```

app/lib/importers/file_importer.rb
```diff
@@ -0,0 +1,40 @@
+class FileImporter
+  require "csv"
+
+  attr_reader :import_csv, :org_id, :number_imported, :failed_imports
```
keithrbennett, line 4: Could we rename `import_csv` to something like `csv_import_file` or `csv_import_filespec`?

app/controllers/imports_controller.rb
keithrbennett: Or did we want to test that the provided file is there?:

```
unless File.file?(file)
  return { type: :error, message: "CSV import file not found: #{file}" }
```

app/lib/importers/file_importer.rb
keithrbennett: Or maybe we don't need `import` in the name at all, it's easy to infer.

app/lib/importers/file_importer.rb
compwron: I am in favor of sorting. Keep in mind that it should happen after teh `.compact` because otherwise sorting nils will error. 

Related- we need more tests around showing good errors for import failures caused by bad data or bad files.
Also, there's an issue open somewhere for "make errors not fade away unless you refresh the page etc" that I think would be useful for testing this

jmkoni:
Don't forget to rebase 😄

compwron:
I want to merge this, it needs rebasing tho :)

compwron:
I tried to resolve conflicts. I may have broken something. I might undo the merge in a bit

Also there are unaddressed question from Keith that need digging into :)

app/controllers/volunteers_controller.rb
```diff
@@ -58,6 +60,11 @@ def deactivate
 
   private
 
+  def _get_available_casa_cases
```
jmkoni, line 63: Why are we starting this method with an underscore? If it's to show that it's private, this just isn't common practice in Ruby, so I don't think we should do this.

jmkoni:
We also just spent 4 hours fixing a bug that this brings back. I would personally appreciate if this doesn't get merged without extremely thorough testing that ensures no change in functionality.

jmkoni (changes requested):
We need to fully test all the cases before merging.

jmkoni (approved):

compwron:
WOOOOOOOOOOOO!!!!!!! ✨✨✨✨✨✨✨✨✨✨✨✨✨✨

compwron:
Many thanks to @keithrbennett @armahillo @jmkoni & all for making this work!

Issue #597, opened 2020-08-26
Associate all objects with a casa_org_id


Zrrrpy:
Part of epic #530

**Description**
This work should be broken into two parts:

- [ ] Identify the places where objects are not associated with `casa_org_id`
- [ ] Associate all things with `casa_org_id` that currently are not. 

Example: every time a domain object is created or called with a query, every single one of those needs to be inspected to see if it is associated with a `casa_org_id`. i.e. `volunteer.create` `volunteer.where`.

abachman:
tl;dr - I don't think we need to add additional casa_org_id fields to anything in the system. We're in a good spot for multi-tenancy w.r.t. who owns what. 

 Right now there are two models associated with CasaOrgs: **User** (which includes Admin, Supervisor, and Volunteer) and **CasaCase**.

Of the the other three models in the system, **CaseContact**, **CaseAssignment**, and **SupervisorVolunteer**, two (CA and SV) are join tables between org-owned models and the third (CC) is only ever viewed in the context of a User or a Case. That means that once CasaOrg is part of the request cycle (e.g., specified by the `current_user` or as part of a subdomain or URL scheme), then all record look-ups will/can/should start with a User or Case and be scoped from there.

For User and Case listing routes, which is only `DashboardController#show`,  CasaOrg should be the parent class from which all users and cases are listed. 

Only CaseContact has listing routes (`DashboardController#show` and `CaseContactReportsController#index`) and those should be scoped to Cases belonging to the current CasaOrg, getting rid of the CaseContact-based lookups. 

Bottom line, adding casa_org_id to the other three models mentioned would be redundant given their existing associations and require extra code be added to every record creation statement. It can be added later if it proves necessary for a new reason, but the way records are used now, we can protect access by scoping all lookups to User or CasaCase.

PR #634, opened 2020-09-03
Ideas for multitenancy [work in progress]

### LINKED ISSUE #606: All users should only be able to access data associated with their casa_org
Part of epic #530

**What type of user is this for? [volunteer/admin/supervisor/all or All CASA Admin]**
all

**Description**
When a user logs in, they should only be able to access data associated with their `casa_org_id`.

Right now, there are two possible `casa_org_id`s objects can be associated with: `other_casa` or `pg_casa`. 
When an admin associated with `other_casa` logs in, they should not see any data associated `pg_casa`, and vice versa.

Dependent on: #597

abachman:

Resolves #606 and probably #597

### What changed, and why?

Makes sure that all data access accounts for the organization (CasaOrg) to which the user and resource belong.

**NOTE:** only User and CasaCase have a `belong_to` relationship with CasaOrg. CaseContact records do not, but I rewrote all the CaseContact lookups to be scoped to cases or users, which both makes more sense and is simpler at the time of record creation.

This update required edits to a _bunch_ of tests to make sure that the `create` statements in the test suite associated records with a specific organization.

### Other notes

This is a fairly complete _logical_ implementation of multi-tenancy, but it's mostly about making sure existing logic doesn't cross the streams w.r.t. organizations. There's a lot of polish needed before the app is properly multi-tenant. 

I also left some pending tests as TODOs / possible future stories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 /tmp/*
 !/log/.keep
 !/tmp/.keep
+.DS_Store
 
 # Ignore uploaded files in development.
 /storage/*
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,9 +1,12 @@
 class ApplicationController < ActionController::Base
   include Pundit
+  include Organizational
+
   protect_from_forgery
   before_action :set_paper_trail_whodunnit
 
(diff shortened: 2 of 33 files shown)


compwron (approved):
I LIKE it!!! ✨😲✨💖✨📈✨👏🏼✨

PR #643, opened 2020-09-04
bundle exec standardrb --fix

compwron:
bundle exec standardrb --fix


compwron:
merging my own PRs
like one does
so hard to quit

PR #650, opened 2020-09-05
update standard to 0.6.0 and fix errors

### LINKED ISSUE #644: fix&test contact_made_chosen model validation (early return means error is never set)

jmkoni:
Resolves #644 and updates standard to most recent version.

### What changed, and why?
Had to update standard because we are a few versions behind. Found a bug #644 and fixed that too.

Not at all.

Most was style issues, added a test for `contact_made_chosen`


app/decorators/case_contact_decorator.rb
```diff
@@ -47,7 +47,7 @@ def contact_types
   end
 
   def report_contact_types
-    object.contact_types&.map { |ct| ct.humanize.titleize }.join("|")
+    object.contact_types&.map { |ct| ct.humanize.titleize }&.join("|")
```
compwron, line 50: This should never be null in a saved case contact, I want to fix the underlying issue if this is causing issues

spec/models/case_contact_spec.rb
```diff
@@ -65,4 +65,10 @@
     expect(case_contact).not_to be_valid
     expect(case_contact.errors[:base]).to eq(["Must enter miles driven to receive driving reimbursement."])
   end
+
+  it "validates that contact_made cannot be null" do
```
compwron, line 69: Yay test!

compwron (approved):
💖✨📈

app/decorators/case_contact_decorator.rb
jmkoni: The issue was that we are doing the `&.map` but then calling join (on something that supposedly might not exist. Should we take both `&` out instead?

jmkoni: OMG you are so quick 😂

PR #660, opened 2020-09-05
Adds "Return to dashboard" button

### LINKED ISSUE #635: add return to dashboard button to all create new user pages
Part of epic #4

admin

On all create new user pages (i.e. create new volunteer, create new supervisor, create new CASA Case)

**Description**
To the right of the "create user" button, there should be a different colored button that says "return to dashboard" which brings an admin back to their dashboard. 

This button should replace any _Back_ button/hyperlinked text that may be on the page.

**Screenshots**
This is what it should look like:
Screen Shot 2020-09-02 at 7 04 20 PM (2)

armahillo:
Resolves #635 

### What changed, and why?
This satisfies the request of the issue to add "Return to Dashboard" button in place of the "Back" links.

Additionally, it:
 - Creates a new helper file (`UiHelper`) that defines a `return_to_dashboard_button` helper. (This will reduce future editing should the behavior or classes need to change)
 - Creates a new SessionHelper for RSpec to mediate sign-ins, and applies those to the affected specs
 - Creates some new View specs to test this feature

n/a

Tests are written!

We used View specs because there were already view specs created, and TBH it would be overkill to test it with a full system test.

I also tested it manually

### Screenshots please :)
spin up your server and click these links:
 - http://localhost:3000/casa_cases/new
 - http://localhost:3000/case_contacts/new
 - http://localhost:3000/volunteers/new
 - http://localhost:3000/supervisors/new

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/helpers/ui_helper.rb
+++ b/app/helpers/ui_helper.rb
@@ -1,0 +1,5 @@
+module UiHelper
+  def return_to_dashboard_button
+    link_to 'Return to Dashboard', root_path, { class: "btn btn-info pull-right" }
+  end
+end
--- a/app/views/casa_cases/new.html.erb
+++ b/app/views/casa_cases/new.html.erb
@@ -23,7 +23,7 @@
 
       <div class="actions">
         <%= form.submit "Create CASA Case", class: "btn btn-primary" %>
-        <%= link_to 'Return to Dashboard', { controller: "dashboard", action: :show}, { class: "btn btn-info pull-right" } %>
+        <%= return_to_dashboard_button %>
       </div>
     <% end %>
   </div>
--- a/app/views/casa_cases/new.html.erb
+++ b/app/views/casa_cases/new.html.erb
@@ -23,9 +23,10 @@
 
       <div class="actions">
         <%= form.submit "Create CASA Case", class: "btn btn-primary" %>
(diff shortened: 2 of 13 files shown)


app/helpers/ui_helper.rb
```diff
@@ -0,0 +1,5 @@
+module UiHelper
+  def return_to_dashboard_button
+    link_to 'Return to Dashboard', root_path, { class: "btn btn-info pull-right" }
+  end
+end
```
compwron, line 5: Missing line ending, might want to make your editor not do this 
Next time we do a linting pass this will get updated, no worries

app/views/casa_cases/new.html.erb
```diff
@@ -23,9 +23,10 @@
 
       <div class="actions">
         <%= form.submit "Create CASA Case", class: "btn btn-primary" %>
+        <%= return_to_dashboard_button %>
       </div>
     <% end %>
   </div>
 </div>
 <br>
-<%= link_to 'Back', :back %>
```
compwron, line 31: Farewell back button

compwron (approved):
Beauuuuutiful! And such tests! 💖🎉🙌🏼

PR #669, opened 2020-09-05
Add deletes to seed

compwron:
none

### What changed, and why?
https://edgeguides.rubyonrails.org/active_record_migrations.html#migrations-and-seed-data
https://blog.saeloun.com/2019/09/30/rails-6-adds-db-seed-replant-task-and-db-truncate_all.html
https://github.com/rails/rails/issues/34765

no

`rake db:seed`
`rake db:reseed`

### DIFF
--- a/db/seeds/development.rb
+++ b/db/seeds/development.rb
@@ -1,5 +1,13 @@
 require "faker"
 
+CaseContact.delete_all
+SupervisorVolunteer.delete_all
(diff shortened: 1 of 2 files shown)


compwron:
Maybe we shouldn't do this, we should use `reseed` every time? But I don't know, and I don't like being halfway in between
Happy to have this changed in the future

PR #671, opened 2020-09-05
Preserve all case contact inputs, even when invalid

### LINKED ISSUE #570: Editing case contact errors should not wipe previous selections
Part of epics #3 #105 #4

**What type of user is this for? [admin/volunteer/supervisor/all or All CASA Admin]**
all

On the edit case contact page, accessible from all user dashboards with exception to All CASA Admin.

**Description**
When a user triggers an error message on the Edit Case Contact page, sometimes the user's previous selections get wiped (not all of them, but it's not consistent). Making any changes when editing a case contact (whether or not an error is triggered) should preserve user selections, not wipe them.

This bug is not consistently replicated.

The most common way to trigger an error and replicate: 
Change the duration to 0 hours and 0 minutes, and click submit.

This bug will probably require testing on the entire Edit Case Contact page to help replicate consistently and identify bug.

**Screenshots**
None yet, but if you are able to replicate, please take a screenshot and add here!

alindeman:
Resolves #570

### What changed, and why?
* Remove helpers that were not used, rename some others for clarity
* Make sure all input is preserved if we are forced to re-render the case contact form because a validation fails

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

I enhanced a few tests to try to cover this area.


compwron (approved):
Very nice! Deleting code is delightful
And testssss
✨

PR #674, opened 2020-09-05
Fixes bullet n+1 on casa cases.

### LINKED ISSUE #519: Fix bullet n+1 issues in tests
In `test.rb` uncomment the line `Bullet.raise = true` and run the tests `rake rspec`
You will see failures. 
Fix the failures (one PR per test is fine&good). 
When there are no failures, leave the line `Bullet.raise = true` not commented and make a PR for that also

mecastelom:
Resolves #519
Resolves N+1 bullet notifications on casa_cases

### What changed, and why?
bullet was raising a n+1 error on case_controller#index.
In order to fix the error, an includes on casa_cases when pulling the organization was added when the current user is not a volunteer.

No permissions were affected

Running bullet the error in that action is not raised.

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/controllers/casa_cases_controller.rb
+++ b/app/controllers/casa_cases_controller.rb
@@ -6,7 +6,14 @@ class CasaCasesController < ApplicationController
   # GET /casa_cases
   # GET /casa_cases.json
   def index
-    @casa_cases = policy_scope(current_organization.casa_cases)
+    org_cases = if current_user.volunteer?
+                  current_organization.casa_cases
+                else
+                  # to avoid N + 1 error raised by bullet
+                  CasaOrg.includes(:casa_cases).find_by(id: current_user.casa_org_id).casa_cases
+                end
+
+    @casa_cases = policy_scope(org_cases)
   end
 
   # GET /casa_cases/1
@@ -67,7 +74,6 @@ class CasaCasesController < ApplicationController
   end
 
   private
-
   # Use callbacks to share common setup or constraints between actions.
   def set_casa_case
     @casa_case = current_organization.casa_cases.find(params[:id])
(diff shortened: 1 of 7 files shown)


app/controllers/casa_cases_controller.rb
```diff
@@ -6,7 +6,14 @@ class CasaCasesController < ApplicationController
   # GET /casa_cases
   # GET /casa_cases.json
   def index
-    @casa_cases = policy_scope(current_organization.casa_cases)
+    org_cases = if current_user.volunteer?
+                  current_organization.casa_cases
+                else
+                  # to avoid N + 1 error raised by bullet
+                  CasaOrg.includes(:casa_cases).find_by(id: current_user.casa_org_id).casa_cases
+                end
```
alindeman, line 14: @mecastelom It's not clear to me why this conditional (`if`/`else`) is needed. Does the user being a volunteer affect what queries are run for this controller/view? I looked at the corresponding view, but couldn't see anything offhand.

mecastelom: @alindeman There is casa_case_policy 2 that avoid the n+1 issue for volunteers.

alindeman: @mecastelom Gotcha. Do you think the `includes` could be moved into the policy? That might be a more centralized place to fix it.

alindeman: Or alternatively, would it be OK to add the `includes` to both cases here, removing the conditional? I'm mostly concerned that this looks a lot like policy being enforced in the controller, which we generally don't like. It's not _actually_, but it looks a lot like it and it might accidentally evolve into that over time if we're not careful. The branching logic also adds an additional conditional that must be maintained and tested.

compwron:
This looks good but it's still in "draft" status- do you want me to review it yet? :)

jmkoni (approved):
Looks great! Are all of these tested?

compwron (approved):

compwron:
Once we've fixed the bullet issues, we can turn on Bullet.config.raise = true for the "test" environment to prevent any new ones from coming up in tested code (and hall code should be tested code 😅

PR #681, opened 2020-09-06
fixes bug that defaults contact made to false

### LINKED ISSUE #673: [BUG] new case contact contact made defaults to false. It should not have a default
[BUG] new case contact contact made defaults to false. It should not have a default
Please also write a test for this to make sure it doesn't break again (for all no-default fields)
This happened while logged into staging with volunteer2

Screen Shot 2020-09-05 at 2 52 57 PM

jmkoni:
Resolves #673

### What changed, and why?
a change was previously made that checked "No" for contact made if `!case_contact.contact_made`. This returns true if `contact_made` is `nil`, so it was effectively defaulting to false. Added back in code and tests that checks if the value is actually set to false. We could potentially simplify this and remove the helper, but that makes it harder to test, so I'm in favor of keeping the helper.

No

Added a system spec that validates that, if it's not selected, it doesn't get automatically selected on submit.

### DIFF
--- a/app/helpers/case_contacts_helper.rb
+++ b/app/helpers/case_contacts_helper.rb
@@ -8,6 +8,10 @@ module CaseContactsHelper
     case_contact.duration_minutes.to_i.remainder(60)
   end
 
+  def set_contact_made_false(case_contact)
+    case_contact.persisted? && case_contact.contact_made == false
+  end
+
   def render_back_link(casa_case)
     return send_home if !current_user || current_user&.volunteer?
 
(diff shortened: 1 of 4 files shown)


app/helpers/case_contacts_helper.rb
```diff
@@ -8,6 +8,10 @@ def duration_minutes(case_contact)
     case_contact.duration_minutes.to_i.remainder(60)
   end
 
+  def set_contact_made_false(case_contact)
+    case_contact.persisted? && case_contact.contact_made == false
```
alindeman, line 12: Thanks for catching this. The `persisted?` behavior was subtle, so I thought it was unnecessary when I removed it. I was obviously wrong :)

compwron (approved):
Whew! And excellent tests :) 💖

PR #684, opened 2020-09-06
update all casa admin sessions/new to match main login

### LINKED ISSUE #683: Make AllCasaAdmin sign in page look like "normal" sign in page

/all_casa_admins/sign_in still looks like Devise defaults. Fix that.

abachman:
Reused devise/sessions/new erb on the all_casa_admins/sessions/new page. 

Fixes #683

Added positive and negative sign in tests (can sign in as, can't sign in as) for all three `User` derived and the `AllCasaAdmin` user types.


compwron (approved):
Shinyyyyyyyy ✨✨✨

PR #688, opened 2020-09-06
Volunteer without name set cannot be reached on case view

### LINKED ISSUE #678: Volunteer without name set cannot be reached on case view
Volunteer without name set cannot be reached on case view
Note how the volunteer without a name set can't be clicked into. Let's make the name here default to email if its not available. something like `volunteer.display_name || volunteer.email` - we might have a helper for this already? I forget
We could also hide the email column, it's not needed here imo

Screen Shot 2020-09-05 at 3 37 21 PM

ErinClaudio:
Resolves #678 

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ yarn-debug.log*
 coverage/
 .idea/
 config/credentials.yml.enc
+vendor/bundle/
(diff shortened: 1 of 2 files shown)


compwron (approved):
🙌🏼 Woot woot woot! 📈✨

Issue #662, opened 2020-09-05
Report filters


Zrrrpy:
Part of epic #529

**What type of user is this for? [admin/volunteer/supervisor/all or All CASA Admin]**
admins and supervisors

**Where does/should this occur**
On the <a href="https://casa-r4g-staging.herokuapp.com/reports">Generate Reports page.</a>

**Description**
In addition to "starting from" and "end date" filters for reports, users should be able to filter by any combination of the following:
[ ] Assigned to supervisors (choose from multi-select dropdown)
[ ] Volunteers a.k.a `created_by` (choose from multi-select dropdown)
[ ] Contact type (choose from multi-select dropdown)
[ ] Wants driving reimbursement (two checkboxes: yes / no)
[ ] Contact made (two checkboxes: yes / no)
[ ] Transition aged youth (two checkboxes: yes/no)

jmkoni:
#698 should be ready to review. Changed a couple of things:
- contact_type has to be single select for now. We can maybe change it to multi-select but that query is hard 😰 
- wants driving reimbursement, contact made, and has transitioned will also be selects (with the option to not select). Since they are boolean values, it made more sense to just pass one or not at all.

jmkoni:
Frontend still needs to be done 😅

PR #698, opened 2020-09-06
662 add filters to case contact report

duaimei:
abstracted the filter logic in the case contact report model to make it easier to test. then added tests to the case contact report spec. successfully added one new creator filter

Resolves [#662](https://github.com/rubyforgood/casa/issues/662)

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/models/case_contact_report.rb
+++ b/app/models/case_contact_report.rb
@@ -2,11 +2,19 @@ class CaseContactReport
   attr_reader :case_contacts
 
   def initialize(args = {})
-    @case_contacts = if args[:start_date] && args[:end_date]
-      CaseContact.occurred_between(args[:start_date], args[:end_date])
-    else
-      CaseContact.all
+    @case_contacts = filtered_case_contacts(args)
+  end
+
+  def filtered_case_contacts(args)
+    return CaseContact.all if args.empty?
+    contact = CaseContact
+    if args[:start_date] && args[:end_date]
+      contact = contact.occurred_between(args[:start_date], args[:end_date])
     end
+    if args[:creator_id]
+      contact = contact.where(creator_id: args[:creator_id])
+    end
+    contact
   end
 
   def to_csv
--- a/app/models/case_contact_report.rb
+++ b/app/models/case_contact_report.rb
@@ -9,27 +9,18 @@ class CaseContactReport
     return CaseContact.all if args.empty?
 
     contact = CaseContact
-    if args.has_key?(:start_date) && args[:end_date] # contact date range
-      contact = contact.occurred_between(args[:start_date], args[:end_date])
-    end
+    contact = contact.where(supervisors: args[:supervisor_ids]) if args.has_key?(:supervisor_ids) # supervisor, as an array
     contact = contact.where(creator_id: args[:creator_id]) if args.has_key?(:creator_id) # volunteer
-    if args.has_key?(:supervisor_ids) # supervisor, as an array
-      contact = contact.where(supervisors: args[:supervisor_ids])
-    end
+    contact = contact.occurred_between(args[:start_date], args[:end_date]) if  args.has_key?(:start_date) && args.has_key?(:end_date)
     contact = contact.where(contact_made: args[:contact_made]) if args.has_key?(:contact_made) # contact made, boolean
-    if args.has_key?(:has_transitioned) # boolean, but also in the casa cases table
-      contact = contact.joins(:casa_case).where("casa_cases.transition_aged_youth in (?)", args[:has_transitioned])
-    end
-    if args.has_key?(:want_driving_reimbursement)
-      contact = contact.where(want_driving_reimbursement: args[:want_driving_reimbursement])
-    end
+    contact = contact.joins(:casa_case).where("casa_cases.transition_aged_youth in (?)", args[:has_transitioned]) if args.has_key?(:has_transitioned) # boolean, but also in the casa cases table
+    contact = contact.where(want_driving_reimbursement: args[:want_driving_reimbursement]) if args.has_key?(:want_driving_reimbursement)
     # This filter is not working properly, commenting it out for now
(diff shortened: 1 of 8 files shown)


duaimei:
@leesharma and I paired on this.

compwron:
I see that this is tagged WIP but it looks pretty done?

app/models/case_contact_report.rb
```diff
@@ -2,11 +2,19 @@ class CaseContactReport
...
-    @case_contacts = if args[:start_date] && args[:end_date]
-      CaseContact.occurred_between(args[:start_date], args[:end_date])
-    else
-      CaseContact.all
+    @case_contacts = filtered_case_contacts(args)
+  end
+
+  def filtered_case_contacts(args)
+    return CaseContact.all if args.empty?
+    contact = CaseContact
```
jmkoni, line 10: What's the benefit of doing this? Is it in case neither set of args that we are checking for exist? If so, we'd probably want to do `contact = CaseContact.all`.

spec/models/case_contact_report_spec.rb
```diff
@@ -29,4 +29,41 @@
...
+      it "returns only the volunteer with date range" do
+        volunteer = create(:volunteer)
+        create(:case_contact, {occurred_at: 20.days.ago, creator_id: volunteer.id})
+        create(:case_contact, {occurred_at: 100.days.ago, creator_id: volunteer.id})
+       
+        create(:case_contact, {occurred_at: 100.days.ago})
+        report = CaseContactReport.new({start_date: 30.days.ago, end_date: 10.days.ago, creator_id: volunteer.id})
+        contacts = report.case_contacts
+        expect(contacts.length).to eq(1)
+      end
```
jmkoni, line 66: Can we add a test for if we pass an arg that we aren't expecting? Because right now I think it would fail. Ideally... you should be able to copy the test on L42 but pass in `{foo: 'bar'}` and it should pass. I think it won't...

app/models/case_contact_report.rb
duaimei: We are in the middle of adding 5 additional filters. If none of the 7 filters are present, then there's no reason to go through the query building process, and we should just return all CaseContacts.

spec/models/case_contact_report_spec.rb
duaimei: I don't think that the model section is the appropriate place for such a test since we should be restricting the accepted parameters from the controller.

But I will keep this comment in mind as the implementation progresses.

duaimei:
> I see that this is tagged WIP but it looks pretty done?

I still have a few filters to add. Lee and I started with the query building on the backend. Do you think that it would be good to add a separate frontend ticket to add the filters on the frontend?

app/models/case_contact_report.rb
```diff
@@ -2,11 +2,26 @@ class CaseContactReport
...
+    contact = contact.where(contact_made: args[:contact_made]) if args.has_key?(:contact_made) # contact made, boolean
+    contact = contact.joins(:casa_case).where("casa_cases.transition_aged_youth in (?)", args[:has_transitioned]) if args.has_key?(:has_transitioned) # boolean, but also in the casa cases table
+    contact = contact.where(want_driving_reimbursement: args[:want_driving_reimbursement]) if args.has_key?(:want_driving_reimbursement)
+    # This filter is not working properly, commenting it out for now
+    # if args[:contact_types]
+    #   args[:contact_types].each do |contact_type|
+    #     contact = contact.where "contact_types @> ARRAY[?]::varchar[]", [contact_type]
+    #   end
+    # end
+    contact
```
littleforest, line 24: @duaimei it looks like this won't take into account multiple filters getting checked, will it?

If you need to account for multiple filters, then a couple different ways I have done this is through either chaining separate scopes or chaining scopes using `inject`.

With scopes:

```
# app/models/case_contact.rb

scope :supervisors, ->(supervisor_ids) { where(supervisors: supervisor_ids) if supervisor_ids.any? }
scope :creator, ->(creator_id) { where(creator_id: creator_id) if creator_id.present? }
scope :contact_made, ->(contact_made) { where(contact_made: contact_made) if args[:contact_made] }
...
```

```
# app/models/case_contact_report.rb

def filtered_case_contacts(arg)
  CaseContact
    .supervisors(args[:supervisor_ids])
    .creator(args[:creator_id])
    .occurred_between(args[:start_date], args[:end_date])
    .contact_made(args[:made_contact])
    ...
end
```

With `inject`:

```
# app/models/case_contact.rb

def self.filtered_results(args)
    args.delete_if { |k, v| v.blank? }
    args.delete_if { |k, v| v.kind_of?(Array) and v.reject(&:empty?).empty? }
    args.inject(self.all) do |scope, (key, value)|
      case key
      when "supervisors_id"
        scope.where(supervisors: value)
      when "creator_id"
        scope.where(creator: value)
      when "has_transitioned"
        scope.joins(:casa_case).where("casa_cases.transition_aged_youth IN (?)", value)
      when 
        ...
      else
        scope
      end
    end
  end
end
```

jmkoni: This will account for multiple filters :) it's building up one long query.

jmkoni: Though I like your idea of moving these to scopes... 🤔

littleforest: Is that method getting called multiple times? I just wasn't seeing how it's building up to a single query if `contact` is getting redefined on each line...? But I am probably missing something!

littleforest (approved):
🎉 Looks good to me!

compwron (approved):
🙌🏼✨🙌🏼✨🙌🏼✨🙌🏼✨🙌🏼✨🙌🏼✨

PR #699, opened 2020-09-06
Make erblint+rubocop more conservative by default, fix long lines

### LINKED ISSUE #602: Make linter enforce line length on ERB files
Related to https://github.com/rubyforgood/casa/issues/489

`bundle exec erblint --lint-all --enable-all-linters --autocorrect` (see readme)
https://github.com/Shopify/erb-lint which uses https://github.com/Shopify/better-html

alindeman:
Resolves #602

### What changed, and why?
When enabled, `erb-lint` runs `rubocop` against the Ruby portions of ERB files. As we saw in #672, this results in fairly aggressive changes (e.g., `frozen_string_literal` in every Ruby section) and some changes that just completely break the ERB syntax. The idea of running Rubocop against the Ruby in ERB is interesting, but it seems like we need to be careful which cops are enabled if it's going to be successful.

This PR proposes we switch to an allowlist of Rubocop cops for ERB, starting with `Layout/LineLength` only.

Fix `Layout/LineLength` where possible by splitting lines at reasonable locations, or refactoring long lines away into helpers or otherwise.

None

The existing tests continue to pass.


compwron (approved):
This is great! Thank you for working through this mess, this makes it SO much better 😅🙌🏼✨🙌🏼🎉

app/views/all_casa_admins/sessions/new.html.erb
```diff
@@ -31,7 +31,7 @@
 
           <%= render "all_casa_admins/shared/links" %>
 
-          <br/>
+          <br>
```
keithrbennett, line 34: I'm curious, why is it better to omit the slash here?

compwron: This is a default of the linting library. I recommend reading the linting library docs for information about the choices it makes. https://github.com/Shopify/erb-lint#SelfClosingTag

alindeman: In HTML5, it's not necessary to use self-closings tags for void elements. `erb-lint` [made this change based on HTML5 style](https://github.com/Shopify/erb-lint#SelfClosingTag)

PR #712, opened 2020-09-10
534 keep errors visible

### LINKED ISSUE #534: Red error messages should not disappear until the page is refreshed
**Part of epics:** #4, #105, #3

all

Anywhere on the frontend where a red error message displays.

**Description**
Current behavior: red error messages disappear after a few seconds. This assumes how much time it takes the user to read the error message. 

Desired behavior: red error messages should not disappear until the page is refreshed.

abachman:

Resolves #534 

### What changed, and why?

Turned off the flash message alert hiding altogether. If users end up needing it, then it would be okay to turn it back on.

None tests, but there weren't any to begin with 🤷 

### Screenshots please :)

image

stays that way, all day, or ur 💰  back

### DIFF
--- a/app/javascript/src/shared.js
+++ b/app/javascript/src/shared.js
@@ -1,0 +1,6 @@
+// /* global $ */
+// window.setTimeout(function () {
+//   $('.alert')
+//     .not('.alert-danger')
+//     .alert('close')
+// }, 5000)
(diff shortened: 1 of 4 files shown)


app/javascript/src/shared.js
```diff
@@ -0,0 +1,6 @@
+// /* global $ */
```
compwron, line 1: Let's delete code instead of commenting it out

compwron (approved):
Woot woot :) ✨

PR #717, opened 2020-09-11
420 disallow contact edit past time

### LINKED ISSUE #420: Disallow case contact edit after last day of month in each quarter
**What**
Don't allow editing a case contact after the last day of the month in the quarter that the case contact was created in. So if a case contact was created on the last day of March, then it will be no longer editable on April 1. 
If a case contact is created on Jan 1, it will be editable until April 1 (of the same year it was created- don't forget about years!)
No user (volunteer, supervisor, admin) should be allowed to edit or delete a case contact after the last day of the month in each quarter. 
Bonus: add a tooltip in place of the edit button for contacts that the user should be able to edit, if it weren't for the cutoff (i.e. Volunteer A should only see the tooltip for contacts User A made, not for contacts on the same case the Volunteer B made- because Volunteer A wouldn't be able to edit the other volunteer's case contacts anyway. But a supervisor should see the tooltip for all case contacts which they are not allowed to edit, because they are allowed to edit any case contact)

**Why?**
From Ann Marie: I think we could do retroactive and it doesn't need to be a brief as 24 hours. It could be a few weeks or it could be on the final day of the month for each quarter (September, December, March and June). Whatever works as long as at some point, it locks so I don't misreport data down the road.

Screen Shot 2020-07-23 at 8 30 21 AM

Garbar:
Resolves #420 

### What changed, and why?
Disallow case contact edit after the last day of the month in each quarter
But from bonus added the only tooltip with the reason of the disallow on editing(I'm not sure what text to add here)

- Volunteer permissions: Yes
- Supervisor permissions: Yes
- Admin permissions: Yes

Added tests


app/models/case_contact.rb
```diff
@@ -65,6 +66,20 @@ def contact_made_chosen
     errors[:base] << "Must enter whether the contact was made." if contact_made.nil?
     !contact_made.nil?
   end
+
+  def allowed_edit?
+    today = Time.zone.now
+    return false if occurred_at.end_of_quarter < today
+    
+    true
+  end
```
littleforest, line 75: Since `occurred_at.end_of_quarter < today` will return `true` or `false`, you can just return that statement:

```
def allowed_edit?
  today = Time.zone.now
  occurred_at.end_of_quarter < today
end
```

littleforest: Though whoops, my operator sign is backwards!

```
def allowed_edit?
  today = Time.zone.now
  occurred_at.end_of_quarter > today
end
```

compwron:
Very nice!
I'm going to hold off on approve/merge for this for a few hours in the hopes of letting #624 get sorted out, but I'll merge everything tomorrow saturday morning before r4g starts up at the latest :)

compwron (approved):
✨✨✨✨✨🙌🏼

PR #720, opened 2020-09-12
standard fixes

jmkoni:
### What changed, and why?
Ran `standardrb --fix`, autofixed most, removed a couple of unused variables.

Not at all.

Ran the test suite.


app/lib/importers/file_importer.rb
```diff
@@ -16,7 +16,7 @@ def import
     CSV.foreach(csv_filespec || [], headers: true, header_converters: :symbol) do |row|
       yield(row)
       @number_imported += 1
-    rescue StandardError => e
```
compwron, line 19: interesting, I wonder why it did this

spec/lib/importers/user_importer_spec.rb
```diff
@@ -11,20 +11,20 @@
     let(:volunteer_importer) { -> { UserImporter.import_volunteers(import_file_path, casa_org_id) } }
 
     it "imports volunteers from a csv file" do
-      expect { volunteer_importer.() }.to change(User, :count).by(3)
+      expect { volunteer_importer.call }.to change(User, :count).by(3)
     end
 
     it "returns a success message with the number of volunteers imported" do
-      alert = volunteer_importer.()
```
compwron, line 18: I like `.()` better but `.call` is easier to google for sure

compwron (approved):

app/lib/importers/file_importer.rb
abachman: looks like an unused argument issue. If e is never used, it's better left out

abachman (approved):

PR #726, opened 2020-09-12
691 no empty display name or email

### LINKED ISSUE #690: Admin can not create or save volunteer account without a display name
Admin can not create or save volunteer account without a display name
If display name is empty, there should be an error on the page requesting that name be set

### LINKED ISSUE #691: Admin can not import volunteers without display names

strangrjrjr:
Resolves #690 
Resolves #691 

### What changed, and why?
- Added validations in the User model to require the presence of a display name and email address for all user types
- Updated test framework to include display names
- Updated system spec to verify the correct error messages are displayed

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

See above: added testing to validate correct errors being displayed when attempting to create a volunteer with empty fields

### Screenshots please :)
Screen Shot 2020-09-12 at 9 26 29 AM

### Feelings gif (optional)
How to embed:
`alt text`


strangrjrjr:
Hmm, something isn't right here...

compwron:
I think these are valid test failures. Several parts of the UI default to email when display name is unavailable, and the line `sequence(:display_name) { |n| "User #{n}"}` makes the display name always set. So you probably need to update several tests :)

strangrjrjr:
Is it an acceptable solution to require a display name? I didn't want to introduce a bunch of rework

compwron:
@strangrjrjr That is an ok change to make, we don't have no-display-name volunteers in prod. We do in staging, but so please do fix that before deploying (either manually via the UI, or via a smart db migration) or the deploy will fail :)

strangrjrjr:
@compwron Gotcha, we're working on getting green lights locally; I'll probably have a question or two about how to successfully deploy to staging :) Thanks!

littleforest:
@strangrjrjr I added the [After Party gem](https://github.com/theSteveMitchell/after_party) so you can add one-off rake tasks that need to be run after deployment to populate data. I prefer that than using db migrations since you end up with a broken migration if the underlying model or database field is ever changed.

compwron:
> @strangrjrjr I added the [After Party gem](https://github.com/theSteveMitchell/after_party) so you can add one-off rake tasks that need to be run after deployment to populate data. I prefer that than using db migrations since you end up with a broken migration if the underlying model or database field is ever changed.
> 
> But if you do use after party, add a note in the PR to run `heroku run rake after_party:run` after deployment (since it's not automatic yet).

@littleforest I don't love even afterparty because it doesn't run until I remember to go run it and it's easy to forget when going from staging to production. I get the point about delicate migrations though. I would rather write a very safe migration than require an extra deploy step but I'll merge it either way, just gotta remember when deploying :)

strangrjrjr:
Thanks for the help @littleforest. I'll start reading up on Smart Migrations and see if I can get one written. Perhaps After Party could be a fallback option? 

As a side note, I think some of our tests for the view are not properly returning the existence of certain values and elements on the page. @keithrbennett and I have been piping html to the console and things are not adding up, especially with tests for `spec/views/*/new.html.erb_spec.rb`

littleforest:
>  I don't love even afterparty because it doesn't run until I remember to go run it and it's easy to forget when going from staging to production

@compwron yes, I hear you! How/where did you set up Heroku to run the migrations by default? I've added `rake after_party:run` just as the line after `rails db:migrate` for other projects.

compwron:
@littleforest aha, migrations get auto-run in the Procfile

strangrjrjr:
Ready for review!

compwron (approved):

compwron:
This is BIG! Very very cool :) :) ✨🎉✨

PR #730, opened 2020-09-12
Add tests for user redirection on authz error

### LINKED ISSUE #536: add test: load all endpoints while logged in as volunteer, should see error/redirect instead of exception for unvisitable pages
If I am signed in as a volunteer, I expect to be redirected to the homescreen if I try to visit a url path that I do not have permission to visit. For example: https://casa-r4g-staging.herokuapp.com/supervisors/new 

I believe that this is an issue that can be resolved with devise and a before_filter in the controllers.
An example: https://stackoverflow.com/questions/23555618/redirect-to-log-in-page-if-user-is-not-authenticated-with-devise

alindeman:
Resolves #536

### What changed, and why?
Adds a couple of high level tests that users see friendly error message when they encounter an authz error. More scenarios can be added if there are unique situations or coverage is desired.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

Via system tests

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,21 +14,21 @@ class ApplicationController < ActionController::Base
   def must_be_admin
     return if current_user&.casa_admin?
 
-    flash[:notice] = "You do not have permission to view that page."
+    flash[:notice] = t("default", scope: "pundit")
     redirect_to root_url
   end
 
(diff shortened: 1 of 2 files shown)


compwron (approved):
Very nice! ✨

PR #741, opened 2020-09-13
Add after_party:run to Procfile

littleforest:
### What changed, and why?

Add `after_party:run` rake task to Procfile so it will run automatically on Heroku.

But @compwron if you are not a fan of After Party for other reasons, happy to move away from it.  I just feel like it is safer than having brittle migrations down the line.


compwron (approved):
I like it! Thank you for automating this, I wish it had occurred to me sooner 😅🙌🏼

PR #767, opened 2020-09-14
Use letter_opener instead of mailcatcher

### LINKED ISSUE #754: Mail was not delivered to Mailcatcher when using Docker
* Start up Docker using instructions in DOCKER.md
* Login as a CASA admin
* Create a new supervisor
* Expected: The supervisor invite email should be available at http://localhost:1080, but it is not

It looks like Mailcatcher is supposed to be hooked up correctly, including for Docker, but when pairing with @dominiquecuevas we did not see mail being delivered there.

littleforest:
Resolves #754 

### What changed, and why?
The `letter_opener` gem had been set up in April, but then the `mailcatcher` gem also got added in July. However, the mail configuration was still set up for `letter_opener` so Mailcatcher was never actually working at all. Removed all references to mailcatcher.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -67,10 +67,6 @@ $ docker-compose exec database psql -h database -Upostgres casa_development
 $ docker-compose exec web rails c
 ```
 
-## Mailcatcher
-We use [Mailcatcher](https://mailcatcher.me/) to receive mail in development.
-All mail sent in the development environment can be viewed at http://localhost:1080.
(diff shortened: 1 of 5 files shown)


alindeman:
Could there be a line or two added to the README about how to use `letter_opener`? Or a link to the `letter_opener` docs?

compwron (approved):
🙌🏼✉️✨

PR #777, opened 2020-09-16
Minor documentation update for Docker

littleforest:
### What changed, and why?

Don't need special instructions about copying test failure screenshots out of container since the tmp directory is not a volume for this Docker setup.

Added some more general docker-compose commands.


compwron (approved):
:) :)

PR #782, opened 2020-09-17
Update miles driven to save as 0 if left empty. Task to update curren…

### LINKED ISSUE #768: [BUG] Correct sorting of case contacts when miles driven is 0 rather than None
This is currently volunteer2 in staging but that data might get deleted- replicate locally instead
Screen Shot 2020-09-14 at 4 02 41 PM
Screen Shot 2020-09-14 at 4 02 46 PM

pollygee:
…t nils for miles driven to become 0

Resolves #768 

### What changed, and why?
Sorting for miles driven on the case contacts page for volunteer wasn't working correctly.
I changed the params to save miles driven as 0 if it's left blank.
I took away the emoji 😢 
I added a font awesome car to hopefully make up for it 😸  (it's not nearly as cute though)

N/A

hmmm Not sure exactly how to test this, happy to add if you have suggestions.  I can think about it more and come back to this.

### Screenshots please :)
PG_CASA

PG_CASA

### Feelings gif (optional)

Alt Text
sorry!  Prob not perfect, but it's a start?

### DIFF
--- a/app/decorators/case_contact_decorator.rb
+++ b/app/decorators/case_contact_decorator.rb
@@ -25,7 +25,7 @@ class CaseContactDecorator < Draper::Decorator
   end
 
   def miles_traveled
-    object.miles_driven ? "#{object.miles_driven} 🚘" : "None"
+    (object.miles_driven.nil? || object.miles_driven == 0) ? "" : object.miles_driven 
   end
 
   def reimbursement
--- a/app/controllers/case_contacts_controller.rb
+++ b/app/controllers/case_contacts_controller.rb
@@ -96,6 +96,7 @@ class CaseContactsController < ApplicationController
   end
 
   def create_case_contact_params
+    params[:case_contact][:miles_driven] = 0 if params[:case_contact][:miles_driven].nil?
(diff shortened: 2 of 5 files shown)


littleforest:
What about just having a database migration that sets the `miles_driven` field to `null: false, default: 0`? That will also automatically back-populate existing records.

app/decorators/case_contact_decorator.rb
```diff
@@ -25,7 +25,7 @@ def report_duration_minutes
   end
 
   def miles_traveled
-    object.miles_driven ? "#{object.miles_driven} 🚘" : "None"
+    (object.miles_driven.nil? || object.miles_driven == 0) ? "" : object.miles_driven
```
jmkoni, line 28: If you do what @littleforest suggests, this can become `object.miles_driven.zero?`

compwron (approved):
I agree with the migration idea *and also* I am going to merge this as-is because it will be an improvement on the current state and I want to get this cleaned up asap. I am open to a new PR which moves to a migration.

PR #785, opened 2020-09-17
add migration to set default value for case contact miles driven

### LINKED ISSUE #784: Set miles_driven field to default to 0 in database
**Description**

Currently we are manually setting the `miles_driven` field for a Case Contact to 0 if the field is left as null when the case contact form is submitted. We would like to modify this by having the database field automatically default to 0.

- [ ] Make sure a request or system spec exists that checks the following (please add a new test if none exists since ultimately the behavior will not change):
* Log in as `casa_admin1@example.com/123456`
* Click on any case number in the Case Numbers column of the Volunteers table
* Click "New Case Contact" button
* Fill out form, leaving miles driven as blank. Save form.
* Miles driven should be set to 0 (instead of null)

- [ ] Add a database migration that changes the `miles_driven` field in `case_contacts` to be `null: false, default: 0`

- [ ] Refactor any changes made in [#782](https://github.com/rubyforgood/casa/pull/782/files) that can be modified due to the database field defaulting to 0.

jmkoni:
### What changed, and why?
Follow up to #782. Instead of a bunch of validations, just set the default to zero.
closes #784 

No permissions changed

Added a test to the model to validate that you can save and it'll default to zero.

### Feelings gif (optional)
How to embed:
`alt text`


littleforest (approved):
🎉 👯‍♀️

PR #786, opened 2020-09-17
WIP -  do not merge

pollygee:
I started another branch to create the migration suggested in my last PR.

This puts 0 in the field in the UI beautifully, so that default value works great.  
However, if I manually delete that 0 out and forget to enter something else, the page breaks.   So, I'd need to add a validation I think to send an error if that field is empty, which it would only be if someone cleared it out, I still would want to handle that.  

It also breaks many tests because they don't have a value for miles_driven.

I can keep going and make these changes, but first wanted to be sure that's worth it and what we want?  Or maybe I'm missing something and went down the wrong path?

@littleforest or @compwron any input here?

Resolves #001

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/decorators/case_contact_decorator.rb
+++ b/app/decorators/case_contact_decorator.rb
@@ -25,7 +25,7 @@ class CaseContactDecorator < Draper::Decorator
   end
 
   def miles_traveled
-    (object.miles_driven.nil? || object.miles_driven == 0) ? "" : object.miles_driven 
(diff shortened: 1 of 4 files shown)


jmkoni:
I already did this in #785

pollygee:
oh ok :(  I didn't realize you were working on this

PR #811, opened 2020-09-21
Add back original logo and refactor helper method

### LINKED ISSUE #744: Complete per-CASA footer to include real info and not just the ruby for good logo

All

**Description**

We have two controller helper methods doing the same work in slightly different ways, refactor all uses of `@casa_org` in views to be `current_organization`.

roxannecojocariu:
Resolves #744 

### What changed, and why?
Replaced usage of `@casa_org` in all Views to `current_organization`.
Added back the organization's logo.
Updated `casa_org` factory to include attributes for address and links.

Changed the header and footer to reflect appropriate logo for the appropriate logged in user.
Extracted out conditional logic from the view into helper method.
Fixed/rewrote specs that failed after this effort.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

Updated associated spec tests.

### Screenshots please :)
All Casa Admin
all_casa_admin_view

Volunteer
volunteer_view

Not Logged In
not_logged_in_view

### Feelings gif (optional)
What gif best describes your feeling working on this issue?
[like this](https://thumbs.gfycat.com/CluelessBeneficialBlackfootedferret-small.gif)


ryanmrodriguez:
How about something like this?
Screen Shot 2020-09-21 at 3 43 38 PM

littleforest:
@roxannecojocariu can you also add back in the specs that were removed with #753?

littleforest:
Also, can you remove the defining of `@casa_org` in the application controller, since it's not needed anymore.

app/views/all_casa_admins/dashboard/show.html.erb
```diff
@@ -13,10 +13,10 @@
     </tr>
   </thead>
   <tbody>
-    <% @casa_orgs.each do |casa_org| %>
+    <% @current_organizations.each do |current_organization| %>
       <tr>
-        <td><%= link_to casa_org.name, all_casa_admins_casa_org_path(casa_org) %></td>
-        <td><%= casa_org.created_at.strftime('%B %e, %Y') %></td>
+        <td><%= link_to current_organization.name, all_casa_admins_casa_org_path(current_organization) %></td>
+        <td><%= current_organization.created_at.strftime('%B %e, %Y') %></td>
```
ds29, line 19: The use of `current_organization` as the variable name may lead to confusion with the application level helper of the same. If one were to search for `current_organization` in the project, we'd find this reference which is a local scope variable and NOT the helper itself.

Consider simplifying by renaming the variables to `@organizations` and `organization`.

ds29 (commented):
@roxannecojocariu @leilaweiss Nicely done! I left one minor suggestion for you but it looks great.

app/views/all_casa_admins/dashboard/show.html.erb
drborges: good call 👍

ds29 (approved):

app/views/layouts/headers/_current_user.html.erb
```diff
@@ -0,0 +1,31 @@
+<header>
```
drborges, line 1: @jasperfurniss, @coleerikson, @ds29 FYI: I think this PR will create quite a few conflicts with the `new-layout` branch.

PR #827, opened 2020-09-22
760/Admin can enter Casa Case's birth month and year

### LINKED ISSUE #760: Admin can enter CASA Case's birth month and year
### Context
- Right now, the burden is on the users to update the case from "Youth Under 14" to "Transition Age"
- We want to make this automatic in the system 
- We also don't want to put too much PII in the system, so this will only use month and year to determine the youth's age
- Open to how this field (or fields) looks. I am picturing something like "Start Year Demonstration" example [here](https://kidsysco.github.io/jquery-ui-month-picker/) but am flexible to engineering constraints. As long as we can capture month and year (and preferably in numeric form), I'm happy!

### Acceptance Criteria
**As an** admin
**I want to** add a youth's birth month and year
**So that** users don't have to manually update ages

Scenario 1: See the field
GIVEN I am an admin 
WHEN the Edit Case Details page loads
THEN I see a date picker field that accepts month and year and is labeled "Youth's Birth Month & Year"

Scenario 2: Saving
WHEN I am on the Edit Case Details page
AND I have edited the Youth Birth Month and Year
WHEN I hit "submit"
THEN the "Youth's Birth Month & Year" saves to the database

KatherineMuedas:
Resolves #760

### What changed, and why?
Adds a `birth_month_year_youth` datetime attribute to `casa cases` model and allow admins to select a month and year. 

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions: Only admins can add/edit `birth_month_year_youth`

### Screenshots please :)
Screen Shot 2020-09-22 at 4 00 34 PM
Screen Shot 2020-09-22 at 4 00 47 PM
Screen Shot 2020-09-22 at 4 01 00 PM
Screen Shot 2020-09-22 at 4 01 14 PM

### Feelings gif (optional)
alt text

### DIFF
--- a/spec/support/session_helper.rb
+++ b/spec/support/session_helper.rb
@@ -7,6 +7,10 @@ module SessionHelper
     sign_in(Volunteer.first || create(:volunteer))
   end
 
+  def sign_in_as_supervisor
+    sign_in(Supervisor.first || create(:supervisor))
+  end
+
   def sign_in_as_all_casa_admin
     sign_in(AllCasaAdmin.first || create(:all_casa_admin))
   end
--- a/spec/views/casa_cases/edit.html.erb_spec.rb
+++ b/spec/views/casa_cases/edit.html.erb_spec.rb
@@ -13,6 +13,7 @@ describe "casa_cases/edit" do
       render template: "casa_cases/edit"
 
       expect(rendered).not_to include("Assign a New Volunteer")
+      expect(rendered).not_to include(CGI.escapeHTML("Youth's Birth Month & Year"))
     end
   end
 
@@ -26,6 +27,7 @@ describe "casa_cases/edit" do
       render template: "casa_cases/edit"
 
       expect(rendered).to include("Assign a New Volunteer")
+      expect(rendered).to include(CGI.escapeHTML("Youth's Birth Month & Year"))
     end
   end
 end
--- a/app/controllers/casa_cases_controller.rb
+++ b/app/controllers/casa_cases_controller.rb
@@ -78,7 +78,7 @@ class CasaCasesController < ApplicationController
 
   # Only allow a list of trusted parameters through.
   def casa_case_params
-    params.require(:casa_case).permit(:case_number, :transition_aged_youth)
+    params.require(:casa_case).permit(:case_number, :transition_aged_youth, :birth_month_year_youth)
   end
 
   # Separate params so only admins can update the case_number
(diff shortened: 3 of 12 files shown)


ryanmrodriguez:
Is there a plan to remove the `Transition Aged Youth` checkbox? I noticed that editing a case and entering a `Youth's Birth Month & Year` that is greater than 14 years old doesn't affect the `Transition Aged Youth` checkbox, meaning that they are allowed to be out of sync. If we want to remove the checkbox, I think we'll also need to add more logic in the backend to handle displaying items such as `Assigned To Transition Aged Youth` using the datetime as a reference instead of the bool. Maybe we could clarify with @Zrrrpy ?

compwron (approved):
Very cool! Yes, now that we have this, we need to 
1. make sure that we have a plan for backfilling it for existing prod cases (not code, just data)
2. remove `transition_aged_youth` and make sure that it is fully replaced by `birth_month_year_youth` 

2a. Take special care with imports- make sure the example import files are updated and add logic for importing these dates 9and making sure to *not* import or store day of birth even if it is provided)
2b. Take special care with report exports- make sure that they don't break (there's no reason that they should imo but... just doublecheck, tests are not adequate there yet)

I am ok with doing all of the above in a separate PR as long as we don't forget! I made a placeholder issue https://github.com/rubyforgood/casa/issues/831

spec/views/casa_cases/edit.html.erb_spec.rb
```diff
@@ -13,6 +13,7 @@
       render template: "casa_cases/edit"
 
       expect(rendered).not_to include("Assign a New Volunteer")
+      expect(rendered).not_to include(CGI.escapeHTML("Youth's Birth Month & Year"))
```
compwron, line 16: CGI! :)

spec/support/session_helper.rb
```diff
@@ -7,6 +7,10 @@ def sign_in_as_volunteer
     sign_in(Volunteer.first || create(:volunteer))
   end
 
+  def sign_in_as_supervisor
+    sign_in(Supervisor.first || create(:supervisor))
```
compwron, line 11: :)

ds29 (commented):
Nice work! A couple minor suggestions but no blockers.

compwron:
In a future PR please also add this to the db seed file with some randomness to get a spread of ages

Issue #751, opened 2020-09-13
Use devise_invitable for all invite emails


alindeman:
**What is the problem, and what should happen instead?**
Right now we have two different mailers that have more or less the same content:
* `app/views/devise/mailer/invitation_instructions.{html,text}.erb`
* `app/views/volunteer_mailer/account_setup.html.erb`

The former is used by the [devise_invitable](https://github.com/scambra/devise_invitable) gem, which we use for inviting supervisors and user imports. The latter is our own mailer, used only by `VolunteerMailer` and `VolunteersController`.

It would be better to decide to use `devise_invitable` for all invite emails, or not at all. The current situation is a bit confusing and duplicative.

compwron:
Hmm yeah that would be a good improvement :)

Zrrrpy:
sounds like consensus then is that we will use `devise_invitable` for all invite emails and scrap the `app/views/volunteer_mailer/account_setup.html.erb` mailer? cc @compwron (If yes I'll update the ticket to reflect this)

stevelandwehr:
In the [PR for this issue](https://github.com/rubyforgood/casa/pull/838), `app/views/volunteer_mailer/account_setup.html.erb` was not scrapped because when a volunteer is activated, the `VolunteerMailer` is being used to update the user's password token and not just sending an invitation email.

PR #838, opened 2020-09-23
751 Use DeviseInvitable for volunteer invites

stevelandwehr:
https://github.com/rubyforgood/casa/issues/751

### What changed, and why?
When creating a new volunteer, use DeviseInvitable instead of VolunteerMailer to send the invitation email. This will consolidate how users are sent invitation emails.

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

This was tested locally in development by configuring ActionMailer to send emails via smtp through Gmail; an email was received after calling `invite!` on the volunteer object.

### Screenshots please :)
image

### Feelings gif (optional)
How to embed:
alt text

### DIFF
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -12,7 +12,7 @@ class VolunteersController < ApplicationController
     @volunteer = Volunteer.new(create_volunteer_params)
 
     if @volunteer.save
-      VolunteerMailer.account_setup(@volunteer).deliver
+      @volunteer.invite!
       redirect_to root_path
     else
       render :new


app/controllers/volunteers_controller.rb
```diff
@@ -12,7 +12,7 @@ def create
     @volunteer = Volunteer.new(create_volunteer_params)
 
     if @volunteer.save
-      VolunteerMailer.account_setup(@volunteer).deliver
```
drborges, line 15: Is there any reason to keep `VolunteerMailer#account_setup` around or can we replace all instances of it with `DeviseInvitable`?

stevelandwehr: We left `VolunteerMailer#account_setup` because it's doing more than just sending an email when a volunteer is activated...the `account_setup` method is also calling `generate_password_reset_token` on the user. Sorry, I didn't mention this in the PR...I left a comment about this in the issue here: https://github.com/rubyforgood/casa/issues/751#issuecomment-697502598

drborges: Ah! That makes perfect sense, I should've checked the issue as well, thanks for the clarification 👌🏼

drborges (approved):
It might be a good idea to add some test coverage for the new behavior before shipping this.

Issue #842, opened 2020-09-23
Add route to admins from sidebar menu


Zrrrpy:
Part of epic #822

As an admin, 
I want to click on text "Admins" in the left sidebar,
and see all other admins appear in a contained wrapper,
so that I can take all actions associated with admins (create/view/edit) from the same place – without being directed to a different page.

**How does it work now?**
Admins can view a table of all other admins at the very bottom of their home dashboard. They display here by default, and don't go away.

**How should it work?**
Clicking on text "Admins" in left sidebar should prompt the admin view to appear in a contained wrapper on the same page. 

**Acceptance criteria**
Text "Admins" appears in left sidebar
Clicking "Admins" opens the Admin view in a contained wrapper.

PR #849, opened 2020-09-23
842 Add Admins link to sidebar

stevelandwehr:
https://github.com/rubyforgood/casa/issues/842

### What changed, and why?
- removed admin variables and partials from dashboard pages
- added Admins link to sidebar
- added route and page for Admins index

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Screenshots please :)
image

### Feelings gif (optional)
How to embed:
alt text


app/views/layouts/_sidebar.html.erb
```diff
@@ -9,6 +9,9 @@
       <div class="list-group list-group-flush">
         <a href="0" class="list-group-item active">Active Link Example</a>
         <a href="0" class="list-group-item">Dummy Link</a>
+        <% if current_user&.casa_admin? %>
+          <%= link_to "Admins", casa_admins_path, class: "list-group-item" %>
+        <% end %>
```
drborges, line 14: I believe `link_to_if` would work here as well:

```erb
<%= link_to_if current_user&.casa_admin?, "Admins", casa_admins_path, class: "list-group-item" %>
```

stevelandwehr: I like this much better! Thanks Diego! I had to pass an empty block to it so nothing would render if `current_user&.casa_admin?` is false. Without the empty block, the plain text "Admins" was still rendering if `current_user&.casa_admin?` is false.

drborges: you're right, my initial understanding of `link_to_if` was incorrect 🤦🏻‍♂️

app/controllers/casa_admins_controller.rb
```diff
@@ -3,6 +3,10 @@ class CasaAdminsController < ApplicationController
   before_action :set_admin, only: [:edit, :update]
   before_action :require_organization!
 
+  def index
+    @admins = policy_scope(current_organization.casa_admins).sort_by(&:email)
```
compwron, line 7: It would be nice if we could move sorting to the activerecord query layer (for performance reasons) but it's not a big deal, there will never be very many casa_admins per org- I just want to make sure that we encourage healthy patterns for bigger datasets

compwron (approved):
:)

app/controllers/casa_admins_controller.rb
drborges: Agreed, given that the coding effort is about the same, we might as well push for the patters we'd like to see in the codebase 👌🏼

drborges: @stevelandwehr do you think we can sneak that change in before merging this? We can also address this before merging `new-layout` into master.

drborges (approved):

PR #876, opened 2020-09-25
[#866] Display user role on the sidebar based off of current logged in user

### LINKED ISSUE #866: Role in sidebar menu reflects role of logged in user
As a volunteer/supervisor/admin/All CASA Admin,
when I am logged in,
I see my role reflected in the left sidebar menu,
so that I know user permissions. 

**How does it work now?**
Role text in left sidebar menu is currently _static_. 

**How should it work?**
Role text in left sidebar menu should be _dynamic_ and reflect the role type of the logged in user:
- Volunteer
- Supervisor
- Admin
- All CASA Admin

It should display like this example:
Your Role 
Volunteer

**Acceptance Criteria**
Text underneath "Your Role" in left sidebar menu is dynamic,
and reflects the role type of the logged in user.

drborges:

Resolves #866 

### What changed, and why?

The **Role** information on the sidebar will be computed based on the current logged in user's role rather than being staticly set to `Supervisor`.

No.

1. Log into Casa as all user types;
2. Verify that the Role section of the sidebar reflects the expected role of the currently logged in user.

### Screenshots please :)

Logged in as `All Casa Admin` user:

all-casa-admin

Logged in as `Volunteer` user:

volunteer

Logged in as `Casa Admin` user:

casa-admin

Logged in as `Supervisor` user:

supervisor

### Feelings gif (optional)

success

### DIFF
--- a/app/controllers/concerns/organizational.rb
+++ b/app/controllers/concerns/organizational.rb
@@ -12,7 +12,16 @@ module Organizational
     @current_organization ||= current_user&.casa_org
   end
 
+  def current_role
+    @current_role ||= if user_signed_in?
+      current_user.role
+    elsif all_casa_admin_signed_in?
+      current_all_casa_admin.role
+    end
+  end
+
   included do
     helper_method :current_organization
+    helper_method :current_role
   end
 end
--- a/app/models/all_casa_admin.rb
+++ b/app/models/all_casa_admin.rb
@@ -1,4 +1,6 @@
 class AllCasaAdmin < ApplicationRecord
+  include Roles
+
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
   devise :database_authenticatable, :recoverable, :validatable, :timeoutable
--- a/app/models/concerns/roles.rb
+++ b/app/models/concerns/roles.rb
@@ -1,0 +1,7 @@
+module Roles
+  extend ActiveSupport::Concern
+
+  def role
+    model_name.human.titleize
+  end
+end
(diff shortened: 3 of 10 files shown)


app/controllers/concerns/organizational.rb
```diff
@@ -12,7 +12,16 @@ def current_organization
     @current_organization ||= current_user&.casa_org
   end
 
+  def current_role
```
drborges, line 15: @jmkoni, @compwron, @ds29 I've added this `current_role` here to handle displaying the user's role on the sidebar based on the currently logged in user type. With that said, I'm now considering creating two different `_sidebar` partials, one for the `:user` auth scope, and another for `:all_casa_admin`, and render the correct sidebar based on the current authenticated scope. This would allow us to:

1. Keep the sidebar a little leaner by avoiding quite a few conditional rendering logic that we currently need;
2. Better isolate functionality pertaining to `:user` vs `:all_casa_admin` authentication scopes.

Any concerns with that approach?

ds29: > Any concerns with that approach?

@drborges I like the suggestion. Separating the concerns would reduce the conditionals and simplify the view to be primarily comprised of the relevant navigational items.

Were you thinking to do this as part of this PR or a separate issue? I would suggest a separate issue given that this PR solves the current issue.

ds29 (approved):
Looks great!

app/controllers/concerns/organizational.rb
drborges: Sounds like a plan 👍🏼

compwron (approved):
Very nice! And I like the tests :)

PR #885, opened 2020-09-28
Add bin/login script to automate application launching and logging in

### LINKED ISSUE #812: Suggestion: Provide ability to include user id and password in URL in development mode

This would be for all users.

**Description**

Is this already possible? If not...

In my pairing and my own work I noticed that logging in to the web site is a frequent occurrence, and given the number of times that the set of all developers logs in, it might be nice to provide an option to automate it -- but only for development and maybe test modes, not production mode.

If the user id and password could be passed as parameters, then it would be a simple matter to automate that (I volunteer to do this) by writing a little script that calls `xdg-open` for Linux, `open` for Mac OS, or `start` for Windows to launch the URL in the OS' default browser. Possibly this script could take a single character parameter to specify user type (e.g. 'v' for volunteer). The appropriate login URL would be built and launched.

Since this is a developer convenience and not included in a production system, it would be not necessary to incur the usual cost in developer time of rigorous testing and documentation.

keithrbennett:
Resolves #812

A `bin/login` script is added that automates the launching and logging in of users seeded in development mode.

This can make the developer workflow faster, eliminating the need for manual browser input of URL, user id, and password.

### DIFF
--- a/bin/login
+++ b/bin/login
@@ -1,16 +1,6 @@
 #!/usr/bin/env ruby
 
-# This script simplifies login for experimentation with the users added to the application when it is seeded
-# in development mode.
-#
-# It outputs the available users and accepts a numbered choice. It then logs in as that choice,
-# at the URL appropriate for that user.
-#
-# The browser window remains open as long as the script has not yet terminated (using Ctrl-C).
-# You can either keep a terminal with this script open, or you can send it to the background with Ctrl-Z.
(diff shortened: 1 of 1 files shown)


keithrbennett:
Here is what it looks like if you run it with no arguments:

image

compwron (approved):
Cool! :)

Issue #831, opened 2020-09-23
casa_case becomes transition_aged_youth when they turn 14


compwron:
**Description**
On the first day of a `casa_case`'s birth month in the year they turn 14 (`birth_month_year_youth` in the `casa_cases` table), 
`transition_aged_youth` should change from `false` to `true`. 

**Acceptance criteria**
- [ ] Backend: Under the conditions listed above, `transition_aged_youth` automatically becomes `true`
- [ ] Frontend: When `transition_aged_youth` is `true`, 🐛 🦋  appears on the front end with the text `Yes`

![Screen Shot 2020-09-30 at 8 06 34 PM (2)](https://user-images.githubusercontent.com/62810851/94760872-89de8880-0358-11eb-9e46-36529c03f999.png)

Zrrrpy:
I don't know that we want to remove `transition_aged_youth`, I was thinking we would keep both and the next piece of work would be the system automatically flagging a `casa_case` as `transition_aged_youth` `true` on the final day of their birth month the year they turn 14. I imagine this is preferable to birth month and year and/or age displaying on the frontend to identify as TAY because less PII.  

Thoughts @compwron?

PR #897, opened 2020-10-01
Prohibit changing of transition_aged_youth

### LINKED ISSUE #896: remove ability to edit transition_aged_youth
**Description**
Previously, it was up to users to mark a `casa_case` as `transition_aged_youth`. After #831 is done, this will be automated. So users no longer need the ability to edit this manually. This change should be made on the frontend _and_ the backend so that there is only one trigger that changes `transition_aged_youth` from `false` to `true`. 

On the Edit CASA Case page accessible by all users. (<a href="https://casa-qa.herokuapp.com/casa_cases/152/edit">Here is an example</a>. Login using: email - supervisor1@example.com // pw: 123456)

**How is it working now?**
All users can change `transition_aged_youth` from `false` to `true` and vice versa. 

**How should it work instead?**
No user should have the ability to adjust `transition_aged_youth` status for any `casa_case`.

rafaeelaudibert:
Resolves #896 

### What changed, and why?
Removed the ability to change the `transition_aged_youth` flag. I still allowed the field to be used when creating the case.

### DIFF
--- a/app/views/casa_cases/_form.html.erb
+++ b/app/views/casa_cases/_form.html.erb
@@ -20,7 +20,7 @@
         </div>
       <% end %>
 
-      <% if !casa_case.persisted? %>
+      <% if casa_case.new_record? %>
         <%# Only show the field when creating, but not when updating %>
         <div class="field form-group">
           <%= form.label :transition_aged_youth %>
--- a/app/views/casa_cases/_form.html.erb
+++ b/app/views/casa_cases/_form.html.erb
@@ -20,10 +20,13 @@
         </div>
       <% end %>
 
-      <div class="field form-group">
-        <%= form.label :transition_aged_youth %>
-        <%= form.check_box :transition_aged_youth %>
-      </div>
+      <% if !casa_case.persisted? %>
+        <%# Only show the field when creating, but not when updating %>
+        <div class="field form-group">
+          <%= form.label :transition_aged_youth %>
+          <%= form.check_box :transition_aged_youth %>
+        </div>
+      <% end %>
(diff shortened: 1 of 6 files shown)


app/views/casa_cases/_form.html.erb
```diff
@@ -20,10 +20,13 @@
         </div>
       <% end %>
 
-      <div class="field form-group">
-        <%= form.label :transition_aged_youth %>
-        <%= form.check_box :transition_aged_youth %>
-      </div>
+      <% if !casa_case.persisted? %>
```
littleforest, line 23: I think it's easier to not read negatives (despite the structure of this sentence) 😬 :

```
<% if casa_case.new_record? %>
```

littleforest (changes requested):

app/views/casa_cases/_form.html.erb
rafaeelaudibert: You are absolutely right. Changed it!

jmkoni:
Downside is  that this is dependent on #831, so we want to get a fix for that before this is merged in. But it does look good!

compwron (approved):
Looks good! I agree that we need to get #831 done but I don't want to hold this up for fear of merge conflicts. @rafaeelaudibert if you have time, do you want to take on #831 also? :)

rafaeelaudibert:
Hey Guys! Sorry for the delay! I could have tackled that issue, but as I see, it has already been closed. Will take a look on another open issue to give it a go! Thanks for the merge

compwron:
https://bit.ly/casaretro

PR #905, opened 2020-10-02
Feature/supervisors and admins can add next court date

### LINKED ISSUE #807: supervisors and admins can add next court date per volunteer
Part of epic #880 

**What type of user is this for? [volunteer/supervisor/admin/all or All CASA Admin]**
mostly supervisors, but also admins

On the Edit CASA Case page (example: https://casa-qa.herokuapp.com/casa_cases/2/edit)

**Description**
As a supervisor or admin, I should be able to add the next Court Date associated with a `casa_case`.

**Backend**
This is a _brand new_ feature, so an addition will need to be made to db for `court_date` which can be associated with a `casa_case`. (see details in epic #880)

**Frontend**
Next `court_date` should appear on the Edit CASA Case Page, in the same section as *Case number* and *Transition Aged Youth* (see screenshot below). The information entered here should appear for all users viewing a `casa_case`. 
Screen Shot 2020-09-26 at 4 30 47 PM (2)

PuZZleDucK:
Resolves #807 and is part of Hactoberfest

### What changed, and why?
Added court_date field to casa_cases and display value in detail view. Allow supervisors and admins to edit.

- Volunteer permissions: view next court date
- Supervisor permissions: view and edit next court date
- Admin permissions: view and edit next court date

Added expectations to admin and user tests to see the date label and if admin see the default fields.

### Screenshots please :)
Display court date:
Screenshot from 2020-10-03 05-56-07

Supervisor editing case:
Screenshot from 2020-10-03 05-56-21

Volunteer editing case:
Screenshot from 2020-10-03 06-10-44

### Feelings gif (optional)
feelings


config/initializers/time_formats.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+Time::DATE_FORMATS[:date_only] = lambda { |time| time.strftime("%A, %B the #{time.day.ordinalize}") }
```
littleforest, line 3: I prefer using i18n to set custom date and time formats. See the [i18n docs](https://guides.rubyonrails.org/i18n.html#adding-date-time-formats) for examples.

app/policies/casa_case_policy.rb
```diff
@@ -40,6 +40,7 @@ def assign_volunteers?
 
   def permitted_attributes
     common_attrs = [
+      :court_date,
```
littleforest, line 43: Since volunteers can't update the court date, let's add a more specific case for `CasaAdmin` and `Supervisor` like so:

```
    case @user
    when CasaAdmin
      common_attrs.concat(%i[case_number birth_month_year_youth])
    when CasaAdmin, Supervisor
      common_attrs.concat(%i[court_date])
    else
      common_attrs
    end
```

spec/system/volunteer_edits_case_spec.rb
```diff
@@ -26,5 +26,10 @@
     check "Court report submitted"
     click_on "Update CASA Case"
     has_checked_field? :court_report_submitted
+
+    expect(page).to have_text("Court Date")
+    expect(page).not_to have_text("Day")
+    expect(page).not_to have_text("Month")
+    expect(page).not_to have_text("Year")
```
littleforest, line 33: Let's also update the `spec/system/supervisor_edits_case_spec.rb` spec.

littleforest (changes requested):
Hi @PuZZleDucK welcome and thanks for your contribution! In the future, please check and make sure an issue you are working on is not already assigned before you take it on. If it is already assigned, please check with the assignee first before working on it.

app/views/casa_cases/_form.html.erb
```diff
@@ -12,6 +12,16 @@
         <% end %>
       </div>
 
+      <div class="field form-group">
+        <% if current_user.casa_admin? || current_user.supervisor?  %>
+          <%= form.label :court_date, "Next Court Date" %>
+          <br>
+          <span class="datetime-year-month"> <%= form.date_select :court_date, {order: [:day, :month, :year], start_year: Date.current.year, end_year: 1989, prompt: {day: 'Day', month: 'Month', year: 'Year'} }, class: "select2 date-input"%></span>
```
littleforest, line 19: This is giving us a year range from 1989 - 2020, but we actually need years to go at least a couple into the future.

littleforest: We can probably make the start year 2020, unless they are going to backdate any records...? @pollygee ?

Zrrrpy: @littleforest  Court dates can be scheduled pretty far in advance, so I'd definitely add three years into the future. 

It doesn't sound like PG CASA has any plans (right now, anyway) to backlog, but I think there will be enough of a use case potentially beyond PG CASA that we should make the lower year into the early 2000s.

compwron (changes requested):
Hi @PuZZleDucK welcome!
Please make the changes suggested by @littleforest and then one of us will re-review and approve. :)
+1 to what @littleforest said, in the future please check the "Assignees" of an issue before working on it. 
Screen Shot 2020-10-02 at 2 49 34 PM

acrosman:
As a _Guardian ad Litem_ in another state, but new to this project (just found it today but am quite interested in it even if my Ruby is a bit rusty), this looks a little overly simple for what we would need. Tracking with court date should also include hearing type, and expected judge. Ideally that list would be customizable for different states to be able to do swap in lists of hearing types that may vary but state. Volunteers need to know what the hearing type is and may want to know which judge to expect when appearing.

compwron:
> As a _Guardian ad Litem_ in another state, but new to this project (just found it today but am quite interested in it even if my Ruby is a bit rusty), this looks a little overly simple for what we would need. Tracking with court date should also include hearing type, and expected judge. Ideally that list would be customizable for different states to be able to do swap in lists of hearing types that may vary but state. Volunteers need to know what the hearing type is and may want to know which judge to expect when appearing.

@acrosman Thanks for the info! We are building incrementally based on what our stakeholders at PG CASA tell us they need, but we are looking for other CASAs to partner with also. I will bring your comment into our roadmap. If you have time to join us in slack and/or weekly office hours, we would love to hear more of your perspective!

app/views/casa_cases/_form.html.erb
PuZZleDucK: Thanks for the feedback @littleforest and @Zrrrpy. I've made it from 3 years in the future to 2000... sorting it backwards keeps the current year visible in the drop down.
Screenshot from 2020-10-04 10-55-07

config/initializers/time_formats.rb
PuZZleDucK: Done... I also renamed it to 'day_and_date' as the more I read it the more I thought 'date_only' was inaccurate

app/policies/casa_case_policy.rb
PuZZleDucK: Done. I made this two separate cases as the admin seemed to only get the first set it matched.

spec/system/volunteer_edits_case_spec.rb
PuZZleDucK: Added supervisor test... and it turned out the admin test was being skipped... I've updated that so both the admin test cases (edit and back out of edit) now pass :partying_face:

PuZZleDucK:
Alright I've made the suggested changes and I think we should be good to go for the next round :grin:

app/policies/casa_case_policy.rb
littleforest: Whoops, my bad, thanks for catching that!

littleforest (approved):
@PuZZleDucK this is great! Thanks for making those changes.

compwron (approved):
Very cool! :) 💪💖

PR #935, opened 2020-10-04
Replace 'CASA/Prince George' with User casa_org name

### LINKED ISSUE #931: replace org specific language from system emails with `casa_org`
AS a volunteer/supervisor/admin,
I want to see my `casa_org` referenced if at all in system emails,
so that I know the email is intended for me.

**How does it work now?**
There are likely a number of system emails that reference "PG / CASA", regardless of what `casa_org` a user is associated with.

**How should it work instead?**
System emails should only reference the `casa_org` associated with the recipient user.

**Acceptance Criteria**
When a user receives a system email,
if any `casa_org` is referenced, it should be the one they belong to.

Garbar:
Resolves #931 

### What changed, and why?
Replace 'CASA/Prince George' with User casa_org name in emails from the site
For Devise::Mailer added method for users

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Feelings gif (optional)
How to embed:
`alt text`


app/mailers/volunteer_mailer.rb
```diff
@@ -3,11 +3,13 @@ class VolunteerMailer < ApplicationMailer
 
   def deactivation(user)
     @user = user
+    @casa_organization = CasaOrg.find(@user.casa_org_id)
```
jmkoni, line 6: Do we need this? Why not just use @user.casa_org in the view?

app/models/user.rb
```diff
@@ -128,6 +128,10 @@ def inactive_message
   def serving_transition_aged_youth?
     casa_cases.where(transition_aged_youth: true).any?
   end
+
+  def casa_org_display_name
```
jmkoni, line 132: This can just be `casa_org.display_name`

jmkoni (approved):
Looks good!

jmkoni:
Question for @Zrrrpy or @compwron: would we want to have a default value if this isn’t set?

compwron:
> Question for @Zrrrpy or @compwron: would we want to have a default value if this isn’t set?

This should always be set. For now, a default value seems unneeded.

compwron (approved):
Very nice :) 🙌🏼

PR #939, opened 2020-10-04
Feature/remove state in datatable

### LINKED ISSUE #932: search presets and results should clear after a page is loaded
AS a **supervisor** or **admin**,
WHEN I search within Volunteers,
my search history should clear after I navigate away from that view,
SO that I don't get confused by limited results when I eventually return to the Volunteer view.

**Where does/should this occur**
In the <a href="https://casa-qa.herokuapp.com/volunteers">Volunteer view</a> accessible by admins and supervisors. 

**How to access QA**
<a href="https://casa-qa.herokuapp.com/volunteers">Log into QA</a> using the following credentials:
email: supervisor1@example.com OR casa_admin1@example.com
password: 123456

**How it works now**
Search presets and results in the Volunteer view _do not_ clear after a user navigates away from it.

**How it should work**
Search presets and results in the Volunteer view _should_ clear after a user navigates away from it.

Mozart2234:
Resolves #932 

### What changed, and why?

I've changed the config of datatable for volunteers view, I change the stateSave parameter to false because this parameter saves not only information about search input even information other parts of data table.

- Volunteer permissions: None
- Admin permissions: None

I write a system test called `spec/system/search_history_clean_after_navigate_away_volunteer_table_spec.rb`

### Screenshots please :)

I think a video is more useful
[Video](https://drive.google.com/file/d/19o9v4jsUQDSpXugpXr3qxFa3KIe6zRtt/view?usp=sharing)

### DIFF
--- a/app/javascript/src/dashboard.js
+++ b/app/javascript/src/dashboard.js
@@ -41,7 +41,7 @@ $('document').ready(() => {
   // Enable all data tables on dashboard but only filter on volunteers table
   var volunteersTable = $('table#volunteers').DataTable({
     autoWidth: false,
-    stateSave: true,
+    stateSave: false,
     columnDefs: [
       {
         targets: [1],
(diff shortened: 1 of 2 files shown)


compwron (approved):
Very cool! Great test! And nice video :) 💪💖

PR #970, opened 2020-10-04
Use same production environment file for staging, qa, prod

### LINKED ISSUE #895: unify envs
don't use non-production.rb files for diff envs, to avoid making duplicate changes and catching them. 

**Description**
Make sure no one can send mail from staging environment.

### LINKED ISSUE #955: [bug] Error handling when email creds are not correct - creating new casa admin in qa
This is caused by incorrect email config in qa somehow. The user is still created but the email is not sent and the user sees a UI exception on the UI. Instead we should catch the exception and show a better error on the UI. 

BugsnagAPP  2:59 PM
New unhandled error in qa from casa in casa_admins#create (details)
Unhandled error
ActionView::Template::Error: Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true
Location
app/views/devise/mailer/invitation_instructions.html.erb:15 - _app_views_devise_mailer_invitation_instructions_html_erb___2990494963078439499_35580

littleforest:
Resolves #895 
Resolves #955 

### What changed, and why?
* Added APP_ENVIRONMENT variable to all Heroku environments as "staging", "qa", and "production".
* Use DOMAIN environment variable instead of DEFAULT_URL_HOST since they are identical and I don't see that they would ever be different.
* Removed separate staging configuration file and only allow mail to get delivered when APP_ENVIRONMENT is "production".
* Changed `RACK_ENV` and `RAILS_ENV` on casa-qa to "production"
* Changed `RACK_ENV` and `RAILS_ENV` on casa-r4g-staging to "production"
* Minor README fixes

### TODO
- [ ]  Remove `DEFAULT_URL_HOST` variable from Heroku after deployment to each environment.


jmkoni (approved):
Yay getting rid of hash rockets too

compwron:
@littleforest I'm worried that changing qa env to production will make the errors in #casa-bots unreadable since they will all appear to come from production. Do you have any ideas on how to make the errors not look weird in this way?

@jmkoni It's ok to merge PRs after you approve them

config/environments/production.rb
```diff
@@ -1,17 +1,18 @@
...
+  config.action_mailer.perform_deliveries = ENV["APP_ENVIRONMENT"] == "production"
   config.action_mailer.delivery_method = :smtp
   config.action_mailer.smtp_settings = { # WARNING do not let standardrb linter change this block, it breaks
-    :address => 'smtp-relay.sendinblue.com',
-    :port => 587,
-    :user_name => ENV["SENDINBLUE_EMAIL"],
-    :password => ENV["SENDINBLUE_PASSWORD"],
-    :authentication => 'login',
-    :enable_starttls_auto => true
+    address: 'smtp-relay.sendinblue.com',
```
compwron, line 10: In the past, this change has made email stop working. See the line above `# WARNING do not let standardrb linter change this block, it breaks` This seems unreasonable but it has been true in the past and it is the reason why the linter explicitly ignores this file. Hopefully this was an artifact of some other weirdness and is not actually true. Please remove the comment also :)
I will merge this and test it carefully 😓

compwron (approved):
Let's do ittttt

littleforest:
@compwron I'm not familiar with how the Slack integration with Bugsnag works, but can we use an environment variable as the Bugsnag environment?

config/environments/production.rb
littleforest: Oh, that seems very strange. I guess let's see how it goes.

PR #971, opened 2020-10-05
Bugsnag environment

littleforest:
### What changed, and why?
Since all environments are using the production config file now, specifically set the release environment on Bugsnag.

https://docs.bugsnag.com/platforms/ruby/rails/configuration-options/#release_stage


compwron (approved):
You rock! :)

PR #1027, opened 2020-10-08
Revert "Revert "Move user permission logic from erb into pundit policy files ""

compwron:
Reverts rubyforgood/casa#1025

This PR is to bring back the logic from this PR once the build is green :)


compwron:
The branch is green, I think the merge will be fine. I'll keep an eye on it. :)

PR #1032, opened 2020-10-09
Fix failing build

jmkoni:
Thanks to @littleforest for figuring out what was causing the build to fail 👏

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -47,8 +47,4 @@ RSpec.configure do |config|
   config.example_status_persistence_file_path = "#{::Rails.root}/tmp/persistent_examples.txt"
 
   config.filter_rails_from_backtrace!
-
-  # Tmp until we handle the multi-tenancy case where we do not know what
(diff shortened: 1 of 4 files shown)


spec/rails_helper.rb
```diff
@@ -47,8 +47,4 @@
   config.example_status_persistence_file_path = "#{::Rails.root}/tmp/persistent_examples.txt"
 
   config.filter_rails_from_backtrace!
-
-  # Tmp until we handle the multi-tenancy case where we do not know what
-  # casa org we are when we hit the sign-in page.
-  config.before(:each) { create(:casa_org) }
```
jmkoni, line 53: I can add this back if we want, but the specs all still passed when I removed this, so let's start each test with a blank slate :)

jmkoni:
Also want to note @littleforest, I didn't remove any of your commits but I did squash some of them to try to tell the story a bit better :)

compwron (approved):

PR #1033, opened 2020-10-10
Adds tests to DB Seeds

### LINKED ISSUE #995: db seed test
Add a test like this

```
describe "database seeder" do
  it "runs successfully" do
    expect { ActiveRecord::Tasks::DatabaseTasks.load_seed }.not_to raise_error
  end
end
```

kinduff:
Resolves #995 

### What changed, and why?
I modded some constants inside each environment seed file as it was not really necessary to have them set like that. Variables work fine and avoids conflicts when loading them all.

I didn't stub at all, just the environment in order to match the seed file. This makes the run kinda lengthy (see screenshot), but hey, we want to make sure the seeds are working.

None.

Just run:

```shell
rspec spec/seeds/seeds_spec.rb
```

### Screenshots please :)
carbon

### Feelings gif (optional)
joy

### DIFF
--- a/db/seeds/qa.rb
+++ b/db/seeds/qa.rb
@@ -15,29 +15,32 @@ ContactTypeGroup.create(casa_org: pg_casa, name: "CASA").tap do |group|
   ContactType.create(contact_type_group: group, name: "Supervisor")
 end
 
+#
+# Config
+#
 # number of volunteer users and casa cases to generate
-VOLUNTEER_USER_COUNT = 100
-CASA_CASE_COUNT = 150
-SUPERVISOR_COUNT = 5
-
-SEED_PASSWORD = "123456"
+volunteer_user_count = 100
+casa_case_count = 150
+supervisor_count = 5
+seed_password = "123456"
+casa_admin_count = 3
 
 AllCasaAdmin.first_or_create(
   email: "allcasaadmin@example.com",
-  password: SEED_PASSWORD,
-  password_confirmation: SEED_PASSWORD
+  password: seed_password,
+  password_confirmation: seed_password
 )
 
 volunteer_users = []
 
 # generate volunteer users via Faker gem
-VOLUNTEER_USER_COUNT.times do |index|
+volunteer_user_count.times do |index|
   volunteer_email = "volunteer#{index + 1}@example.com"
   volunteer_user = Volunteer.where(email: volunteer_email).first_or_create!(
     casa_org: pg_casa,
     display_name: Faker::Name.unique.name,
-    password: SEED_PASSWORD,
-    password_confirmation: SEED_PASSWORD,
+    password: seed_password,
+    password_confirmation: seed_password,
     active: index % 30 != 0 # creates an inactive user every 30 times this is run
   )
   volunteer_users.push(volunteer_user)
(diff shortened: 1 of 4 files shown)


spec/seeds/seeds_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+require "rake"
+
+RSpec.describe "Seeds" do
+  ["development", "qa", "staging"].each do |environment|
+    describe "for environment: #{environment}" do
+      before do
+        Rails.application.load_tasks
+        allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new(environment))
```
littleforest, line 9: cool, did not know about the `ActiveSupport::StringInquirer` class.

db/seeds/qa.rb
```diff
@@ -15,56 +15,58 @@
...
-VOLUNTEER_USER_COUNT = 100
-CASA_CASE_COUNT = 150
-SUPERVISOR_COUNT = 5
-
-SEED_PASSWORD = "123456"
+volunteer_user_count = 100
+casa_case_count = 150
+supervisor_count = 5
+seed_password = "123456"
+casa_admin_count = 3
```
littleforest, line 26: I think this is a good change to get rid of the constants.

littleforest (approved):
Thanks for working on this! Looks great. 🎉

db/seeds/qa.rb
kinduff: Want me to refactor it a little bit? I see some areas it could be improved.

jmkoni: I’ve been thinking about this and wondering if we can use env variables and just have one seed file.

kinduff: We could. Given the environment we can set some options to add certain data. What I like of the current implementation is that each file has it's environment responsibility, so I kinda have mixed feelings there.

jmkoni: Yeah, but if we need to test and are duplicating logic (I think the files are the same other than the variables), it seems worth it to just have one.

littleforest: That's a good idea to store the counts as environment variables. I think the only reason there are separate files is that we wanted a lot of users created on staging (and qa?), but not that many on development.

jmkoni: Yeah, we should still be able to have the differences without having separate files. Unless there’s another difference I’m missing 😬

kinduff: I can work on this if you want. This PR would remain the same since we're checking to not raise exceptions.

littleforest: I think one file would be better as it seems that the staging and qa seed files have not been getting updated regularly like the development file has, and it seems that the only difference is the quantity of records that are being created.

kinduff: Can this PR be merge then? Let's create an issue for the seed refactor

jmkoni: I opened a new issue. If you want to take it, comment and feel free, otherwise I’ll probably try to do it Monday morning https://github.com/rubyforgood/casa/issues/1035

compwron: Confirmed, we want lots of data on staging but not lots of data on "dev" (local) environments, because we want realism in staging (because we do user acceptance testing and user training in staging) but we want not-slow-to-load in local dev environments, especially for contributors using less fancy hardware. 
As for QA, it didn't exist when we made this decision and I don't have a preference. :)

compwron:
https://bit.ly/casaretro

PR #1044, opened 2020-10-11
Move all docs except README to docs subdirectory

keithrbennett:
It's been hard to easily see all available documentation files in the project root directory in full directory listings of the project root.

This PR moves all of them except for README.md into a `docs` subdirectory, and adjusts the links in README.md accordingly. (There were no links in the other files to README so no modifications were necessary.)

I tested the links in all files locally using `grip` to display the file and `linkchecker -r1 -t100 http://localhost:6419/ ` to check the links. (`pip3 install` installs both of them locally).


keithrbennett:
I see now that there is already a `doc` directory, and we should probably have only one such directory. Personally I think `docs` is the better name but I can move the files in either direction, as you like, just let me know.

keithrbennett:
I've:

* moved the files to the `doc` directory and deleted the `docs` directory
* moved `LICENSE.md` back up to the project root where it probably belongs

compwron (approved):

compwron:
:)

PR #1063, opened 2020-10-13
Turn off auto width for case contacts dataTable

### LINKED ISSUE #1038: Case Contacts page text is not vertically aligned correctly when the window is wide

On the Case Contacts page, when the window is wide, the data is not vertically aligned with the column headings. It shifts to the right. To reproduce, log in to dev seeded application as volunteer1@example.com (you can use `bin/login 1` to do this), select "Case Contacts", and increase window width continuously until you see this effect.

image

carterrwood:

Resolves #1038 

### What changed, and why?

Our data tables were calculating the wrong width. Overwriting those classes width's to be `100%` made them render as expected. 

-- Another way we could fix this is by turning the autoWidth function off, just like how we're doing in `dashboard.js` to the `volunteersTable`. However if we do this, the table doesn't scale when shrinking and expanding the window.

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

Added a test to make sure both the `case-contact-table`'s have the same width. 

### Screenshots please :)

image
Scaled
image

### Feelings gif (optional)
alt text

### DIFF
--- a/spec/system/admin_views_case_contact_index_spec.rb
+++ b/spec/system/admin_views_case_contact_index_spec.rb
@@ -1,0 +1,22 @@
+require "rails_helper"
+
+RSpec.describe "admin views case contacts index page", type: :system do
+  let(:organization) { create(:casa_org) }
+  let(:casa_case) { create(:casa_case, casa_org: organization) }
+  let!(:case_contact) { create(:case_contact, duration_minutes: 105, casa_case: casa_case) }
+
+  it "successfully renders the table header and table body as the same width" do
+    admin = create(:casa_admin, casa_org: organization)
+    sign_in admin
+
+    visit case_contacts_path
+
+    table_header = all('.case-contacts-table').first
+    table_body = all('.case-contacts-table').last
+    expect(table_header['style']).to have_text table_body['style']
+
+    # Resize page and check again
+    page.driver.browser.manage.window.resize_to(2000,2000)
+    expect(table_header['style']).to have_text table_body['style']
+  end
+end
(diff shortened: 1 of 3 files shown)


compwron (approved):
Yay! Sometimes the smallest piece of code makes a big difference :) 
I see that this PR is in "draft" state so I will not merge it yet. When you are ready for final review set it to "ready for review" instead of draft. :)

carterrwood:
> Yay! Sometimes the smallest piece of code makes a big difference :)
> I see that this PR is in "draft" state so I will not merge it yet. When you are ready for final review set it to "ready for review" instead of draft. :)

Yes indeed! You saw my previous changes but they didn't scale the table when the page changed sizes, which strategy do you think is better?

compwron:
Scaling the table on size change is preferrable

carterrwood:
Should be good to go then! 😄

spec/system/admin_views_case_contact_index_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    admin = create(:casa_admin, casa_org: organization)
+    sign_in admin
+
+    visit case_contacts_path
+
+    table_header = all('.case-contacts-table').first
+    table_body = all('.case-contacts-table').last
+    expect(table_header['style']).to have_text table_body['style']
+
+    # Resize page and check again
```
compwron, line 18: very very nice :)

compwron (approved):

compwron:
https://bit.ly/casaretro

PR #1066, opened 2020-10-13
Fix time zone error (#1065).

### LINKED ISSUE #1065: Time zone issue causes test failures when server's time zone is not UTC

When running the tests on my dev host (whose time zone is Eastern US) I get this:

```
    Failure/Error:
       expect(
         response.headers["Content-Disposition"]
       ).to include 'attachment; filename="case-contacts-report-1577836800.csv'
     
       expected "attachment; filename=\"case-contacts-report-1577854800.csv\"; filename*=UTF-8''case-contacts-report-1577854800.csv" to include "attachment; filename=\"case-contacts-report-1577836800.csv"
     Shared Example Group: "can view reports" called from ./spec/requests/case_contact_reports_spec.rb:55
     # ./spec/requests/case_contact_reports_spec.rb:47:in `block (5 levels) in <top (required)>'
```

This test passes CI because the CI's test server is set to UTC as its local time zone.

keithrbennett:
Resolves #1065

### What changed, and why?
A call to `Time.local` was changed to `Time.zone.local` to ensure that the time zone being used in the tests was the time zone for which the Rails server was configured (and not necessarily the server host's native time zone).

### DIFF
--- a/spec/requests/case_contact_reports_spec.rb
+++ b/spec/requests/case_contact_reports_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe "/case_contact_reports", type: :request do
   let!(:case_contact) { create(:case_contact) }
 
   before do
-    travel_to Time.local(2020,1,1)
+    travel_to Time.zone.local(2020,1,1)
     sign_in user
   end
   after { travel_back }


littleforest (approved):
Thanks for fixing this @keithrbennett!

PR #1078, opened 2020-10-14
WIP: Add specified_style_attributes method, use it in a test to restrict test to 1 attribute.

keithrbennett:
The method added assists in testing CSS specified attribute values. I say _specified_ attributes because it tests the values _specified_ for the element, which may be modified at runtime. It parses the string returned by `capybara_element['style']` and returns a hash of attribute key/value pairs.

I don't know, is this functionality provided somewhere else? It is _not_ the same as interrogating the actual values at runtime.

In addition, I've modified a unit test to use this method so that it is more immune to future unrelated attribute changes.

I've put the method in `rails_helper.rb` but realize there is probably a better place for it. What would that place be?

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -27,6 +27,7 @@ RSpec.configure do |config|
   config.include SessionHelper, type: :view
   config.include SessionHelper, type: :request
   config.include Organizational, type: :view
+  config.include CssHelpers
   config.after do
     Warden.test_reset!
   end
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -48,3 +48,16 @@
...
+def specified_style_attributes(capybara_element)
+  style_string = capybara_element['style']
+  attribute_strings = style_string.split(';')
+  attribute_strings.each_with_object({}) do |string, style_hash|
+    first_colon_position = string.index(':')
+    key = string[0...first_colon_position].strip
+    value = string[(first_colon_position+1)..-1].strip
+    style_hash[key] = value
+  end
+end
```
littleforest, line 63: @keithrbennett I would but this in a helper file in order to keep this file clean. You can create a new file in `spec/support`, and then just include it in the `rails_helper`.  For example:

```
# spec/support/css_helpers.rb
module CssHelpers
  def specified_style_attributes(capybara_element)
     ...
  end
end
```

Then add:

```
config.include CssHelpers
```

to the `rails_helper.rb` file.

keithrbennett: Perfect. This is exactly the advice I was seeking. I've made the change.

compwron (approved):
lgtm! Very cool ✨

PR #1099, opened 2020-10-15
Correct linter errors on that were present in master [#1098]

### LINKED ISSUE #1098: Fix linter errors
**Description**
Fix the JS, erb and standardrb errors present on master

jimnanney:
Resolves #1098  

### What changed, and why?

Updated standardrb to ignore the one off after party task. Corrected the linter errors in other files automagically.

No permissions changes

existing test suite

### DIFF
--- a/.standard.yml
+++ b/.standard.yml
@@ -1,3 +1,4 @@
 ignore:
   - 'config/environments/staging.rb'
   - 'config/environments/production.rb'
+  - 'lib/tasks/deployment/20201005203405_populate_case_contact_contact_type.rake'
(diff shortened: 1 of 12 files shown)


.standard.yml
```diff
@@ -1,3 +1,4 @@
 ignore:
   - 'config/environments/staging.rb'
-  - 'config/environments/production.rb'
\ No newline at end of file
+  - 'config/environments/production.rb'
+  - 'lib/tasks/deployment/20201005203405_populate_case_contact_contact_type.rake'
```
compwron, line 4: we should fix this

compwron (approved):
Very cool! ✨

.standard.yml
jimnanney: I thought about doing it, but it is a one off after_party task, that probably only got run once.

compwron: @jimnanney kind of- it got run once on qa, once (successfully on staging, and will on friday be run (hopefully only once) on production. Plus it will be run on every new developer's laptop, once each- until we remove the need for it somehow, probably via smarter db seeding. I'm not too worried about it but it does need to keep working. There's an issue open to write a test for it also

PR #1107, opened 2020-10-16
Fixes the login styling where it looks wonky on the sign_in page

seanmarcia:
Tiny PR to fix stylings.

Before:

before

After:

after


compwron (approved):

compwron:
:)

PR #1111, opened 2020-10-16
Use datepicker for selecting dates

### LINKED ISSUE #820: Occurred at calendar not displaying in Safari on desktop and mobile web
Part of epic #3

**What type of user is this for? [admin/volunteer/supervisor/all or All CASA Admin]**
primarily volunteer, but also admin and supervisor

On the create/edit case contact page accessible via all user dashboards.

**Desired behavior**
A small calendar icon should display on the far right of the "Occurred at" field. Clicking this calendar should trigger a calendar pop out that users can use to select a date to associate with their `case_contact`. 

**Current behavior**
No calendar icon displays in the "Occurred at" field **when accessed on Safari** (mobile and desktop web). 

The fix to this will probably to use some library which makes Safari respect https://caniuse.com/input-datetime
Screen Shot 2020-10-11 at 10 30 11 AM
I checked and it it does work on Firefox and Chrome.

NickSchimek:
Resolves #820 

### What changed, and why?

The date/calendar picker was not displaying on all browsers because the HTML5 date type is not well supported. `bootstrap-datepicker` was already installed and being used in the reports view for selecting dates. Added the necessary attributes to the `occurred_at` field to enable it in the contact form. However, the calendar was not styled correctly and was difficult to read since the spacing between columns was virtually non-existent (This was also an issue on the reports view). Imported the `bootstra-datepicker` css file which fixed the styling issue.

The keyboard can also be used to select a date. The datepicker also displays on my mobile device, which looks better than the native date picker: Here is screen cap of native date picker on iphone x. For whatever reason it was not displaying correctly.

image0

I would prefer to have the native date picker on mobile, but I'm not that well versed in the frontend and I'm not willing to investigate why this is happening. Also it does not look like we are supporting mobile views at this point in time? The datepicker works well on the mobile device and can be entered in manually via the keyboard. Although, I did find a bootstrap css file that was not being used, and when imported it made everything have a smaller/nicer font on the page. This might fix the issue with the native date picker for the mobile phone. I decided to keep that css file out, because the page had been optimized for the larger font display.

### How will this affect user permissions? N/a
- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

Visually

### Screenshots please :)

Screen Shot 2020-10-16 at 6 04 14 PM

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/javascript/src/stylesheets/application.scss
+++ b/app/javascript/src/stylesheets/application.scss
@@ -1,6 +1,7 @@
 @import "base/variables";
 
 @import "bootstrap";
+@import "node_modules/bootstrap-datepicker/dist/css/bootstrap-datepicker";
 @import "~@fortawesome/fontawesome-free/scss/fontawesome";
 @import "~@fortawesome/fontawesome-free/scss/solid";
 
--- a/app/views/case_contacts/_form.html.erb
+++ b/app/views/case_contacts/_form.html.erb
@@ -108,10 +108,9 @@
   </div>
 
   <div class="field occurred-at form-group">
-
-<h2><%= form.label :occurred_at %></h2>
-    <% occurred_at = @case_contact.occurred_at.present? ? @case_contact.occurred_at : Time.zone.now %>
-    <%= form.date_field :occurred_at, value: occurred_at.strftime('%Y-%m-%d'), class: "form-control" %>
+    <h2><%= form.label :occurred_at %></h2>
+    <% occurred_at = @case_contact.occurred_at || Time.zone.now %>
+    <%= form.text_field :occurred_at, value: occurred_at.to_date,  data:{ provide: "datepicker", date_format: "yyyy/mm/dd" }, class: "form-control" %>
   </div>
 
   <div class="field miles-driven form-group">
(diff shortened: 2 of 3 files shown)


compwron (approved):

compwron:
mobile- we want to be more mobile friendly, but currently that has not been focused on
very nice! :)

NickSchimek:
> we want to be more mobile friendly

OK. If that is the case then I think we should use `type=date` setting for mobile devices. I viewed the page on my mobile device using `type=date` and had no issues with Safari, chrome, duck-duck-go mobile browser (FireFox probably works). The native date picker worked like a charm. Where I saw the issue yesterday was with the Brave browser, The brave browser is built off of the chrome browser so I'm surprised it had issues. There was no effect on the native date picker when using the aforementioned bootstrap css file; however the mobile page was easier to view as the font was displayed in a way that could be easily read without having to scroll so much.

I found an article that said `type=date` is fully supported on mobile web browsers, apparently that is not always the case.

compwron:
Sounds good.

compwron:
https://bit.ly/casaretro

PR #1114, opened 2020-10-17
#1112 - Add dependent/destroy relationships to models as appropriate so that seeding is idempotent

### LINKED ISSUE #1112: db seeds file needs to be idempotent, and dependent/destroy relationships implemented
Running the seeds file with no data in the database is fine, but running it a second time gives foreign key violation errors. Need to add `dependent: :destroy` to some of the associations.

keithrbennett:
...which will make the development seeding script idempotent.

Resolves #1112

### What changed, and why?
Some `dependent: :destroy` relationships have been added. Currently, the only top level objects that need to be destroyed in order to empty the data base are `CasaOrg`, `AllCasaAdmin`, and `ContactType`.

A test has been added to the script to ensure that all the known tables have been emptied before records are added.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,7 +14,7 @@ class User < ApplicationRecord
 
   has_many :case_assignments, foreign_key: "volunteer_id", dependent: :destroy
   has_many :casa_cases, through: :case_assignments
-  has_many :case_contacts, foreign_key: "creator_id", dependent: :destroy
+  has_many :case_contacts, foreign_key: "creator_id"
 
   has_many :supervisor_volunteers, foreign_key: "supervisor_id"
   has_many :volunteers, -> { includes(:supervisor_volunteer).order(:display_name) },
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,9 +12,9 @@ class User < ApplicationRecord
 
   belongs_to :casa_org
 
-  has_many :case_assignments, foreign_key: "volunteer_id"
+  has_many :case_assignments, foreign_key: "volunteer_id", dependent: :destroy
   has_many :casa_cases, through: :case_assignments
-  has_many :case_contacts, foreign_key: "creator_id"
+  has_many :case_contacts, foreign_key: "creator_id", dependent: :destroy
 
   has_many :supervisor_volunteers, foreign_key: "supervisor_id"
   has_many :volunteers, -> { includes(:supervisor_volunteer).order(:display_name) },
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -12,17 +12,17 @@ class User < ApplicationRecord
 
   belongs_to :casa_org
 
-  has_many :case_assignments, foreign_key: "volunteer_id"
+  has_many :case_assignments, foreign_key: "volunteer_id", dependent: :destroy
   has_many :casa_cases, through: :case_assignments
-  has_many :case_contacts, foreign_key: "creator_id"
+  has_many :case_contacts, foreign_key: "creator_id", dependent: :destroy
```
littleforest, line 17: I don't think we want `dependent: :destroy` on these because we don't want to lose these `case_assignments` and `case_contacts`. So let's do `dependent: :nullify` which will just clear the `volunteer_id` and `creator_id` fields.

littleforest (changes requested):

app/models/user.rb
littleforest: But actually, nullify will make the corresponding association invalid. Let's just remove the `dependent: ` from here completely. If we ever want to delete a user for reals and we hit a foreign key error, then we will need a more sophisticated approach to reassign this. Does it affect the seeds file if you just remove this change?

keithrbennett: Assuming a user can be _deactivated_ rather than _deleted_, what would be the reason to delete a user if that user is part of the history of a case that is _not_ deleted? If you want to save the case contact history, don't you also want to save the part of that history that identifies the party who made the contact? Wouldn't it be important for practical and maybe even legal reasons?  I noticed that there is an active flag on the User instance and was assuming it would be used instead of deletion for users with case history.

If we remove the dependent:destroy from here, then as you said, we would be leaving the DB in an invalid state. A workaround might be to have a pseudo-user object to represent deleted users, but I think we're voluntarily forfeiting informational value if we erase that history.

Any class whose objects are not deleted implicitly can easily be deleted explicitly in the seeds file, so we have complete flexibility with the domain models.

I'm fine either way but wanted to mention this first before making the change.

keithrbennett: @littleforest I've removed the dependent/destroy as you suggested and destroy the CaseContact instances explicitly in the seeder instead. As an aside, I tried dependent/nullify first, and the user deletion failed because there is a not null constraint on the creator field in case contact.

littleforest: Yes, probably the user would be deactivated rather than deleted, but it's not impossible for someone who has access to the production database to delete a user outside of the UI. In that case we want an error thrown, rather than the code deleting these records. So we should remove the `dependent: :destroy` here, and just make sure the seeds file works fine in terms of the order it deletes records, etc.

keithrbennett: I believe this PR is ready to merge. @compwron @littleforest

compwron:
This PR deserves a lot of review and testing! I will dig into it soonest. .

compwron (approved):
Let's DO this!

Issue #1073, opened 2020-10-14
Make "Manage Court Details" Active column match columns above it true vs Yes


compwron:
**What type of user is this for? volunteer/supervisor/admin/all**
admin

**Description**
Make "Manage Court Details" Active column match columns above it true vs Yes

**Screenshots of current behavior, if any**
<img width="441" alt="Screen Shot 2020-10-13 at 10 20 03 PM" src="https://user-images.githubusercontent.com/578159/95946954-785a9f00-0da2-11eb-8460-9817955f77d0.png">

compwron:
This is a good bitesize issue for a new contributor.

JadeDickinson:
I'm working on this one if that's ok?

PR #1121, opened 2020-10-17
Make "Manage Court Details" Active column match columns above it true vs Yes

JadeDickinson:
Resolves https://github.com/rubyforgood/casa/issues/1073

### What changed, and why?
Changed the "Manage Court Details" table to display Yes/No instead of true/false for Active.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

I've added to the edit page system a test that checks there is a row with 'Spec Test Hearing Type', 'Yes', 'Edit' in the hearing types table.

I've have preferred to use the `:with_cols` option to only match on "Active: Yes" rather than coupling the test to the other values in the same row. However this matcher doesn't understand the layout of the table as is.

### Screenshots please :)
Screenshot 2020-10-18 at 00 43 02

### Outstanding questions
Should I also change the column header from "Active" to "Active?" for consistency?

### DIFF
--- a/app/views/casa_orgs/_hearing_types.html.erb
+++ b/app/views/casa_orgs/_hearing_types.html.erb
@@ -5,7 +5,7 @@
   </div>
 </div>
 
-<table class="table table-striped table-bordered">
+<table class="table table-striped table-bordered" id="hearing-types">
   <thead>
     <tr>
       <th>Name</th>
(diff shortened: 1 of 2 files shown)


compwron (approved):

compwron:
https://bit.ly/casaretro

PR #1124, opened 2020-10-18
1089 Combine 3 seeder scripts into one, improve seeding.

### LINKED ISSUE #1089: Combine and parameterize seed data files
Dev seeds should have minimal data- basically one of each type
QA and Staging should have lots and lots (for realistic test data)

`seeds/dev.rb`-> `Seeder.new(1, 1, 3).run()` (example)
`seeds/qa.rb`-> `Seeder.new(100, 100, 3).run()` (example)
`seeds/staging.rb`-> `Seeder.new(100, 100, 3, random: false).run()` (example)

- [ ] combine dev, qa, staging seed files (remove code duplication)
- [ ] parametrize how many of each type are generated
- [ ] introduce TPM as well as CINA case numbers (the rest of the string stays the same)
- [ ] make sure that `rake db:seed:replant` works (not missing any delete statements up top)
- [ ] testing - try for speed. Keep or improve functionality in `seeds_spec.rb`
- [ ] bonus: provide an option to un-random faker https://github.com/faker-ruby/faker#deterministic-random
- [ ] bonus: introduce a seed value (like rspec) !!

keithrbennett:
Resolves #1089

### What changed, and why?
Combine 3 seeder scripts into one, improve seeding.

Tests are passing. The seeds_spec test was somewhat improved by testing to see that all tables were not empty.


keithrbennett:
Some notes on the commit just pushed:

1. The first two after party tasks are not necessary for seeding, but the next two are necessary. I pulled the functionality out of the latter two and moved it into modules into a new lib/tasks/data_post_processors directory. These are then called by the respective after party tasks. I call the modules directly for seeding, bypassing after party altogether.

(Do we still need the first two tasks for anything? I think the new seeding will eliminate the need for them.)

2. `ContactTypeGroup` contained:

```
  validates :name, presence: true, uniqueness: true
```

This became a problem when there was more than 1 org in the DB, so I changed it to:

```
  validates :name, presence: true
  validates_uniqueness_of :name, scope: :casa_org_id
```

...which fixed the problem. This is consistent with the intent, right?

db/seeds.rb
```diff
@@ -1 +1,100 @@
...
+        CaseContact,
+    ]
+  end
+
+  def destroy_all
+    # Order is important here; CaseContact must be destroyed before the User that created it.
+    # The User is destroyed as a result of destroying the CasaOrg.
+    [SupervisorVolunteer, CaseContact, CasaOrg, AllCasaAdmin, ContactTypeGroup, ContactType].each { |klass| klass.destroy_all }
+    non_empty_classes = active_record_classes.select { |klass| klass.count > 0 }
+    unless non_empty_classes.empty?
```
littleforest, line 66: I think an`if` statements might be a little easier to read here:
```
if non_empty_classes.any?
```

db/seeds/casa_org_populator_presets.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def for_environment
+    {
+        'development' => CasaOrgPopulatorPresets.medium_dataset_options,
+        'qa'          => CasaOrgPopulatorPresets.large_dataset_options,
+        'staging'     => CasaOrgPopulatorPresets.large_dataset_options,
+        'test'        => CasaOrgPopulatorPresets.small_dataset_options
+    }[Rails.env]
+  end
+end
```
littleforest, line 51: Picky: 2 spaces for indent

db/seeds/db_populator.rb
```diff
@@ -0,0 +1,151 @@
...
+    create_users_of_type = ->(klass, count) do
+      (1..count).each do |n|
+        attributes = {
+            casa_org: casa_org,
+            email: email.(klass, n),
+            display_name: Faker::Name.name,
+            password: SEED_PASSWORD,
+            password_confirmation: SEED_PASSWORD
+        }
+        if klass == 'Volunteer' && random(30, random: rng) == 0
```
littleforest, line 72: Can we just use mod here to ensure that at least every 30th record is marked as inactive instead of relying on the randomization?

```
if klass == 'Volunteer' && n % 30 == 0
```

db/seeds/db_populator.rb
```diff
@@ -0,0 +1,151 @@
...
+          attributes.merge(active: false)
+        end
+        klass.create!(attributes)
+      end
+    end
+
+    set_some_volunteers_inactive = -> do
+      volunteers = Volunteer.all.to_a
+      inactive_count = volunteers.size / 30
```
littleforest, line 81: Can we set this 30 value as a constant since it is used in multiple places?

db/seeds/db_populator.rb
```diff
@@ -0,0 +1,151 @@
...
+        casa_case:                  casa_case,
+        creator:                    casa_case.volunteers.sample(random: rng),
+        duration_minutes:           likely_contact_durations.sample(random: rng),
+        occurred_at:                rng.rand(0..6).months.ago,
+        contact_types:              ContactType.all.sample(2, random: rng),
+        medium_type:                CaseContact::CONTACT_MEDIUMS.sample(random: rng),
+        miles_driven:               rng.rand(5..40),
+        want_driving_reimbursement: random_true_false,
+        contact_made:               random_true_false)
+  end
```
littleforest, line 135: picky: 2 spaces for indent

lib/tasks/data_post_processors/case_contact_populator.rb
```diff
@@ -0,0 +1,44 @@
...
+        foster_parent: "Foster Parent",
+        medical_professional: "Medical Professional",
+        other_family: "Other Family",
+        other_support_worker: "Other Support Worker",
+        school: "School",
+        social_worker: "Social Worker",
+        supervisor: "Supervisor",
+        therapist: "Therapist",
+        youth: "Youth"
+    }
```
littleforest, line 17: picky: 2 spaces for indent

db/seeds.rb
```diff
@@ -1 +1,100 @@
...
+        Supervisor,
+        CasaAdmin,
+        AllCasaAdmin,
+        SupervisorVolunteer,
+        CaseAssignment,
+        ContactType,
+        ContactTypeGroup,
+        CaseContact,
+    ]
+  end
```
littleforest, line 59: picky: 2 spaces for indent

db/seeds/db_populator.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  def create_cases(casa_org, options)
+    volunteers = Volunteer.where(active: true).to_a
+    options.case_count.times do
+      new_casa_case = CasaCase.create!(
+          casa_org_id:           casa_org.id,
+          case_number:           generate_case_number,
+          transition_aged_youth: random_true_false)
```
littleforest, line 143: picky: 2 spaces for indent

spec/models/contact_type_group_spec.rb
```diff
@@ -3,8 +3,9 @@
 
 RSpec.describe ContactTypeGroup, type: :model do
   it "should not have duplicate names" do
-    name = create(:contact_type_group, {name: "Test1"})
-
-    expect { create(:contact_type_group, {name: "Test1"})}.to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Name has already been taken")
+    org_id = create(:casa_org).id
+    create_contact_type_group = -> { create(:contact_type_group, {name: "Test1", casa_org_id: org_id }) }
+    create_contact_type_group.()
```
compwron, line 8: fancy :)

compwron (approved):

spec/seeds/seeds_spec.rb
```diff
@@ -1,6 +1,24 @@
...
+      CasaOrg,
+      CaseAssignment,
+      CaseContact,
+      ContactType,
+      ContactTypeGroup,
+      Supervisor,
+      SupervisorVolunteer,
+      User,
+      Volunteer,
+  ]
```
littleforest, line 18: picky: 2 spaces for indent

compwron:
> Some notes on the commit just pushed:
> 
> 1. The first two after party tasks are not necessary for seeding, but the next two are necessary. I pulled the functionality out of the latter two and moved it into modules into a new lib/tasks/data_post_processors directory. These are then called by the respective after party tasks. For seeding, I call the modules directly, bypassing after party altogether.
> 
> (Do we still need the first two tasks for anything? I think the new seeding will eliminate the need for them.)
> 
> 1. `ContactTypeGroup` contained:
> 
> ```
>   validates :name, presence: true, uniqueness: true
> ```
> 
> This became a problem when there was more than 1 org in the DB, so I changed it to:
> 
> ```
>   validates :name, presence: true
>   validates_uniqueness_of :name, scope: :casa_org_id
> ```
> 
> ...which fixed the problem. This is consistent with the intent, right?

Yep, name uniqueness per org is correct. Thanks for the fix!

I'm not sure about the first two rake tasks. I'm wary of removing rake tasks that have already been run in prod- @littleforest do you have thoughts on this?

compwron:
🎉🎉🎉

littleforest (changes requested):
@keithrbennett thanks for all your hard work on this!! 

Did we lose the overall `Rake::Task["after_party:run"].invoke` in the seeds file? I like to have that there so when future population tasks are added, then they are automatically loaded with the seeds file and the seeds file doesn't also have to be updated.

littleforest:
@compwron if the tasks have already been run in prod and we don't need them in development for general seeding purposes then it is absolutely fine to delete them.

littleforest:
Oh, but since the rake tasks have been massaged into new tasks essentially, then it is okay as long as they are idempotent. Let me take another look.

spec/models/contact_type_group_spec.rb
keithrbennett: Thank you. ;) The reasons I like using the lambda are 1) it communicates more clearly that the 2 operations are identical, and 2) it separates low level behavior from higher level behavior, and names that behavior. I'm a big fan of this.

app/models/contact_type_group.rb
```diff
@@ -2,7 +2,8 @@ class ContactTypeGroup < ApplicationRecord
   belongs_to :casa_org
   has_many :contact_types
 
-  validates :name, presence: true, uniqueness: true
+  validates :name, presence: true
+  validates_uniqueness_of :name, scope: :casa_org_id
```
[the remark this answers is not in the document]
keithrbennett: Thanks, I didn't know you could do that, that's probably better.

It looks like this PR has already been merged but I will address your other comments below.

db/seeds.rb
keithrbennett: Yeah, about those 2 extra spaces... ;)

For many years I have been indenting an extra tab stop to indicate that the line(s) are not a new nesting level, but rather a continuation of the previous line. RubyMine has separate settings for the two; I'm pasting an image of the configuration screen to illustrate (there is an "Indent:" field and a "Continuation Indent:" field). This extra indentation communicates information to the reader that is missing when no such distinction is made. The mandate to erase this distinction, in my opinion, is due to lack of awareness about its benefit, and habit.

That all said, if you all feel strongly about enforcing this code standard, I will conform to it.

image

db/seeds.rb
keithrbennett: I agree with you, that would be easier to understand.

db/seeds/casa_org_populator_presets.rb
compwron: The linter will get it when we run it next :)

db/seeds/db_populator.rb
keithrbennett: Yes, as long as you're ok with the lesser amount of variation. For example, if we test against mod zero then specifying a single record for the seeding would result in that sole record always being inactive, and that might be confusing; and there are many more examples. Now that we have the option to specify a random seed we can have or not have predictability as we like.

db/seeds/db_populator.rb
keithrbennett: Yes, excellent idea.

keithrbennett (commented):
If we're pretty sure we'll never need them, and we don't want to delete them, could we maybe move them to another directory (e.g. `attic` or `retired`) so they are inactive and their status is clearer to the reader?

db/seeds/casa_org_populator_presets.rb
littleforest: Oh right! I always forget.  Okay, sorry about all my picky comments.

db/seeds.rb
littleforest: I will let the linter enforce whatever it is set for :)

db/seeds/db_populator.rb
littleforest: Okay, either is fine! Welcome to leave it as is.

keithrbennett:
I'd like to explain why I removed the call to afterparty, and moved a couple of its tasks to separate files.


### Afterparty

Afterparty migrations, like DB migrations, are _all_ run; there is no ability to pick and choose which ones run. For reasons mentioned below, calling afterparty directly was not an option.


### Metadata vs. Data

This is just a gut feeling, but I'm thinking that afterparty should be used for _metadata_ (e.g. contact types), but not for _data_ (e.g. an organization name). Using it for _data_ would not likely be correct across organizations; and including data in seeding instead would make the content of the generated data sets more predictable and manageable. The first two tasks (see below) seem to me to be intended to address temporary omissions in the seeding process that have now been addressed, and those tasks are no longer necessary; but if they _are_ still needed, I would suggest making them available as one-off scripts or rake tasks, rather than as afterparty tasks.

(I'm not sure that the term 'metadata' is the correct term for things like contact types, but for the purposes of this discussion, let's assume that it does; and feel free to correct me if there's a better term.)


### Randomization

The seeding code uses a single random seed in multiple places to ensure the predicatabilty of the generated data. If there is any randomization not under its control (i.e. in afterparty tasks) that affects the data base, that randomization will destroy that predictability.

The test of this would be to ensure that data dumps from multiple app/DB instances created with the same random seed should be identical, except for timestamps.

Although afterparty tasks could support an optional random seed to remedy this, this would probably need to be built into the individual task files. It's likely that over time, the need to do this will be forgotten.


###   Deployment task #1: populate_prince_george_casa_org_details

This task is seeding data.

Given that we are now aiming towards a multitenant app, I thought it should be possible to seed an organization with a generic name, which may not necessarily be PG CASA. I first had the seeding create organizations of random names. Then when I ran afterparty, the PG CASA org was created. This meant that specifying the number of organizations for seeding to create was not reliable. I eventually hard coded the PG CASA name in the first created org, but I don't think that's a great idea, and I'm hoping we can change it in the future. I think having seeded data _not_ appearing to be real data is a good thing. I'm wondering if we could use a randomized city name in the org name such as "#{Faker::Address.city} CASA".

In case part of this was intended to guarantee a consistent org name, currently, the default random behavior is to use a random seed of zero, so we should always see the same organization name for each seeded organization unless we have overridden the default.


### Deployment task #2: populate_missing_display_names

This task is seeding data.

### Deployment task #3: populate_contact_type_groups_and_contact_types

This task populates metadata that needs to exist _prior to_ seeding.

### Deployment task #4: populate_case_contact_contact_type

This task is seeding data.

### Production Use

The fact that tasks #2 and #4 do seeding indicates to me that we are not intending to use afterparty in production. If this is the case, then how are the contact types and type groups being populated in production? 

### Afterparty Timing Conflicts with seeds.rb

Having afterparty tasks do seeding when seeds.rb is run is competitive and not complementary. If any seeding of contact types is to be done in seeds.rb, then step #3 must run _before_ that. And yet step #4 can only work _after_ the data is already there. And having seeding done in 2 (actually 1 + several afterparty tasks) places is confusing.

The current behavior of seeds.rb is now questionable because it attempts to populate contact types but they have not yet been made available.

I think we have a lot to think about.

----

I'd like to start another PR to fix the above error and to implement @littleforest's suggestions made in earlier comments.

compwron:
We have at least two different types of data: 
1. per-casa data (casa name, casa logo, contact type groups, contact types, hearing types, judges)
1. within-casa data (volunteers, supervisors, casa_admins, case contacts, supervisor_vounteer associations, case_assignment associations etc)
1. [bonus] all_casa_admin records, which are neither of the above.
I am fine with populating any data in an afterparty task. 

We DO intentionally use afterparty in production. afterparty is intended for production use. 
We never run seeds in production.

afterparty tasks are run:
1. once in qa (until we delete the db... which has happened and will probably happen again, but which I would rather not do in general)
1. once in staging (until we delete the db... which has happened and will probably happen again, but which I would rather not do in general)
1. once in PROD!
1. on *every* new developer's laptop! (maybe more than once if a developer deletes their database sometimes- which happens often) 

Purposes of afterparty tasks:
1. If data needs to be updated in prod in a way not accomplishable via the UI, afterparty tasks *with tests* are hugely preferable to me updating data in prod by hand in rails console. 
1. afterparty tasks, just like rails migrations, provide a record of what we did and why
1. maybe something else that doesn't come to mind right now?

> Deployment task #3: populate_contact_type_groups_and_contact_types This task populates metadata that needs to exist prior to seeding.
Well that's no good, because seed will always run before afterparty. It doesn't matter on qa/staging/prod (until we drop qa/staging dbs) but new contributors will not get the benefit of this task.

littleforest:
Thanks for all your thoughts @keithrbennett and for clarifications @compwron. Keith, I think a better solution than adding the after party tasks back into the seeds file is to use a populator model like you set up for some of these tables (like the `CaseContactPopulator`). If there is data that we want seeded regularly in our dev environments but we also need populated once on production, we can create a new populator. Then we can create an after party task that calls the populator, and also add a line to the seeds file to call the populator, and we won't end up with a bunch of duplicated code.

PR #1135, opened 2020-10-19
Improve copy and error for past-quarter case contact editing

compwron:
none

### What changed, and why?
Change tooltip copy to make it more understandable based on user testing feedback
> I checked under cases and made sure - these are both cases assigned to this Supervisor.  I see that the ? says "disallow edit contact after quarter" -- not sure if that's a note to programmers or if that would be the note to me as a supervisor. But I'm not sure why some get it and others don't?

no

existing spec us updated

### Screenshots please :)

Screen Shot 2020-10-19 at 4 40 02 PM

### DIFF
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -82,15 +82,15 @@ class CaseContact < ApplicationRecord
     !contact_made.nil?
   end
 
-  def allowed_edit?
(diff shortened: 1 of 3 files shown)


littleforest (approved):
🎉 👯‍♀️ 🚢

PR #1140, opened 2020-10-20
881 court date clear

### LINKED ISSUE #881: Court Date/Report info "clears" once date has passed
Part of epic #880 // dependent on #808 & #805

**What type of user is this for? [ volunteer/supervisor/admin/all or All CASA Admin]**
all

On the Edit CASA Case page (example: https://casa-qa.herokuapp.com/casa_cases/2/edit)
Screen Shot 2020-09-26 at 4 30 47 PM (2)

**Description**
After the Next Court Date has passed, that date should "clear" and display under "`past_court_dates` (see #805), and all other associated data should reset, including:
- `court_report_due_date`
- `court_report_submission` status

**Frontend**
After the Next Court date has passed, *no* date should appear next to text: Next Court Date on the `casa_case` details page, and all associated information above should also reset.

big-meel:
Resolves #881 

### What changed, and why?
Created ".clear_court_dates" method in model
Added scope to filter out cases with passed dates
Added tests to "casa_case_spec.rb" 
Added task to scheduler.rake to update dates 

Tested with rspec

### DIFF
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -65,8 +65,13 @@ class CasaCase < ApplicationRecord
     end
   end
 
-  def clear_date_if_passed
-    update(court_date: nil) if court_date < Time.now
+  def clear_court_dates_when_passed
+    
+  update(court_date: nil, 
+    court_report_due_date: nil,
+    court_report_submitted: false
+  ) if court_date < Time.now
+  
   end
 
   private
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -65,6 +65,10 @@ class CasaCase < ApplicationRecord
     end
   end
 
+  def clear_date_if_passed
+    update(court_date: nil) if court_date < Time.now
+  end
+
   private
 
   def validate_date(day, month, year)
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -35,6 +35,10 @@ class CasaCase < ApplicationRecord
       .where("birth_month_year_youth <= ?", 14.years.ago)
   }
 
+  scope :due_date_passed, -> {
+    where("court_date < ?", Time.now)
+  }
+
   def self.available_for_volunteer(volunteer)
     ids = connection.select_values(%{
       SELECT casa_cases.id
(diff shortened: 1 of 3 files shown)


app/models/casa_case.rb
```diff
@@ -35,6 +35,10 @@ class CasaCase < ApplicationRecord
       .where("birth_month_year_youth <= ?", 14.years.ago)
   }
 
+  scope :due_date_passed, -> {
```
compwron, line 38: yay scope :)

app/models/casa_case.rb
```diff
@@ -65,6 +69,15 @@ def update_cleaning_contact_types(args)
     end
   end
 
+  def clear_court_dates
+    
+  update(court_date: nil,
```
compwron, line 74: This code should be more indented, but the linter will do it the next time we run it (see the readme for lint info). :)

app/models/casa_case.rb
```diff
@@ -65,6 +69,15 @@ def update_cleaning_contact_types(args)
     end
   end
 
+  def clear_court_dates
```
compwron, line 72: I think the scope should be used in here, or else all court dates might get wiped... not just the ones in scope?

lib/tasks/scheduler.rake
```diff
@@ -4,3 +4,10 @@ task transition_youth: :environment do
   CasaCase.should_transition.update_all(transition_aged_youth: true)
   puts "done."
 end
+
+desc "Clear court dates and report information when date has passed, run by heroku scheduler"
+task clear_passed_dates: :environment do
+  puts "Checking case due dates..."
+  CasaCase.due_date_passed.each { |key| key.clear_court_dates }
```
compwron, line 11: :)

compwron (approved):
This works! Ina future PR, I want to make this safer by including a conditional in `.clear_court dates` so that someone doesn't call it by accident on a casa case which should not have its court date cleared! :)

app/models/casa_case.rb
big-meel: It did look like something was a bit off. I had indented it earlier but at some point when making changes I may have missed that. *collapses in disgrace* Thanks for the feedback :)

big-meel:
Thanks so much for this feedback :). going to look for another issue to tackle

Issue #1117, opened 2020-10-17
[bug] casa admin downloading reports


compwron:
bug seen on prod, which was just deployed and is using the same code as in staging and qa
https://rubyforgood.slack.com/archives/C010Z8RK8E6/p1602904458457200
BugsnagAPP  8:14 PM
New unhandled error in production from casa in case_contact_reports#index (details)
Unhandled error
ActiveRecord::StatementInvalid: PG::InvalidTextRepresentation: ERROR:  invalid input syntax for type bigint: ""
LINE 1: ...17') AND (contact_types.contact_type_group_id in ('','1','2'...
                                                            ^
Location
app/models/case_contact_report.rb:24 - block in to_csv
New
8:15
Event in production from casa in case_contact_reports#index (details)
Unhandled error
ActiveRecord::StatementInvalid: PG::InvalidTextRepresentation: ERROR:  invalid input syntax for type bigint: ""
LINE 1: ...17') AND (contact_types.contact_type_group_id in ('','1','2'...
                                                            ^
Location
app/models/case_contact_report.rb:24 - block in to_csv

<img width="816" alt="Screen Shot 2020-10-16 at 8 15 12 PM" src="https://user-images.githubusercontent.com/578159/96327202-74ba5880-0fec-11eb-8623-0fcf0d050621.png">

erik-trantt:
Mind if I take this one?

erik-trantt:
Sample params of a report request: 
```
{
"start_date"=>"2020-05-19", 
"end_date"=>"2020-10-20", 
"supervisor_ids"=>["", "2"], 
"creator_ids"=>["", "1"], 
"contact_type_ids"=>["", "1", "2"], 
"contact_type_group_ids"=>["", "1", "2", "3", "4", "5", "6"], 
"want_driving_reimbursement"=>"true", 
"contact_made"=>"false", 
"has_transitioned"=>"", 
"commit"=>"Download Report"
}
```

Contact Type Group filter fails when selections include the prompt option `-Select Contact Type Groups-`. 
The generated SQL statement has the `WHERE` clause with empty string option, which causes the error.

```rb
# app/models/case_contact.rb

scope :contact_type_groups,  ->(contact_type_group_ids =  nil)  {
  joins(:contact_types).where("contact_types.contact_type_group_id in (?)", contact_type_group_ids)
    if contact_type_group_ids.present?
}

## That JOINS results to this SQL statement:
# SELECT case_contacts.* 
[...]

PR #1143, opened 2020-10-20
1117 bugfix casa admin download csv report

erik-trantt:
Resolves the first 2 bugs mentioned in #1117 

### What changed, and why?
Refactored scope `contact_type_groups` to fix unhandled bug for 
- unhandled situation when user selects the prompt option (with value as am empty string) in Contact Type Group filter
  - `where("contact_types.contact_type_group_id in (?)", contact_type_group_ids)`: `contact_type_group_id` is expecting a `bigint` type, but is given empty string `""`
- duplicate records when `case_contact` involves many `contact_types` of the same `contact_type_group`

```rb
# app/models/case_contact.rb  

  scope :contact_type_groups, ->(contact_type_group_ids = nil) {
    # to handle case when passing ids == [''] && ids == nil
    if contact_type_group_ids&.join&.length&.positive?
      joins(contact_types: :contact_type_group)
        .where(contact_type_groups: { id: contact_type_group_ids })
        .group(:id)
    end
  }
```

In test file `app/models/case_contact.rb`, move other test examples out of "occured at range filter" text example. So that, it makes more sense when looking from high level view. See screenshot:

image

N/A

Refactored and added tests for Contact Type Group filter to cover for edge cases when 
- user selects prompt option => contact_type_group_ids = [""] 
- user selects prompt option and other group types => contact_type_group_ids = ["", "1", "2"] 
- user selects nothing in the Contact Type Group filter => contact_type_group_ids = nil

Test cases locates at
  - `app/models/case_contact.rb`

### Screenshots please :)
image

**Report outcome when select 1 group type: CASA & with / without prompt option**
image

**Report outcome when select prompt option / select nothing**
image

### DIFF
--- a/spec/models/case_contact_report_spec.rb
+++ b/spec/models/case_contact_report_spec.rb
@@ -78,177 +78,203 @@ RSpec.describe CaseContactReport, type: :model do
         expect(contacts.length).to eq(1)
         expect(contacts).to eq([contact])
       end
-      describe "case contact behavior" do
-        it "returns only the case contacts with where contact was made" do
-          create(:case_contact, {contact_made: true})
-          create(:case_contact, {contact_made: false})
+    end
+    describe "case contact behavior" do
+      it "returns only the case contacts with where contact was made" do
+        create(:case_contact, {contact_made: true})
+        create(:case_contact, {contact_made: false})
 
-          report = CaseContactReport.new({contact_made: true})
-          contacts = report.case_contacts
-          expect(contacts.length).to eq(1)
-        end
-        it "returns only the case contacts with where contact was NOT made" do
-          create(:case_contact, {contact_made: true})
-          create(:case_contact, {contact_made: false})
+        report = CaseContactReport.new({contact_made: true})
+        contacts = report.case_contacts
+        expect(contacts.length).to eq(1)
+      end
+      it "returns only the case contacts with where contact was NOT made" do
+        create(:case_contact, {contact_made: true})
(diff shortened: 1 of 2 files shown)


spec/models/case_contact_report_spec.rb
```diff
@@ -78,166 +78,204 @@
...
+      end
+    end
 
-          report = CaseContactReport.new({want_driving_reimbursement: true})
-          contacts = report.case_contacts
-          expect(contacts.length).to eq(1)
-        end
+    describe "wanting driving reimbursement functionality" do
+      it "returns only contacts that want reimbursement" do
+        # want_driving_reimbursement
```
compwron, line 143: This comment is not needed

compwron (approved):
This is a great and very explanatory PR! Thanks for the bug fixes and great tests :) ✨💪

spec/models/case_contact_report_spec.rb
erik-trantt: > In test file `app/models/case_contact.rb`, move other test examples out of "occured at range filter" text example. So that, it makes more sense when looking from high level view. See screenshot:

Good find. That bit is unnecessary.
I think it was there before, cos' I just worked on the "contact type group filter functionality" and indented the rest.

erik-trantt: I didnt realize that reindenting code would create this situation. When I used `git diff` on VSCode, it just showed the code I added and ignore reindentation. 😅

PR #1144, opened 2020-10-20
1090 add default casa logo

### LINKED ISSUE #1090: casa_org without custom logo displays neutral CASA logo by default
AS any user (volunteer/supervisor/admin),
IF my `casa_org` does not have a custom logo,
THEN I should see the neutral CASA logo (the one I see on the log in/out screen) in the site header. 

### Acceptance Criteria
The neutral CASA logo (below) displays by default in a `casa_org`'s header until an Admin customizes their `casa_org` instance and uploads a custom logo to the header.
logo copy

RachelWyatt:
Resolves #1090 

### What changed, and why?
For organizations without a set logo, the sidebar will now display the default CASA logo.

No change to permissions.

- Tests added in ```spec/system/admin_views_volunteers_page_spec.rb``` and ```spec/views/layouts/sidebar.html.erb_spec.rb``` to ensure xpath to default-logo is present on page
- Let me know if there are any thoughts on where/how to test this and I can add those in 

### Screenshots please :)
View before change (logged in as other_casa_admin@example.com, but this will also change supervisor and volunteer views):
Screen Shot 2020-10-20 at 9 36 33 AM

View after change:
Screen Shot 2020-10-20 at 9 35 59 AM

### Feelings gif (optional)

happy dance


spec/system/admin_views_volunteers_page_spec.rb
```diff
@@ -26,6 +26,14 @@
       expect(page).to have_text("Miles Driven")
       expect(page).to have_text("Want reimbursement?")
     end
+
+    it "displays default logo" do
+      sign_in admin
+
+      visit volunteers_path
+
+      expect(page).to have_xpath("//img[@src = '/packs-test/media/src/images/default-logo-c9048fc43854499e952e4b62a505bf35.png' and @alt='CASA Logo']")
```
compwron, line 35: I don't think that this hash will stay the same over time? But maybe it will? 
`'/packs-test/media/src/images/default-logo-c9048fc43854499e952e4b62a505bf35.png`

RachelWyatt: I *think* this hash will stay the same since we're on webpacker v 5.2, which uses ```contenthash``` instead of ```chunkhash``` to generate this -- it should be 100% based on the file contents.

https://github.com/webpack/webpack.js.org/issues/2096

That said, I'm happy to update this to just look for 'default-logo' so the test is more resilient :)

compwron: Let's leave it as is and if it breaks then we'll change it :)

compwron (approved):
:)

Issue #929, opened 2020-10-04
admins can customize judge_names per casa_org


Zrrrpy:
Part of epic #880

AS an **admin**, 
I want to add/edit/show/hide `judge_names` specific to my `casa_org`, 
so that only relevant `judge_names` display for users within my organization. 

**Where does/should this occur?**
Under "Edit Organization" 

**How to access in QA**
<a href="https://casa-qa.herokuapp.com/casa_orgs/2/edit">Log in to QA</a> with the following credentials:
email: casa_admin1@example.com
password: 123456

**Acceptance Criteria**
- [ ] On the Edit CASA Organization page, there is a section called "Manage Court Details" (check to make sure this was not already handled by #928)
In this section, an admin can: 
- [ ] create new `judge_names`
- [ ] edit existing `judge_names`
- [ ] hide `judge_names`
- [ ] view hidden `judge_names`
- [ ] show hidden `judge_names`

matisnape:
@Zrrrpy 
1) How would you like the hiding work? Should it happen on edit judge or from the list of judges in the table (button in Actions column)?
2) How the judges are relevant to `hearing_types` and where can I find this view?

jmkoni:
@matisnape hey, make sure to only claim one story at a time. you are currently assigned to one and sorta claiming two more.

matisnape:
@jmkoni sure, I'm just collecting additional info :)

khiga8:
Hi! Can I claim this issue? :-)

khiga8:
Hey! I'm imagining this implementation will be very similar to Hearing types [here](https://github.com/rubyforgood/casa/pull/1000/files). 

I am thinking I'll need to introduce a `JudgeName` model that `belongs_to` `:casa_org` .
Besides a name column, do we need anything else?  

I'm a little confused about the last point: 
`changes admins make here are reflected in hearing_types available to supervisors`

compwron:
@khiga8 that matches what I am thinking. I am also a little confused about `changes admins make here are reflected in hearing_types available to supervisors` 
@Zrrrpy can you elaborate? :)

Zrrrpy:
just removed that line from the acceptance criteria. Apologies for the confusion!

Zrrrpy:
@khiga8 active / inactive status would be great. This will allow admins to show/hide judges :)

Issue #927, opened 2020-10-03
supervisors and admins can associate judge_name and hearing_type with next court_date


compwron:
Part of epic #880 

AS a supervisor or admin, 
I want to be able to associate a `court_date` with a `judge_name` and `hearing_type`,
so that `judge_name` and `hearing_type` display along with a volunteer's next `court_date` associated with a `casa_case` they are assigned to.

**Acceptance criteria**
- [ ] Supervisors and admins can select a `hearing_type` to associate with a `court_date` from a dropdown menu (or similar alternative)
- [ ] Supervisors and admins can select a `judge_name` to associate with a `court_date` from a dropdown menu 
- [ ] `judge_name` and `hearing_type` display along with next `court_date` on its associated `casa_case` for all users.

haydenrou:
I'd like to pick this up please, but I do not have permissions to assign it to myself.

compwron:
@haydenrou All yours :)

compwron:
Resolved by #1000

Zrrrpy:
#1000 lays the foundation work for this issue (gives admins the ability to create and customize `hearing_types` for their organization), but does not complete it. See acceptance criteria above @haydenrou and @compwron.

haydenrou:
As mentioned in #1000 , that was just to address #928 - I will do this in a separate PR.

Zrrrpy:
Thanks @haydenrou! Moved this back to In Progress so that PR can be linked to this issue. :)

haydenrou:
Waiting for #929 before submitting a PR for the second half of this.

PR #1157, opened 2020-10-22
Add selection from casa cases to hearing types

haydenrou:
This PR addresses half of #927 
Due to #929 not being complete, I cannot complete the second half of this PR (adding the association and selection of judge names), so I will submit that separately later.

### What changed, and why?
An association and column has been added between casa cases and hearing types, and a selection for this in the create page has been added. I have shown it on the casa case show and index page too.

### Screenshots please :)
addhearingtypes


spec/models/casa_case_spec.rb
```diff
@@ -75,7 +76,7 @@
 
     context "when volunteer has no case assignments" do
       it "returns all cases in volunteer's organization" do
-        expect(described_class.available_for_volunteer(volunteer)).to eq [casa_case2, casa_case3, casa_case1]
+        expect(described_class.available_for_volunteer(volunteer).ids).to eq [casa_case2.id, casa_case3.id, casa_case1.id]
```
haydenrou, line 79: specerror
Some reason I'm getting this error if I don't change this - so I amended to fix.

app/views/casa_cases/_form.html.erb
```diff
@@ -12,6 +12,16 @@
         <% end %>
       </div>
 
+      <div class="field form-group">
+        <%= form.label :hearing_type_id %>
+        <%= form.collection_select(
+          :hearing_type_id,
+          HearingType.active.for_organization(current_user.casa_org),
```
haydenrou, line 19: stackerror
Getting a `stack level too deep`error here if using `current_organization` for some reason. Tried to debug briefly with Jeanine in Slack with no luck as of yet. Open to suggestions.

compwron: hmm weird. probably a factories problem. We should fix this
I made https://github.com/rubyforgood/casa/issues/1166 to track it, feel free to pick it up or not  :)

spec/models/casa_case_spec.rb
compwron: If this screenshot included the section under "Diff" maybe I could tell what was up (or maybe not) but this seems fine for now :)

compwron (approved):

spec/models/casa_case_spec.rb
haydenrou: Argh, it must've been to do with some of the other tests I hadn't fixed yet. I can't replicate it. Sorry! Would you like me to revert that part in a separate PR or shall we leave it?

compwron: Revert in separate PR would be great :)

Issue #1164, opened 2020-10-22
Revisiting Standardrb Rules


keithrbennett:
I just ran standardrb --fix on master and viewed the differences. I believe that some of the rules it enforces are examples of poorer, not better code. I'd like to see if we can get consensus on any of this, and if so, I volunteer to implement the changes. Here are the issues I found:

### Horizontal Alignment of Values and Blocks

![image](https://user-images.githubusercontent.com/28410/96820695-7de46480-13f4-11eb-9131-b76684a8c038.png)
![image](https://user-images.githubusercontent.com/28410/96820719-89d02680-13f4-11eb-9b0e-deb972c65cda.png)

This is one of the most egregious ones because it violates basic user interface design principles. Horizontally aligning the values allows the reader to spot the values with a minimum of eye movement and effort. Would any of us ever implement text fields in a web form in this way, where labels are horizontally aligned, but the input fields are horizontally scattered? Why aren't we developers worthy of the same consideration as our users?

### And vs. &&

![image](https://user-images.githubusercontent.com/28410/96820782-aff5c680-13f4-11eb-9a5f-8357bd312af1.png)

The prevailing wisdom, as I understand it, is that `&&/||` should be used for logical operations but that `and/or` should be used for flow control (see Avdi Grimm's article at https://avdi.codes/using-and-and-or-in-ruby/). So `and/or` should be allowed. Also, the two pairs differ in precedence rules; I wonder if this could result in runtime errors in some cases as well.

### Parenthesizing Ternary Conditions

![image](https://user-images.githubusercontent.com/28410/96820821-bf750f80-13f4-11eb-84d9-8a6348ad344a.png)

It is often helpful to the reader to wrap a ternary expression's condition in  in parentheses, to explicitly show its beginning and end. This is especially true when the condition is more complex than the one in this example.

### Method Parameter Indentation

![image](https://user-images.githubusercontent.com/28410/96820859-d4ea3980-13f4-11eb-9a32-4d0109f2c934.png)

The example in this image is not so bad, but I've seen cases where horizontal space is limited, and this formatting style pushes parameters way to the right, sometimes out of view.

### Expression Continuation Indentation

![image](https://user-images.githubusercontent.com/28410/96821521-09aac080-13f6-11eb-8fcd-c18e06317229.png)

[...]

PR #1169, opened 2020-10-22
and return not && return

jmkoni:
As part of the conversation in #1164, we noticed that it was autocorrecting `and return` to `&& return` which isn't the same. Added a comment to disable the rule for those three lines.


searls:
For reference, this issue was raised on the cop: https://github.com/rubocop-hq/rubocop/issues/1083

PR #1170, opened 2020-10-22
Test seeding for only development, no longer for qa and staging.

keithrbennett:
Now that the same seed code is used for all three environments, we can test that code once. This cuts the seed test time to a quarter of what it was before.

### DIFF
--- a/spec/seeds/seeds_spec.rb
+++ b/spec/seeds/seeds_spec.rb
@@ -3,18 +3,18 @@ require "rake"
 
 def empty_ar_classes
   ar_classes = [
-      AllCasaAdmin,
-      CasaAdmin,
-      CasaCase,
-      CasaOrg,
-      CaseAssignment,
-      CaseContact,
-      ContactType,
-      ContactTypeGroup,
-      Supervisor,
-      SupervisorVolunteer,
-      User,
-      Volunteer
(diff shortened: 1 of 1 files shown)


littleforest:
Yay for speeding up the test suite! Needs to pass linting, but otherwise good to go.

spec/seeds/seeds_spec.rb
```diff
@@ -3,34 +3,32 @@
...
+      Rails.application.load_tasks
+      allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new('development'))
+    end
 
-      it "successfully populates all necessary tables" do
-        ActiveRecord::Tasks::DatabaseTasks.load_seed
-        expect(empty_ar_classes).to eq([])
-      end
+    it "successfully populates all necessary tables" do
+      ActiveRecord::Tasks::DatabaseTasks.load_seed
```
compwron, line 30: At some point I would like to also add more asserts here- something like 
`expect(CasaOrg.first.supervisors.first.volunteers.first.cases.first.case_contacts.first.contact_type).not_to be_null` (that syntax might need some work) - basically something to make sure that everything in there is hooked together lately

compwron (approved):

PR #1172, opened 2020-10-23
Stub current_organization in view spec

### LINKED ISSUE #1166: make factory not recurse
see https://github.com/rubyforgood/casa/pull/1157/files#r509862385

littleforest:
Resolves #1166 

### What changed, and why?
* Reverted to `current_organization` in casa case form.

* Stubbed `current_organization` in view spec. Did not figure out why the stack level too deep error was occurring, but since view spec doesn't actually hit the controller, okay to stub. There are existing request specs that work that test the controller and form view work just fine.


haydenrou (approved):
Awesome, thanks!

jmkoni (approved):

PR #1211, opened 2020-10-28
error when admin imports inactive casa_case

### LINKED ISSUE #1190: error when admin attempts to import inactive casa_case
dependent on #1182

AS an **admin**,
WHEN I attempt to import an inactive `casa_case`,
I _do not_ want the `casa_case` to be re-imported, 
_instead_, I WANT to receive an informative error message,
SO THAT I understand why the inactive `casa_case` was not imported.

### How it works now
`casa_cases` currently only have one state, which is active, but #1182 is changing that – soon, `casa_cases` will have two states: active or inactive.

### How it should work
When an admin attempts to import an _inactive_ `casa_case` which is already in the system, the inactive `casa_case` should not be imported. Instead, a reason why that row of the import failed should be listed in the "Errors" column available for download after an import attempt.

That error should read something like:
_CASA case already exists, but is inactive. Reactivate the CASA case instead._

steph-hickman9:
An additional check was made to provide a different error
message to an admin importing an existing case that is
inactive. This Provides more useful information than just telling
them that the case already exists.

Tests were added to ensure that the error message renders correctly.

Resolves #1190 

### What changed, and why?
An additional check was made to provide a different error
message to an admin importing an existing case that is
inactive. This provides more useful information than just telling
them that the case already exists.

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

Tests were added to ensure that the error message renders correctly.

### Screenshots please :)
Screen Shot 2020-10-28 at 12 43 16 PM

### Feelings gif (optional)
How to embed:
gif

### DIFF
--- a/spec/requests/imports_request_spec.rb
+++ b/spec/requests/imports_request_spec.rb
@@ -3,6 +3,8 @@ require "rails_helper"
 RSpec.describe "/imports", type: :request do
   let(:volunteer_file) { Rails.root.join("spec", "fixtures", "volunteers.csv") }
   let(:supervisor_file) { Rails.root.join("spec", "fixtures", "supervisors.csv") }
+  let(:case_file) { Rails.root.join("spec", "fixtures", "casa_cases.csv") }
+  let(:existing_case_file) { Rails.root.join("spec", "fixtures", "existing_casa_case.csv") }
   let(:supervisor_volunteers_file) { Rails.root.join("spec", "fixtures", "supervisor_volunteers.csv") }
   let(:casa_admin) { create(:casa_admin) }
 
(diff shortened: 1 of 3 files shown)


spec/requests/imports_request_spec.rb
```diff
@@ -120,6 +122,59 @@
...
+          }
+      }.to change(CasaCase, :count).by(0)
+
+      expect(request.session[:import_error]).to include("Not all rows were imported.")
+      expect(request.session[:exported_rows]).to include("Case CINA-00-0000 already exists, but is inactive. Reactivate the CASA case instead.")
+      expect(response).to redirect_to(imports_url(import_type: "casa_case"))
+    end
   end
 end
 # TODO: dominique test for message success/fail
```
compwron, line 180: @dominiquecuevas I think you missed a spot :)

compwron (approved):
Looks great! And yay tests!!!!

compwron:
The linter is mad. I'm going to just merge it and then fix the lint.

compwron:
https://bit.ly/casaretro

PR #1213, opened 2020-10-28
Refactor/Improve specs readability

cassianoblonski:
### What changed, and why?
* Enforce that subject is the first definition in the test.
* Fix incorrect uses of context to specify methods.
* Change some descriptions
* Remove should syntax in favor of expect
* Use do end for multiline blocks `(example: let(:params) {... multilines ...} to  let(:params) do ... multilines ... end)`

#### Use the `disable_monkey_patching!` configuration option to disable all monkey patching done by RSpec:
* stops exposing DSL globally
* disables should and should_not syntax for rspec-expectations
* disables stub, should_receive, and should_not_receive syntax for rspec-mocks

### How is this tested? 💖💪
Running the full test suite.

### DIFF
--- a/spec/models/case_contact_spec.rb
+++ b/spec/models/case_contact_spec.rb
@@ -2,8 +2,8 @@ require "rails_helper"
 
 RSpec.describe CaseContact, type: :model do
   context "validations" do
-    it { should validate_numericality_of(:miles_driven).is_less_than 10000 }
-    it { should validate_numericality_of(:miles_driven).is_greater_than_or_equal_to 0 }
+    it { is_expected.to validate_numericality_of(:miles_driven).is_less_than 10_000 }
+    it { is_expected.to validate_numericality_of(:miles_driven).is_greater_than_or_equal_to 0 }
   end
 
   it "belongs to a creator" do
@@ -82,7 +82,7 @@ RSpec.describe CaseContact, type: :model do
     expect(case_contact.errors[:base]).to eq(["cannot edit case contacts created before the current quarter"])
   end
 
-  context "#update_cleaning_contact_types" do
+  describe "#update_cleaning_contact_types" do
     it "cleans up contact types before saving" do
       group = create(:contact_type_group)
       type1 = create(:contact_type, contact_type_group: group)
@@ -90,12 +90,12 @@ RSpec.describe CaseContact, type: :model do
 
       case_contact = create(:case_contact, contact_types: [type1])
 
-      expect(case_contact.case_contact_contact_type.count).to eql 1
+      expect(case_contact.case_contact_contact_type.count).to be 1
       expect(case_contact.contact_types).to match_array([type1])
 
       case_contact.update_cleaning_contact_types({case_contact_contact_type_attributes: [{contact_type_id: type2.id}]})
 
-      expect(case_contact.case_contact_contact_type.count).to eql 1
+      expect(case_contact.case_contact_contact_type.count).to be 1
       expect(case_contact.contact_types.reload).to match_array([type2])
     end
   end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -17,6 +17,7 @@ RSpec.describe User, type: :model do
     user = build(:user, display_name: "")
     expect(user.valid?).to be false
   end
+
   it "returns all case_contacts associated with this user and the casa case id supplied" do
     volunteer = create(:volunteer, :with_casa_cases)
 
@@ -51,7 +52,7 @@ RSpec.describe User, type: :model do
   end
 
   describe "supervisors" do
-    context "#volunteers_serving_transistion_aged_youth" do
+    describe "#volunteers_serving_transistion_aged_youth" do
       it "returns the number of transition aged youth on a supervisor" do
         assignment1 = create(:case_assignment, casa_case: create(:casa_case, transition_aged_youth: true))
         assignment2 = create(:case_assignment, casa_case: create(:casa_case, transition_aged_youth: true))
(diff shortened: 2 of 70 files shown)


cassianoblonski:
@jmkoni @compwron 
Can you guys take a look here and review these before some conflicts show up?

spec/models/case_contact_spec.rb
```diff
@@ -2,8 +2,8 @@
 
 RSpec.describe CaseContact, type: :model do
   context "validations" do
-    it { should validate_numericality_of(:miles_driven).is_less_than 10000 }
-    it { should validate_numericality_of(:miles_driven).is_greater_than_or_equal_to 0 }
+    it { is_expected.to validate_numericality_of(:miles_driven).is_less_than 10_000 }
+    it { is_expected.to validate_numericality_of(:miles_driven).is_greater_than_or_equal_to 0 }
```
jmkoni, line 6: Why this change? Not saying it's wrong, just curious.

spec/models/case_contact_spec.rb
```diff
@@ -82,20 +82,20 @@
     expect(case_contact.errors[:base]).to eq(["cannot edit case contacts created before the current quarter"])
   end
 
-  context "#update_cleaning_contact_types" do
+  describe "#update_cleaning_contact_types" do
```
jmkoni, line 85: Same here... did you make this change because there's not a describe within context?

spec/models/user_spec.rb
```diff
@@ -51,7 +52,7 @@
   end
 
   describe "supervisors" do
-    context "#volunteers_serving_transistion_aged_youth" do
+    describe "#volunteers_serving_transistion_aged_youth" do
```
jmkoni, line 55: Yeah, can you clarify a bit more on why you changed some of these? What was your standard for `describe` over `context` and vice versa?

spec/models/volunteer_spec.rb
```diff
@@ -75,11 +75,12 @@
   describe "#made_contact_with_all_cases_in_days?" do
     let(:volunteer) { create(:volunteer) }
     let(:casa_case) { create(:casa_case) }
-    let(:create_case_contact) {
-      ->(occurred_at, contact_made) {
+    let(:create_case_contact) do
+      lambda { |occurred_at, contact_made|
```
jmkoni, line 79: Not sure this change was necessary...

spec/requests/all_casa_admins_spec.rb
```diff
@@ -45,14 +45,14 @@
     before { sign_in admin }
 
     context "with valid parameters" do
-      let(:params) {
+      let(:params) do
```
jmkoni, line 48: Don't think this is necessary either.

spec/requests/all_casa_admins_spec.rb
```diff
@@ -62,14 +62,14 @@
     end
 
     context "with invalid parameters" do
-      let(:params) {
+      let(:params) do
```
jmkoni, line 65: Same here.

spec/requests/case_contact_reports_spec.rb
```diff
@@ -21,12 +22,12 @@
 
     shared_examples "can view reports" do
       context "with start_date and end_date" do
-        let(:case_contact_report_params) {
+        let(:case_contact_report_params) do
```
jmkoni, line 25: Same here, unnecessary change.

spec/requests/volunteers_spec.rb
```diff
@@ -36,16 +36,18 @@
     before do
       sign_in admin
     end
+
     context "with valid params" do
-      let(:params) {
+      let(:params) do
```
jmkoni, line 41: Same

spec/requests/volunteers_spec.rb
```diff
@@ -61,16 +63,18 @@
         }.to change { ActionMailer::Base.deliveries.count }.by(1)
       end
     end
+
     context "with invalid parameters" do
-      let(:params) {
+      let(:params) do
```
jmkoni, line 68: Same

jmkoni (commented):
I left a few comments, but I'm mostly unsure why this change is necessary...

spec/models/case_contact_spec.rb
compwron: this is the new rspec style 
I have used https://www.betterspecs.org/#expect in the past and I believe it is one of the influences for this change.

spec/models/case_contact_spec.rb
compwron: context and describe do the same thing, it's just a matter of which one "sounds" better
In general, describe is for methods, and context is for data setup- but it is not a hard and fast rule https://www.betterspecs.org/#describe

spec/models/volunteer_spec.rb
compwron: I am happy with lambda over `->` as it is more google-able. it's a little odd to me that `standardrb` doesn't enforce it. As long as they are fully equivalent for test setup (i.e. the tests don't fail or falsely-pass) and the linter likes it, I am happy with the more searchable term.

spec/models/case_contact_spec.rb
jmkoni: Yeah, that was my understanding. That's why I'm unsure as to why these were changed in this way.

spec/requests/all_casa_admins_spec.rb
compwron: I usually see `do` used over `{` on multi-line blocks and I thought that standardrb enforced it. Is our linter skipping our specs? And if so, is that intentional? They're not in the ignore file

jmkoni: Not sure about that (re: standard). I'm not sure how much it digs into rspec yet. I don't really care, I'm just not in favor of making a change like this just because. But 🤷‍♀️ this is why I didn't mark them as blocking.

jmkoni: `let` and other rspec-isms are ignored by Standard's semantic blocks https://github.com/testdouble/standard/blob/master/config/base.yml#L745

spec/models/case_contact_spec.rb
compwron: This usage of "describe" for a method is more correct than "context" for a method according to https://www.betterspecs.org/#describe above which I assume is the explanation :)

compwron (approved):
The vast majority of these changes look unequivocally positive! I am happy to review a follow-up PR to change any lingering issues. :)

compwron:
https://bit.ly/casaretro

PR #1283, opened 2020-11-08
Docker scripts

### LINKED ISSUE #1116: Docker Scripts
I'd like to add scripts for executing Docker commands.  This would free people from having to repeatedly type in "docker-compose run --rm web".

Examples of Docker scripts are:
- https://github.com/openfoodfoundation/openfoodnetwork/tree/master/docker
- https://github.com/codebar/planner/tree/master/bin

jhsu802701:
Resolves #1116 

### What changed, and why?
* I've added a new "docker" directory containing scripts for using Docker to execute tasks.  This removes the need to keep typing in "docker-compose run --rm web" and "docker-compose exec web".
* After downloading this repository, just go to the main directory and enter "docker/build".  This automatically builds the app, seeds the database, starts the web server, runs the test suite, AND logs the screen outputs.
* To run the test suite, just enter "docker/test".
* To reseed the database, just enter "docker/seed".
* To restart the web server, just enter "docker/server".
* To run a command in the app, just enter "docker/run (command)".  For example, you can run "cat /etc/os-release" in the Docker container by entering "docker/run cat /etc/os-release".

### Updating the documentation
I know that I need to update the doc/DOCKER.md file.  Should I just add in a section called "The Easy Way" at the beginning and leave everything else alone?  Or should I integrate these new Docker scripts into what's already written?


compwron:
Easy way sounds fine :)

littleforest:
@jhsu802701 thanks for working on this. Could you title the section "Using automated scripts" or something rather than "The easy way"? I think it's important for people who are learning docker to be able to run the setup commands themselves so they can understand how to use docker (with docker-compose). That often gets lost when using scripts that can be hard to follow for new developers, especially those not well-versed in bash.

jhsu802701:
Thanks for your input.  I've added the "Automated Scripts" section with instructions on how to use the scripts in the docker directory.

littleforest (changes requested):
Thanks for adding these @jhsu802701 . Could you also add a line or two under the "Using Automated Scripts" and/or the (new) "Manual setup" that users should use one section or the other? I'm just worried about newer developers being confused about having two sets of instructions.

Maybe something like: "The following scripts can be used to quickly build and start the Docker containers. If you prefer to manually start the containers, see the #manual-setup section" (something like this, feel free to edit)

littleforest (approved):
Thanks for making those changes!

PR #1300, opened 2020-11-10
WIP: render volunteer names in plain text if current user is volunteer

### LINKED ISSUE #1269: Names should not be clickable under "Assigned to" for volunteers
AS a **volunteer**,
I WANT to see my name (and the names of any other volunteers assigned to my `casa_case`), 
but I SHOULD NOT be able to click on them,
BECAUSE I do not have permission to edit that page. 

### How it works now
Volunteers can click on names under "Assigned to" in the "My Cases" view, but when they do, they get an error message: _Sorry, you are not authorized to perform this action._

### How it should work
Names under "Assigned to" in "My Cases" should _not_ be clickable for volunteers (but they _should_ remain clickable for admins and supervisors).

metamoni:
Resolves #1269 

### What changed, and why?
In Casa Case Detail View, volunteer names appeared as a link to all users. Now it appears as plain text if the current user is a volunteer.

- Volunteer permissions: n/a
- Supervisor permissions: n/a
- Admin permissions: n/a

### Screenshots please :)

Screenshot 2020-11-10 at 20 52 35

### DIFF
--- a/app/views/casa_cases/show.html.erb
+++ b/app/views/casa_cases/show.html.erb
@@ -55,7 +55,11 @@
     <div>
       <h6><strong>Assigned Volunteers:</strong></h6>
       <% policy_scope(assigned_volunteers(@casa_case)).each_with_index do |volunteer, index| %>
-        <p><%= link_to "#{volunteer.decorate.name} ", edit_volunteer_path(volunteer) %></p>
+        <% if current_user.volunteer? %>
+          <p><%= volunteer.decorate.name %></p>
+        <% else %>
+          <p><%= link_to "#{volunteer.decorate.name} ", edit_volunteer_path(volunteer) %></p>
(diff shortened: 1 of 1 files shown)


compwron (approved):
I would like to also add a test for this, to keep the functionality from being left out by accident when we change this page again / refactor (as has happened so many times before) 😅
On the other hand we don't want too many tests... maybe we can add a line to a pre-existing test?

jhsu802701 (changes requested):
Please add integration tests to confirm that this feature works as expected.

metamoni:
No worries, I'll add some integration tests 👍

compwron:
I'm going to merge this for now, feel free to open a new PR on the same branch with new commits if that's easiest for you, or a new PR on a different branch, whatever works :)

PR #1288, opened 2020-11-09
Move from travis to github actions

compwron:
### What github issue is this PR for, if any?
Resolves #1245 

### What changed, and why?
Move from travis to github actions

Travis CI .com is going to stop providing services for free soon and .org is currently very slow and also going away in December. 

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

### How will this affect user permissions?
no

### How is this tested? (please write tests!) 💖💪
run the build

### Feelings gif (optional)
alt text

PR #1303, opened 2020-11-10
Updated SimpleCov configuration

jhsu802701:
I made the following changes to the SimpleCov configuration:
* Filtered out the db and spec directories
* Provided grouped coverage results

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,8 +1,19 @@
 require "simplecov"
 require "pry"
 SimpleCov.start do
-  track_files "{app,lib}/**/*.rb"
+  add_filter '/db/'
   add_filter '/spec/'
+  add_group "Models", "app/models"
+  add_group "Controllers", "app/controllers"
+  add_group "Channels", "app/channels"
+  add_group "Decorators", "app/decorators"
+  add_group "Helpers", "app/helpers"
+  add_group "Jobs", "app/jobs"
+  add_group "Importers", "app/lib/importers"
+  add_group "Mailers", "app/mailers"
+  add_group "Policies", "app/policies"
+  add_group "Values", "app/values"
+  add_group "Libraries", "lib"
 end
 
 RSpec.configure do |config|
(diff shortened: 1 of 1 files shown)


compwron:
@jhsu802701 What change does this make to the output of simplecov? Please add before/after screenshots or similar.

While you're in there, if you could add the simplecov publisher to our github actions that would be very useful! (It got removed when we switched from travisci to github actions.) https://github.com/rubyforgood/casa/pull/1288 https://github.com/rubyforgood/casa/issues/1304

spec/spec_helper.rb
```diff
@@ -1,7 +1,19 @@
 require "simplecov"
 require "pry"
 SimpleCov.start do
-  track_files "{app,lib}/**/*.rb"
+  add_filter '/db/'
```
compwron, line 4: We do actually want to assert coverage on `db/seeds` so please don't filter out all of `db`

spec/spec_helper.rb
```diff
@@ -1,7 +1,19 @@
...
+  add_group "Controllers", "app/controllers"
+  add_group "Channels", "app/channels"
+  add_group "Decorators", "app/decorators"
+  add_group "Helpers", "app/helpers"
+  add_group "Jobs", "app/jobs"
+  add_group "Importers", "app/lib/importers"
+  add_group "Mailers", "app/mailers"
+  add_group "Policies", "app/policies"
+  add_group "Values", "app/values"
+  add_group "Libraries", "lib"
```
compwron, line 16: If this grouping changed the output in a nice way, let's also add a group for `after_party` tasks, which should have coverage.

jhsu802701: I replaced the "lib" group with a "/lib/tasks" group.

spec/spec_helper.rb
jhsu802701: I removed the "/db" filter.

compwron:
I ran it locally rather than requesting again @jhsu802701 to post a screenshot, it doesn't break anything. I'll merge it. 

@jhsu802701 if you could add the simplecov publisher to our github actions that would be very useful! (It got removed when we switched from travisci to github actions.) 
PR: #1288 
github issue: #1304

compwron (approved):

PR #1315, opened 2020-11-11
Make print statements conditional so that they will not show up in rspec

jhsu802701:
* The puts commands in the deployment tasks are disabled for the test environment only.
* The puts commands in the db/seeds.rb script are commented out to minimize excessive screen output when running the test suite.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -23,10 +23,10 @@ class SeederMain
   end
 
   def seed
-    puts "Erasing all objects from the data base..."
+    # puts "Erasing all objects from the data base..."
     destroy_all
 
-    puts "Creating the objects in the data base..."
+    # puts "Creating the objects in the data base..."
     db_populator.create_all_casa_admin("allcasaadmin@example.com")
     db_populator.create_org(CasaOrgPopulatorPresets.for_environment.merge({org_name: "Prince George CASA"}))
     db_populator.create_org(CasaOrgPopulatorPresets.minimal_dataset_options)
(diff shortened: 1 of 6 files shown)


db/seeds.rb
```diff
@@ -77,21 +77,21 @@ def get_seed_specification
...
     end
     seed
   end
 
   def report_object_counts
-    puts "\nRecords written to the DB:\n\nCount  Class Name\n-----  ----------\n\n"
+    # puts "\nRecords written to the DB:\n\nCount  Class Name\n-----  ----------\n\n"
     active_record_classes.each do |klass|
-      puts "%5d  %s" % [klass.count, klass.name]
+      # puts "%5d  %s" % [klass.count, klass.name]
```
littleforest, line 94: The `puts` statements are useful when first setting up the application or when re-seeding. Please revert these.

littleforest (changes requested):

db/seeds.rb
jhsu802701: OK, I reinstated the puts commands.

compwron:
I am ambivalent about this. @jhsu802701 can you describe more the benefit of this change? 

If you're looking for high priority work to do, check out the todo column on our issues board :) https://github.com/rubyforgood/casa/projects/1

If you're having trouble getting started, happy to do some interactive debugging in slack
And/or join office hours weds evening (Pacific time) - zoom link in slack
https://rubyforgood.herokuapp.com/ #casa channel

littleforest:
I don't mind merging this as it stands now, but I don't think folks will regularly be adding this test environment exception for future after party tasks are going to be created. So it just seems that the code is going to end up kind of inconsistent.

jhsu802701:
I'm trying to clean up the screen output from running the test suite.  So I disabled the puts commands for the lib/tasks functions in the test environment only.  (The puts commands still work in the development and production environments.)

compwron (approved):

compwron:
https://bit.ly/casaretro

PR #1317, opened 2020-11-11
remove link from My Cases if user is volunteer

### LINKED ISSUE #1270: "Assigned volunteers" should not be clickable for volunteers
relates to #1269 

AS a **volunteer**, 
I WANT the name/s under "Assigned volunteers" in CASA Case Details to be static text,
SO THAT it is clear I can not click on this. 

### Where does/should this occur on the frontend?
On the CASA Case Details page.

### How it works now
Names listed under "Assigned volunteers" are clickable for users signed in as volunteers. But clicking on it redirects to the "My Cases" page and flashes this error message:

### How it should work
Volunteers should not be able to click on names listed under "Assigned volunteers" in the CASA Case Details page. (Permissions to do this should _not_ change for admins and supervisors.)

metamoni:
Resolves #1270 

### What changed, and why?
Volunteers can't click on names in 'Assigned to' column anymore, as they don't have permissions to view/edit them anyway.

- Volunteer permissions: Volunteers can't click on names in 'Assigned to' column anymore
- Supervisor permissions: n/a
- Admin permissions: n/a

added tests to `spec/system/admin_views_dashboard_spec.rb` and `spec/system/volunteer_views_dashboard_spec.rb`
updated test in `spec/views/casa_cases/index.html.erb_spec.rb` (this test was only passing because it was checking for the wrong thing; updated it with correct details)

### Screenshots please :)
Screenshot 2020-11-11 at 22 32 01

### DIFF
--- a/spec/views/casa_cases/index.html.erb_spec.rb
+++ b/spec/views/casa_cases/index.html.erb_spec.rb
@@ -43,16 +43,18 @@ RSpec.describe "casa_cases/index", type: :system do
 
   it "Filters active/inactive casa_cases" do
     active_case = create(:casa_case, active: true, casa_org: organization)
+    active_case1 = create(:casa_case, active: true, casa_org: organization)
     inactive_case = create(:casa_case, active: false, casa_org: organization)
 
     create(:case_assignment, volunteer: volunteer, casa_case: active_case)
+    create(:case_assignment, volunteer: volunteer, casa_case: active_case1)
     create(:case_assignment, volunteer: volunteer, casa_case: inactive_case)
 
     visit casa_cases_path
     expect(page).to have_selector(".casa-case-filters")
 
     # by default, only active casa cases are shown
-    expect(page.all("table#casa-cases tbody tr").count).to eq [active_case].size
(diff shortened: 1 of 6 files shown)


spec/views/casa_cases/index.html.erb_spec.rb
```diff
@@ -42,17 +42,19 @@
   end
 
   it "Filters active/inactive casa_cases" do
-    active_case = create(:casa_case, active: true, casa_org: organization)
+    active_case = create(:casa_case, active: true, casa_org: organization, case_number: "CINA-1")
+    active_case1 = create(:casa_case, active: true, casa_org: organization, case_number: "CINA-2")
```
littleforest, line 46: @metamoni I think you had to add the case number due to the newish case-prefix filter:

Screen Shot 2020-11-11 at 2 52 01 PM

Would you mind adjusting the `casa_case` factory so that the case_number matches the `CINA-XXX` format so others don't run into the same issue you did?

littleforest (changes requested):

littleforest (approved):
Looks great! Thanks for working on this. 🎉 🚢

PR #1343, opened 2020-11-14
hide volunteer link in case contacts when current user is volunteer

### LINKED ISSUE #1271: volunteers should not be able to click names in the "Created by" column
relates to #1270 and #1269 

AS a **volunteer**,
I WANT names to appear as static text in the "Created by" column in `case_contacts`, 
SO THAT it is clear I can not click on them.

### Where does/should this occur on the frontend?
Everywhere volunteers can view `case_contacts`, including:
- CASA Case Details
- Case Contacts

### How it works now
Names of people who created `case_contacts` are clickable for volunteers in all places where `case_contacts` appear. However, when a volunteer clicks on the name, they are redirected to "My Cases" where an error flashes this message:

### How it should work
Volunteers should not be able to click on the names in the "Created by" column associated with `case_contacts`, wherever they appear. (_But_ these should still be clickable for supervisors and admins.)

metamoni:
Resolves #1271 

### What changed, and why?
All users could click volunteer names in the 'Created by' column, in the Case Contacts table. But volunteers did not have permission to view these pages, so the link has been removed and names now appear as plain text.

- Volunteer permissions: n/a
- Supervisor permissions: n/a
- Admin permissions: n/a

Tests added in `spec/system/volunteer_views_case_contacts_spec.rb` and `/spec/system/admin_views_a_case_spec.rb`

### Screenshots please :)
Screenshot 2020-11-14 at 20 11 54

### DIFF
--- a/app/views/case_contacts/_case_contact.html.erb
+++ b/app/views/case_contacts/_case_contact.html.erb
@@ -31,15 +31,19 @@
   <td>
     <span class="mobile-label">Creator</span>
     <% if policy(contact).edit? %>
-      <% if contact.creator&.supervisor? %>
-        <%= link_to "#{contact.creator&.display_name} ", edit_supervisor_path(current_user) %>
-      <% elsif contact.creator&.casa_admin? %>
-        <%= link_to "#{contact.creator&.display_name} ", edit_users_path %>
+      <% if current_user.volunteer? %>
+        <%= contact.creator&.decorate.name %>
       <% else %>
-        <%= link_to "#{contact.creator&.display_name} ", edit_volunteer_path(contact.creator) %>
+        <% if contact.creator&.supervisor? %>
+          <%= link_to contact.creator&.decorate.name, edit_supervisor_path(current_user) %>
+        <% elsif contact.creator&.casa_admin? %>
+          <%= link_to contact.creator&.decorate.name, edit_users_path %>
+        <% else %>
+          <%= link_to contact.creator&.decorate.name, edit_volunteer_path(contact.creator) %>
+        <% end %>
       <% end %>
     <% else %>
-      <%= contact.creator&.display_name %>
+      <%= contact.creator&.decorate.name %>
     <% end %>
   </td>
   <td class="read-more">
--- a/spec/system/volunteer_views_case_contacts_spec.rb
+++ b/spec/system/volunteer_views_case_contacts_spec.rb
@@ -1,0 +1,22 @@
+require "rails_helper"
+
+RSpec.describe "volunteer views case contacts", type: :system do
+  let(:organization) { create(:casa_org) }
+  let(:volunteer) { create(:volunteer, display_name: "Bob Loblaw", casa_org: organization) }
+  let(:casa_case) { create(:casa_case, casa_org: organization, case_number: "CINA-1") }
+  let!(:case_assignment) { create(:case_assignment, volunteer: volunteer, casa_case: casa_case) }
+  let!(:case_contact) { create(:case_contact, creator: volunteer, casa_case: casa_case) }
+
+  before(:each) do
+    sign_in volunteer
+    visit case_contacts_path
+  end
+
+  it "can see case creator in table" do
+    expect(page).to have_text("Bob Loblaw")
+  end
+
+  it "can navigate to edit volunteer page" do
+    expect(page).to have_no_link("Bob Loblaw")
+  end
+end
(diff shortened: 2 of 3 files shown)


app/views/case_contacts/_case_contact.html.erb
```diff
@@ -31,12 +31,16 @@
   <td>
     <span class="mobile-label">Creator</span>
     <% if policy(contact).edit? %>
-      <% if contact.creator&.supervisor? %>
-        <%= link_to "#{contact.creator&.display_name} ", edit_supervisor_path(current_user) %>
-      <% elsif contact.creator&.casa_admin? %>
-        <%= link_to "#{contact.creator&.display_name} ", edit_users_path %>
+      <% if current_user.volunteer? %>
+        <%= contact.creator&.display_name %>
```
littleforest, line 35: This should actually be using the `name` decorator, which will always exist. So you can change it to:

```
contact.creator.decorate.name
```

app/views/case_contacts/_case_contact.html.erb
```diff
@@ -31,12 +31,16 @@
...
       <% else %>
-        <%= link_to "#{contact.creator&.display_name} ", edit_volunteer_path(contact.creator) %>
+        <% if contact.creator&.supervisor? %>
+          <%= link_to "#{contact.creator&.display_name} ", edit_supervisor_path(current_user) %>
```
littleforest, line 38: Please use decorator

app/views/case_contacts/_case_contact.html.erb
```diff
@@ -31,12 +31,16 @@
...
+        <% elsif contact.creator&.casa_admin? %>
+          <%= link_to "#{contact.creator&.display_name} ", edit_users_path %>
```
littleforest, line 40: Please use decorator

app/views/case_contacts/_case_contact.html.erb
```diff
@@ -31,12 +31,16 @@
...
+        <% else %>
+          <%= link_to "#{contact.creator&.display_name} ", edit_volunteer_path(contact.creator) %>
```
littleforest, line 42: Please use decorator, also fix on line 46.

app/views/case_contacts/_case_contact.html.erb
```diff
@@ -31,12 +31,16 @@
...
+        <% end %>
```
littleforest, line 43: @metamoni for all of these cases, I know the code was like this before, but can you remove the quotes with the extra space? It's not necessary:


```
<%= link_to contact.creator.decorate.name, edit_volunteer_path(contact.creator) %>
```

etc.

spec/system/volunteer_views_case_contacts_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    sign_in volunteer
+    visit case_contacts_path
+  end
+
+  it "can see case creator in table" do
+    expect(page).to have_text("Bob Loblaw")
+  end
+
+  it "can navigate to edit volunteer page" do
+    expect(page).to have_no_link("Bob Loblaw", href: "/volunteers/#{volunteer.id}/edit")
```
littleforest, line 20: Let's remove the `href` on this test, since that could pass if there was a link but it pointed elsewhere. Since we want to test for no link no matter what, just do `expect(page).to have_no_link("Bob Loblaw")`.

littleforest (changes requested):
@metamoni this looks great! Just asking for some code cleanup for existing code if you don't mind!

app/views/case_contacts/_case_contact.html.erb
metamoni: I've been dying to fix those, just didn't want to seem like I'm fiddling with things out of scope haha. Great, will do!

app/views/case_contacts/_case_contact.html.erb
metamoni: Just to double check, the decorator added some extra bits of info to the user's name. Is this what it's supposed to look like or have I done something wrong?
Screenshot 2020-11-15 at 08 33 19

compwron: decorator name vs display_name is an interesting conversation. I don't think it's an essential distinction here. before names were required, we would default name to email when name was missing. The "org1MM" prefix is from a recent seed data change I made to help us track down multi-casa bugs.

compwron (approved):
Good tests :)

compwron (approved):

PR #1341, opened 2020-11-14
Cache deps when building CI

NickSchimek:
This can speed up build times when deps are not altered.

### What github issue is this PR for, if any?
I don't believe there is an issue for this. Is it preferable to have one?

### What changed, and why?
Cache dependencies when building on the CI server

### How will this affect user permissions?
- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

### How is this tested? (please write tests!) 💖💪
hmmm. There are no tests. If it runs it runs.

### Screenshots please :)
How about a gif?

### Feelings gif (optional)
What gif best describes your feeling working on this issue? https://giphy.com/
How to embed:
alt text

PR #1345, opened 2020-11-15
cache deps in CI

NickSchimek:
Another attempt from where I left off in PR #1341 
This can speed up build times when deps are not altered.

I don't believe there is an issue for this. Is it preferable to have one?

### What changed, and why?
Cache dependencies when building on the CI server

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

hmmm. There are no tests. If it runs it runs.

### Screenshots please :)
How about a gif?

### Feelings gif (optional)
How to embed:
alt text

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.7.1
+          bundler-cache: true
 
       - name: Install PostgreSQL client
         run: |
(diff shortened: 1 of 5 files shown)


compwron (approved):
Nice!
> if it runs, it runs
lol very true

> I don't believe there is an issue for this. Is it preferable to have one?
Kinda but not really
An issue is a representation of "yes we definitely want this" so in general I think that having an issue prevents two people doing the same work by accident. But for this thing in particular, you and I already talked, so there's no real need :)

PR #1346, opened 2020-11-15
Weekly supervisor digest mailer

### LINKED ISSUE #23: supervisor receives weekly digest email of their volunteers' updates
Part of epic #105 

**What type of user is this for? [volunteer/supervisor/admin/all**
supervisor

system email sent to supervisor

**Description**
Digest emails should include the following:
- most recent `case_contact` attempted
- number of successful `case_contacts` made
- number of unsuccessful `case_contacts` made

**Screenshots**
Here is an example of a PG CASA email:
PG CASA

And here are some examples of digest emails:
LinkedIn
Sprout Social
DropBox

**Related issues**
#125, #22

pollygee:
Resolves #23 

### What changed, and why?
Added a weekly digest email to be sent to supervisors with the week in review of the contacts attempted for each of their volunteers.  

N/A

I added a preview to for the emails so we can view the sample email.

### Screenshots please :)
Action_Controller__Exception_caught

It isn't beautiful but the info is there.  Maybe we can find someone good at making things pretty help make this look cooler!

### Feelings gif (optional)
It isn't pretty


app/decorators/casa_case_decorator.rb
```diff
@@ -25,6 +25,20 @@ def case_contacts_ordered_by_occurred_at
     object.case_contacts.sort_by(&:occurred_at)
   end
 
+  def case_contacts_latest
+    object.case_contacts.max_by(&:occurred_at)
```
compwron, line 29: This should be a scope on the model, to avoid having to pull all the case contacts every time.

app/decorators/casa_case_decorator.rb
```diff
@@ -25,6 +25,20 @@ def case_contacts_ordered_by_occurred_at
...
+  end
+
+  def successful_contacts_this_week
+    this_week = Date.today - 7.days..Date.today
```
compwron, line 33: I thought we already had something like this somewhere... I want to have all this logic in one central place. I will take a look around.

compwron: I would like it if we could split out a helper or something(?) to hold all our date-related logic like this

Existing in user: 
```
  def recent_contacts_made(days_counter = 60)
    case_contacts.where(contact_made: true, occurred_at: days_counter.days.ago..Date.today).size
  end
```

lib/tasks/send_supervisor_digest.rake
```diff
@@ -0,0 +1,6 @@
+desc "Send an email to supervisors each week to share an overview of their volunteers' activities"
+task send_supervisor_digest: :environment do
```
compwron, line 2: We'll need a cron for this or similar- I assume heroku has something for this?

compwron (approved):

lib/tasks/send_supervisor_digest.rake
pollygee: I wrote a mailer for reminders when court reports are due.  I believe I was told to create a task that would be run by heroku.  I never followed up about that to see if that was scheduled.  I would assume heroku has something for that, but I'm not super familiar.  I can look it up and try to set that up once this gets merged so we can set it up for this one and the court report reminder.

pollygee:
Thanks @compwron I'll make some updates in the next couple of days when I get a few minutes to work on this. :)

PR #1360, opened 2020-11-19
remove location link

### LINKED ISSUE #1266: Location should not be clickable
AS any user,
I WANT Location in my sidebar to be static,
SO THAT I am clear it is static data that does not lead anywhere and do not feel tempted to click on it.

### How it works now
Clicking on a CASA Org's Location in the sidebar menu brings a user to their default home view. 

Location should not be clickable, it should display in the sidebar menu as static data.

gabrielbaldao:
Resolves #1266

### What changed, and why?
I removed the location link, because the link was unnecessary

### Screenshots please :)

image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -15,8 +15,7 @@ module ApplicationHelper
   def page_header
     return default_page_header unless user_signed_in?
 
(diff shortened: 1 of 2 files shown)


compwron (approved):

compwron:
:)

PR #1361, opened 2020-11-19
Refact: remove duplicate methods in polices scopes

gabrielbaldao:
### What changed, and why?
I removed the duplicate methods inside the policies scopes.

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

I considered that the tests are already written.


compwron (approved):
Seems reasonable :) We might want to do some renaming of the base methods for clarity also eventually.

PR #1372, opened 2020-11-19
Renaming policy aliases

compwron:
This is a follow-up to https://github.com/rubyforgood/casa/pull/1361

### What changed, and why?
More naming improvements, inspired by https://github.com/rubyforgood/casa/pull/1361

no change

existing spec

### DIFF
--- a/app/policies/casa_case_policy.rb
+++ b/app/policies/casa_case_policy.rb
@@ -26,7 +26,7 @@ class CasaCasePolicy
     end
   end
 
-  def update_case_number?
+  def is_admin?
     user.is_a?(CasaAdmin)
   end
 
@@ -34,7 +34,7 @@ class CasaCasePolicy
     user.is_a?(Supervisor)
   end
 
-  def update_court_date?
+  def admin_or_supervisor?
     user.casa_admin? || user.supervisor?
   end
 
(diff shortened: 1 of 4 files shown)


app/policies/casa_case_policy.rb
```diff
@@ -68,24 +70,26 @@ def permitted_attributes
...
-  alias_method :edit?, :show?
-  alias_method :update?, :show?
-  alias_method :create?, :new?
-  alias_method :destroy?, :new?
+  alias_method :show?, :same_org_supervisor_admin_or_assigned?
+  alias_method :edit?, :same_org_supervisor_admin_or_assigned?
+  alias_method :update?, :same_org_supervisor_admin_or_assigned?
+  alias_method :new?, :same_org_supervisor_admin?
+  alias_method :create?, :same_org_supervisor_admin?
+  alias_method :destroy?, :same_org_supervisor_admin?
```
littleforest, line 92: The default style for Pundit is to have methods named `show?`, `new?` etc. What about instead of having aliases, just having a private method that calls `same_org_supervisor_admin_or_assigned`? Like:

```
def new?
  same_org_supervisor_admin_or_assigned
end

def show?
  same_org_supervisor_admin_or_assigned
end

private

def same_org_supervisor_admin_or_assigned
  is_in_same_org? && is_supervisor_or_casa_admin?
end
```

It isn't really less code, but it sticks with the style that one expects when using Pundit.

compwron: I think alias is the same as def in this case, is it not?

PR #1383, opened 2020-11-23
Change case_numbers to only be unique within a CASA org

### LINKED ISSUE #1355: Allow duplicate case numbers as long as they are for different CASA orgs
Allow duplicate case numbers as long as they are for different CASA orgs
The code that generates this line will have to change
`#  index_casa_cases_on_case_number      (case_number) UNIQUE`
We will beed a db migration and a rails validation

notapatch:
- Issue: https://github.com/rubyforgood/casa/issues/1355
 - Currently all case_numbers are unique within the application.
   We want to change it so that case_numbers are unique only
   within a CASA org. To following changes are required:
     - The `casa_cases` table needs to:
       - remove the more restrictive `case_number` index which
         requires all case_numbers to be unique
       - add new `case_number` index which is unique by
         `case_number` and `casa_org_id`
     - `CasaCase` model changes validation
     - Associated tests

---

Resolves #1355 

### What changed, and why?
Removed the index that restricted case_numbers to be unique within the application.
Added an index that restricted case_number to be unique within a CASA org.

It does not

The changes broke tests at the model and system level which I have now fixed.

I then hand tested the behavior of the index before and after the change and made
sure you can rollback without error.

### DIFF
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -16,7 +16,7 @@ class CasaCase < ApplicationRecord
   has_many :assigned_volunteers, -> { active }, through: :active_case_assignments, source: :volunteer, class_name: "Volunteer"
   has_many :case_contacts, dependent: :destroy
   has_many :past_court_dates, dependent: :destroy
-  validates :case_number, uniqueness: {case_sensitive: false}, presence: true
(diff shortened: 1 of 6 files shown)


littleforest (approved):
🎉 🚢 looks great! Thanks for working on this.

compwron (approved):
Yay!!

PR #1384, opened 2020-11-24
Add per-casa-org scope and use it in pundit policies

### LINKED ISSUE #1371: Scope per casa org in CaseContactPolicy
There should never be `.all` only `.where(casa_org: @user.casa_org)` in this code and code like it

Add tests to go with this change

```
 def resolve
      case @user
      when CasaAdmin # scope.in_casa_administered_by(@user)
        scope.all
      when Volunteer
        scope.where(casa_case: CasaCase.actively_assigned_to(@user), creator: @user)
      when Supervisor
        scope.all
      else
        raise "unrecognized user type #{@user.type}"
      end
```

gabrielbaldao:
Resolves #1371 

### What changed, and why?
As asking in #1371 I'm changing the scopes to returns just `.where(casa_org: organization)`

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Feelings gif (optional)
How to embed:
`alt text`


app/policies/casa_case_policy.rb
```diff
@@ -17,7 +17,7 @@ def initialize(user, scope)
     def resolve
       case @user
       when CasaAdmin, Supervisor
-        scope
+        scope.where(casa_org: @user.casa_org)
```
wenderjean, line 20: Would it make sense given the context to move it into a `scope` in the model?

compwron: yes please :)

compwron:
How's it going? I am excited to merge this! It looks like right now there are some spec failures and a small merge conflict. Please let me know here or in slack if you'd like help. And we have office hours every wednesday at 6pm pacific time if you want to come by and ask with voice/video.

gabrielbaldao:
Sorry about it @compwron , are you sure about use where(casa_org:) in every case?
Like, for case contact should be where(casa_case:) right?

compwron:
> Sorry about it @compwron , are you sure about use where(casa_org:) in every case?
> Like, for case contact should be where(casa_case:) right?

@gabrielbaldao I don't understand what your question is. Don't worry about "Would it make sense given the context to move it into a scope in the model?"- let's just get the functionality merged :)

compwron (approved):
Yes yes yes! Very cool :)
We need more tests around this to make sure we never leak between orgs... 😅

PR #1387, opened 2020-11-24
1258 - Hide casa case prefix filter to volunteer user

### LINKED ISSUE #1258: Do not show casa case prefix filter to volunteer
volunteer

**Description**
Do not show casa case prefix filter to volunteer, because they will only ever have 1 or 2 cases

Screen Shot 2020-11-04 at 11 47 21 PM

notapatch:
- do not show casa case prefix filter to volunteer, because
   they will only ever have 1 or 2 cases

---

Resolves #1258

### What changed, and why?
Added a conditional around "Casa case prefix filter". Originally I had the conditional around the dropdown but if the dropdown is missing the case rows are filtered out.

It won't

So testing for the absence of something, in this case a button, is tricky. Instead I tested for the presence of the button in the administration mode. I could add a test for volunteer user that was that it didn't find the button?

### Screenshots please :)
# Volunteer screen
1258-volunteer

# Admin screen
1258-admin

### DIFF
--- a/app/views/casa_cases/index.html.erb
+++ b/app/views/casa_cases/index.html.erb
@@ -60,13 +60,15 @@
       </div>
     </div>
     <div class="dropdown pull-left mr-2">
-      <button class="btn btn-secondary dropdown-toggle show" type="button" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-        Casa Case Prefix
-      </button>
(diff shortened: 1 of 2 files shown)


littleforest (changes requested):
@notapatch this is great, thanks! Would you mind adding a "volunteer" context in the spec and checking that the "Casa Case Prefix" does not show up in that case?

littleforest (approved):
🎉 Thanks for adding that extra spec! 🌟

notapatch:
@littleforest marvelous!

PR #1402, opened 2020-11-26
Autoformat views folder + centralize date formats + cleanup

compwron:
Autoformat views folder
Extract allowed to update name to policy file
Fix erb line length, extract method case_contact.decorate.show_contac… …
Remove usages of user decorator name since display_name is always pre… …
move policy method edit_name? to User for future reuse and because th… …
Remove unused magic variable in admin view of volunteers Pick Display… …
Fix policy typo that breaks so very many things
Add TODOs, add tooltip for clock icon, fix failing usages of deleted … …
Refactor date formats
Fix invalid html
Fix view usage of controller
Null checks and less data in seed spec
Add back css import
Cleanup old TODOs


compwron:
I want to give this a full week to soak in qa/staging so I won't merge it until after the friday deploy

app/helpers/date_format.rb
```diff
@@ -0,0 +1,6 @@
+module DateFormat
+  MM_DD_YYYY = "%m-%d-%Y"
+  CONTACT_OCCURRED_AT = "%Y%m%d%H%M%s"
+  FULL = "%B %-d, %Y" # no leading 0
+  YOUTH_DATE_OF_BIRTH = "%B %Y" # never show day of month for youth date of birth
+end
```
littleforest, line 6: @compwron we're using [I18n for formatting dates](https://guides.rubyonrails.org/i18n.html#adding-date-time-formats) in two other places:

https://github.com/rubyforgood/casa/blob/main/app/views/casa_cases/_form.html.erb
https://github.com/rubyforgood/casa/blob/main/app/views/casa_cases/show.html.erb

Do you want to use that method instead, or if not, then probably change these ^^ two cases to use your new DateFormat module.

compwron: Oh, interesting. Right now I'm leaning towards doing the dateformat module above because it's so customizable and if not easy to read then at least easy to look up. But I do really like the `format: :day_and_date` syntax... well regardless, I do want some centralization of format. I'll think about this for a bit. :) What do you prefer?

littleforest: Shrug. I always use the i18n format so obviously partial to that, but doesn't really matter. You can name the keys whatever you like using i18n, so still customizable. One of the benefits of using i18n is that if you internationalize the application, then it's easy to have dates display in formats used by other countries/languages, but that doesn't seem like a benefit we would need for CASA.

compwron: I am convinced
I don't know if I have the energy to do it rn so I made https://github.com/rubyforgood/casa/issues/1408

PR #1406, opened 2020-11-27
Fix time in rack attack specs to avoid flaky tests

### LINKED ISSUE #1379: Fix Rack Attack specs on Docker
A few rack attack specs are failing when run on Docker: `--seed=40871`

littleforest:
Resolves #1379 

### What changed, and why?
Rack Attack creates a bucket based on the timestamp and counts the number of hits in that bucket. Since the tests may cross into the next bucket as time passes, the test for blocking a request sometimes fails. So in the testing environment, use a fixed timestamp so only one bucket is used, and then check that too many hits in the bucket successfully throttles the request.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -46,10 +46,6 @@ Rails.application.configure do
   config.active_support.deprecation = :stderr
 
   # Rack Attack configuration
-  #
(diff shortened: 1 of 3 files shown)


compwron (approved):
Nice :)

Issue #1352, opened 2020-11-16
paginate db query results and use that pagination on the frontend


compwron:
**What type of user is this for? volunteer/supervisor/admin/all**
all, but improvements will be most visible for supervisor and admin

**Description**
paginate db query results and use that pagination on the frontend
The point of this is to make page load faster for supervisor & admin
This was prompted by https://ideaboardz.com/for/R4G%20CASA%20Nov%202020%20retro%2Ffuturespective/3512118

**Screenshots of current behavior, if any**

patrickarnett:
Can I jump on this one?

patrickarnett:
@compwron This spec is actually testing that the cell is blank when the joining `supervisor_volunteer` is inactive. Should the cell be blank when the `supervisor` itself is inactive, as well?
https://github.com/rubyforgood/casa/blob/b29b1427b1a6e92fbb74468ac744f84ba0a767f8/spec/system/admin_views_volunteers_page_spec.rb#L141-L151

compwron:
Yes, if the volunteer's supervisor is inactive then this cell should be blank.
The cell should also be blank if the volunteer used to be, but is no longer associated to the supervisor i.e. what it's actually testing now)

patrickarnett:
It's outside the scope of this issue, but should the deactivation of a `supervisor` or `volunteer` in turn deactivate any associated `supervisor_volunteers` and/or `case_assignments`?

compwron:
Closing this because it is done in PR #1412

PR #1412, opened 2020-11-29
Paginate volunteers index

patrickarnett:
#1352 (this PR does not fully resolve the issue)

### What changed, and why?
Instead of pulling all (e.g.) of an org's volunteers from the DB and dumping them in the view, tables will load filtered and paginated data via ajax.

n/a

Datatable specs are added; controller, decorator, model, request, system, and view specs are updated.


littleforest (changes requested):
Hi @patrickarnett thanks for your PR. Haven't had time to go over the whole thing, but one request I have is to modify the methods like `made_contact_with_all_cases_in_14_days` so that the number of days is not embedded in the method name. The "14" in this case should really be set as a constant and should be something we can easily change if we would instead like to start tracking contacts made within a week, for example.

patrickarnett:
@littleforest I updated that to use associations so that eager loading would prevent N+1 queries, but Bullet says it was a bad idea so I'll revert it to the original method. Should that `14` still be a constant?

littleforest:
@patrickarnett yes it would be great if you could set 14 as a constant. I know it wasn't already set as one, but it should be.

littleforest:
@patrickarnett could you also change the `contacts_made_in_past_60_days` method as well?

jmkoni:
@patrickarnett is there a reason you didn't want to use a gem like kaminari?

patrickarnett:
@jmkoni That's what I figured I'd go with when taking the issue, but the DataTable plugin used on the front end has its own pagination functionality. With no need for the view helper, `offset`ing, `limit`ing, and `count`ing are easy enough to handle without Kaminari (imo).

I did attempt to implement the existing server-side gem (https://github.com/jbox-web/ajax-datatables-rails), but I was unable to get it to filter and sort correctly (not only on some of the more complex columns, but even on supervisor name). What I've got here is an imitation.

compwron:
Oh wow this is a LOT! I am very excited. :) I will go through it in detail tonight after work. :)

app/datatables/volunteer_datatable.rb
```diff
@@ -0,0 +1,176 @@
...
+  def contacts_made_in_past_days_subquery
+    @contacts_made_in_past_days_subquery ||=
+      CaseContact
+        .select(
+          <<-SQL
+          creator_id,
+          COUNT(*) AS contact_count
+          SQL
+        )
+        .where(contact_made: true, occurred_at: Volunteer::CONTACT_MADE_IN_PAST_DAYS_NUM.days.ago.to_date..)
```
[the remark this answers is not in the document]
compwron: Something is wrong in the "Contacts made in past 60 days" column (running this code locally) but honestly I don't know if that's caused by this PR or not. At least nothing explodes or OOMs here and that reassures me about the `.days.ago.to_date..` 
Screen Shot 2020-12-01 at 7 01 48 PM

app/decorators/user_decorator.rb
```diff
@@ -7,21 +7,10 @@ def status
     "Inactive"
   end
 
-  # If all of a volunteers cases are not transition youth eligible, then
-  # we return "No", otherwise they have at least one transition youth eligible case
-  # and we return "Yes"
-  def assigned_to_transition_aged_youth?
```
compwron, line 13: aha, replaced with sql calculation. Nice

app/decorators/user_decorator.rb
```diff
@@ -7,21 +7,10 @@ def status
...
-    volunteer_no_transition_youth_cases ? "No" : "Yes 🐛🦋"
-  end
-
-  def last_contact_made
-    if object.most_recent_contact.nil?
-      "None ❌"
-    else
-      object.most_recent_contact.occurred_at.strftime(DateFormat::FULL)
-    end
+    object.display_name
```
compwron, line 13: This is not needed, decorator will "fall through" to allow the  `display_name` method on the underlying object to be called. It would be better to call `display_name` in the UI than alias it here.

app/javascript/src/dashboard.js
```diff
@@ -10,51 +10,6 @@ var defineCaseContactsTable = function () {
 }
 
 $('document').ready(() => {
-  $.fn.dataTable.ext.search.push(
```
compwron, line 13: Yay less js! (I would prefer testing sql and ruby to testing js in this codebase, we are better at it)

app/javascript/src/dashboard.js
```diff
@@ -116,24 +71,141 @@ $('document').ready(() => {
...
 
+  const handleAjaxError = e => {
+    if (e.status === 401) {
+      location.reload();
+    } else {
+      console.log(e);
+      if (e.responseJSON && e.responseJSON.error) {
+        alert(e.responseJSON.error);
+      } else if (e.response.statusText) {
+        alert(`Unable to complete your request: ${e.response.statusText}`);
```
compwron, line 82: We will want to make this message more user-friendly - something like "Sorry, try that again?" and "Sorry, if you're seeing a problem, please fill out the Report A Site Issue 
link to the bottom left near your email address" (is that too long?)
http://meetcontent.com/blog/the-little-things-why-microcopy-matters/

app/javascript/src/dashboard.js
```diff
@@ -116,24 +71,141 @@ $('document').ready(() => {
...
+
   // Enable all data tables on dashboard but only filter on volunteers table
+  const editSupervisorPath = id => `/supervisors/${id}/edit`;
+  const editVolunteerPath = id => `/volunteers/${id}/edit`;
+  const casaCasePath = id => `/casa_cases/${id}`;
   var volunteersTable = $('table#volunteers').DataTable({
     autoWidth: false,
     stateSave: false,
-    columnDefs: [
+    columns: [
```
compwron, line 96: Is there any particular reason this changed? Is this a datatables version thing?

app/javascript/src/dashboard.js
```diff
@@ -116,24 +71,141 @@ $('document').ready(() => {
...
       {
-        targets: [1],
+        name: "display_name",
+        render: (data, type, row, meta) => {
+          return `
+            <a href="${editVolunteerPath(row.id)}">
+              ${row.display_name || row.email}
+              ${row.made_contact_with_all_cases_in_days === 'false'
+                ? `🕐 <i class="fa fa-question-circle" aria-hidden="true" data-toggle="tooltip" title="Has at least one case with no contact in 14 days"></i>`
```
compwron, line 104: This tooltip is unclickable because it is included in the volunteers link. This needs to get fixed. (was it always broken like that? Yes I know I'm the one who wrote it, sorry y'all 😅)

app/javascript/src/dashboard.js
```diff
@@ -116,24 +71,141 @@ $('document').ready(() => {
...
+          });
+
+          return links.join(", ");
+        },
+        orderable: false
+      },
+      {
+        name: "most_recent_contact_occurred_at",
+        render: (data, type, row, meta) => {
+          if (!row.most_recent_contact.case_id) return "None ❌";
```
compwron, line 153: Please always use brackets with if statements in javascript. Right now the js linter is not enabled in CI but hopefully it will catch this when we turn it on.

app/javascript/src/dashboard.js
```diff
@@ -116,24 +71,141 @@ $('document').ready(() => {
...
+            <a href="${editVolunteerPath(row.id)}">
+              Edit
+            </a>
+          `
+        },
+        searchable: false
       }
-    ]
+    ],
+    processing: true,
```
compwron, line 183: We need capybara test coverage for this or I worry that it will break the next time we update datatables

app/models/volunteer.rb
```diff
@@ -8,7 +8,8 @@ class Volunteer < User
   ASSIGNED_TO_TRANSITION_AGED_YOUTH_COLUMN = "assigned_to_transition_aged_youth"
   CASE_NUMBER_COLUMN = "case_number"
   LAST_CONTACT_MADE_COLUMN = "last_contact_made"
-  CONTACT_MADE_IN_PAST_60_DAYS_COLUMN = "contact_made_in_past_60_days"
+  CONTACT_MADE_IN_PAST_DAYS_NUM = 60
+  CONTACT_MADE_IN_PAST_DAYS_COLUMN = "contact_made_in_past_#{CONTACT_MADE_IN_PAST_DAYS_NUM}_days".freeze
```
compwron, line 12: yay variable CONTACT_MADE_IN_PAST_DAYS_NUM

app/models/volunteer.rb
```diff
@@ -60,21 +62,21 @@ def supervised_by?(supervisor)
     self.supervisor == supervisor
   end
 
-  def made_contact_with_all_cases_in_days?(num_days = 14)
+  # false if volunteer has any case with no contact in the past 30 days
+  def made_contact_with_all_cases_in_days?(num_days = CONTACT_MADE_IN_DAYS_NUM)
     # TODO this should do the same thing as no_contact_for_two_weeks but for a volunteer
-    total_cases_count = casa_cases.count
+    total_cases_count = casa_cases.size
```
compwron, line 68: interesting, why size vs count? is one of them more database-y?

app/models/volunteer.rb
```diff
@@ -60,21 +62,21 @@ def supervised_by?(supervisor)
...
 
   private
 
-  def cases_where_contact_made_in_days(num_days = 14)
+  def cases_where_contact_made_in_days(num_days = CONTACT_MADE_IN_DAYS_NUM)
     casa_cases
       .joins(:case_contacts)
-      .where(case_contacts: {contact_made: true})
-      .where("case_contacts.occurred_at > ?", Date.current - num_days.days)
+      .where(case_contacts: {contact_made: true, occurred_at: 14.days.ago.to_date..})
```
compwron, line 79: let's use the `CONTACT_MADE_IN_DAYS_NUM ` variable instead of `14` here

app/views/volunteers/index.html.erb
```diff
@@ -24,12 +24,12 @@
...
               id="unassigned-vol-filter"
               type="checkbox"
-              data-value="No supervisor assignment"
+              data-value=""
               unchecked>
             No supervisor assignment
           </a>
         </li>
-        <% Supervisor.decorate.each do |supervisor| %>
+        <% current_organization.supervisors.each do |supervisor| %>
```
compwron, line 32: YAY

app/views/volunteers/index.html.erb
```diff
@@ -24,12 +24,12 @@
             <input
               id="unassigned-vol-filter"
               type="checkbox"
-              data-value="No supervisor assignment"
```
compwron, line 27: Why did this go away?

spec/datatables/volunteer_datatable_spec.rb
```diff
@@ -0,0 +1,430 @@
...
+      additional_filters: additional_filters,
+      order_by: order_by,
+      order_direction: order_direction,
+      page: page,
+      per_page: per_page,
+      search_term: search_term
+    )
+  end
+
+  before do
```
compwron, line 32: this runs before each test. Since we're only testing filtering not editing, maybe we can run this only once for this file instead of once per test in the file. https://relishapp.com/rspec/rspec-core/v/2-2/docs/hooks/before-and-after-hooks `before(:all)`

spec/datatables/volunteer_datatable_spec.rb
```diff
@@ -0,0 +1,430 @@
+require "rails_helper"
```
compwron, line 1: beauuuuutiful tests :)

spec/factories/volunteer.rb
```diff
@@ -38,5 +38,11 @@
         create(:supervisor_volunteer, :inactive, volunteer: user)
       end
     end
+
+    trait :with_inactive_supervisor_assignment do
```
compwron, line 42: :)

spec/rails_helper.rb
```diff
@@ -45,6 +46,10 @@
 
   config.infer_spec_type_from_file_location!
 
+  config.define_derived_metadata(file_path: Regexp.new("/spec/datatables/")) do |metadata|
```
compwron, line 49: What is this needed for? Maybe we should add a comment here.

spec/requests/volunteers_spec.rb
```diff
@@ -13,6 +13,28 @@
...
+    end
+
+    it "renders json data" do
+      sign_in admin
+
+      post datatable_volunteers_path
+      expect(response.body).to eq data.to_json
+    end
+  end
+
```
compwron, line 37: :)

spec/system/volunteers/index_spec.rb
```diff
@@ -65,7 +65,7 @@
     it "can filter volunteers" do
       assigned_volunteers = create_list(:volunteer, 3, :with_assigned_supervisor, casa_org: organization)
       inactive_volunteers = create_list(:volunteer, 2, :inactive, casa_org: organization)
-      unassigned_volunteers = create_list(:volunteer, 1)
+      unassigned_volunteers = create_list(:volunteer, 1, casa_org: organization)
```
compwron, line 68: 😅🙌🏼

spec/system/volunteers/index_spec.rb
```diff
@@ -146,11 +152,23 @@
...
         click_on "Supervisor"
         find(:css, "#unassigned-vol-filter").set(true)
-        supervisor_cell = page.find(".supervisor-column")
+        supervisor_cell = page.find("tbody .supervisor-column")
 
         expect(supervisor_cell.text).to eq ""
       end
     end
+
+    context "when timed out" do
```
compwron, line 161: nice :)

spec/system/volunteers/index_spec.rb
```diff
@@ -173,13 +191,13 @@
...
 
       click_on "Status"
-      find(:css, 'input[data-value="Active"]').set(false)
-
-      # when all users are hidden, the tr count will be 1 for "no results" row
-      expect(page.all("table#volunteers tbody tr").count).to eq 1
-
-      find(:css, 'input[data-value="Inactive"]').set(true)
+      find(:css, 'input[data-value="true"]').set(false)
+      expect(page).to have_text("No matching records found")
```
compwron, line 195: We might want to make this copy more friendly- something like "No volunteers match your filters. Try removing some filters and try again!"

spec/system/volunteers/index_spec.rb
```diff
@@ -223,5 +241,17 @@
...
+      it "prompts login" do
+        sign_in supervisor
+        visit volunteers_path
+        click_on "Supervisor"
+        allow_any_instance_of(User).to receive(:timedout?).and_return true
+        find(:css, "#unassigned-vol-filter").set(true)
+        expect(page).to have_text "You need to sign in before continuing."
+        expect(current_path).to eq new_user_session_path
+      end
+    end
```
compwron, line 255: I am very happy to see tests for this. You rock :)

spec/views/volunteers/index.html.erb_spec.rb
```diff
@@ -22,8 +23,4 @@
 
     it { is_expected.to have_selector("a", text: "New Volunteer") }
   end
-
-  context "links" do
```
compwron, line 26: Yep, this is very covered in the other test.

compwron (approved):
WOW
I like it :)

I will open a github issue to track the comments above so we can get them done in future PRs. I will assign it to you but feel free to put it down if you don't have time.

compwron:
I made https://github.com/rubyforgood/casa/issues/1417 to follow up on the above comments.

app/datatables/application_datatable.rb
```diff
@@ -0,0 +1,88 @@
...
+  def order_by
+    @order_by ||=
+      lambda {
+        order_by = params[:columns][order_column_index][:name]
+        order_by if self.class::ORDERABLE_FIELDS.include? order_by.try :downcase
+      }.call
+  end
+
+  def order_column_index
+    params[:order]["0"][:column]
```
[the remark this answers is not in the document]
patrickarnett: @compwron I'm also not a fan! But it's part of the mess of params DataTable sends through. By default it sends
```ruby
{
  "draw" => "1",
  "columns" => {
    "0" => {
      "data" => "0",
      "name" => "display_name",
      "searchable" => "true",
      "orderable" => "true",
      "search" => {
        "value" => "",
        "regex" => "false"
      }
    },
    ...,
  },
  "order" => {
    "0" => {
      "column" => "0",
      "dir" => "asc"
    }
  },
  "start" => "0",
  "length" => "10",
  "search" => {
    "value" => "",
    "regex" => "false"
  }
}
```

app/datatables/application_datatable.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def offset
+    params[:start].to_i
+  end
+
+  def bool_filter(filter)
+    # expects filter to be an array
+
+    if filter.blank?
+      "FALSE"
```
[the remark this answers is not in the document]
patrickarnett: @compwron `filter` is submitted from the front end in the `additional_params` param. It's used for the supervisor/status/etc drop downs.

app/datatables/volunteer_datatable.rb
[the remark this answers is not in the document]
patrickarnett: @compwron Yes, it's an endless range. ActiveRecord interprets `where(occurred_at: start_date..)` to SQL as `WHERE occurred_at >= start_date`.

Can you clarify what problem you're seeing in that column?

app/models/volunteer.rb
patrickarnett: @compwron When `casa_cases` have already been loaded, `size` pulls the length of the relation, whereas `count` issues another DB query.

app/javascript/src/dashboard.js
patrickarnett: @compwron I honestly don't 100% remember why I made the change. _But_, `columnDefs` requires a `targets` property indicating the column's position in the table. `columns`, on the other hand, just uses the entered order, and each column requires its own configuration anyway.

app/decorators/user_decorator.rb
patrickarnett: @compwron I think this was a mistake; I didn't end up using any decorators at `#index`.

app/datatables/application_datatable.rb
```diff
@@ -0,0 +1,88 @@
...
+  def additional_filters
+    @additional_filters ||= params[:additional_filters] || {}
+  end
+
+  def search_term
+    @search_term ||= params[:search][:value]
+  end
+
+  def build_order_clause
+    Arel.sql "#{order_by} #{order_direction}" if order_by.present?
```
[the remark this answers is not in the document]
patrickarnett: @compwron I try to avoid it, since **A**rel **R**eally **E**xasperates **L**ogicians, but alas...
```
DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called
with non-attribute argument(s): "COALESCE(users.display_name, users.email) ASC". Non-attribute arguments
will be disallowed in Rails 6.1. This method should not be called with user-provided values, such as
request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql()
```

app/views/volunteers/index.html.erb
patrickarnett: @compwron Filtering is now done server-side, and the empty value is easier to handle.

spec/rails_helper.rb
patrickarnett: @compwron It just saves the effort of adding `type: :datatable` to spec files in `/spec/datatables` so that the `DatatableHelper` will be included by default.

app/models/volunteer.rb
patrickarnett: 🤦

PR #1418, opened 2020-12-02
Added validations for court report submission date and status

### LINKED ISSUE #1389: Add validation for court_report fields
**Description**
There are two related fields on the `casa_case` model, the `court_report_submitted_at` and `court_report_status` field. Add a validation to make sure that:

1. If `court_report_submitted_at` is not null, then `court_report_status` is something other than "not submitted"
2. If `court_report_submitted_at` is null, then `court_report_status` is "not submitted"

Rockenfels:
Resolves #1389 

### What changed, and why?
I added a custom validation class that extends `ActiveModel::Validator` and uses conditional logic to verify that `court_report_status` is appropriately set relative to the value of `court_report_submitted_at`. Now `court_report_status` must be undefined unless a time value is assigned to `court_report_submitted_at`.

- Volunteer permissions: n/a
- Supervisor permissions: n/a
- Admin permissions: n/a

Tests have been added to the `casa_case_spec.rb` file (lines 233-254) that verify the functionality of the changes addd.

### Screenshots please :)
image

image

image

### Feelings gif (optional)

`alt text`

### DIFF
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -1,5 +1,3 @@
-require_relative "../validators/court_report_validator"
-
 class CasaCase < ApplicationRecord
   TABLE_COLUMNS = %w[
     case_number
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -1,4 +1,4 @@
-require_relative '../validators/court_report_validator'
+require_relative "../validators/court_report_validator"
 
 class CasaCase < ApplicationRecord
   TABLE_COLUMNS = %w[
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -66,6 +66,22 @@ class CasaCase < ApplicationRecord
   delegate :name, to: :hearing_type, prefix: true, allow_nil: true
   delegate :name, to: :judge, prefix: true, allow_nil: true
 
+  # Validation class for :court_report_status & :court_report_submitted_at. Adds specific errors based on record data.
+  class CourtReportValidator < ActiveModel::Validator
+    def validate(record)
+      # Check for no timestamp and a submission value other than 'not_submitted'.
+      if record.court_report_submitted_at.nil? && record.court_report_status != "not_submitted"
+        record.errors[:court_report_status] << "Court report submission date can't be nil if status is anything but not_submitted."
+      # Check for timestamp with a 'not_submitted' status.
+      elsif !record.court_report_submitted_at.nil? && record.court_report_status == "not_submitted"
+        record.errors[:court_report_submitted_at] << "Submission date must be nil if court report status is not submitted."
+      end
+    end
+  end
+
+  # Validation to check timestamp and submission status of a case
+  validates_with CourtReportValidator, fields: [:court_report_status, :court_report_submitted_at]
+
   def court_report_status=(value)
     super
     if court_report_not_submitted?
--- a/app/models/casa_case.rb
+++ b/app/models/casa_case.rb
@@ -1,3 +1,5 @@
+require_relative '../validators/court_report_validator'
+
 class CasaCase < ApplicationRecord
   TABLE_COLUMNS = %w[
     case_number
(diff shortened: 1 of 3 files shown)


app/models/casa_case.rb
```diff
@@ -66,6 +66,22 @@ class CasaCase < ApplicationRecord
   delegate :name, to: :hearing_type, prefix: true, allow_nil: true
   delegate :name, to: :judge, prefix: true, allow_nil: true
 
+  # Validation class for :court_report_status & :court_report_submitted_at. Adds specific errors based on record data.
+  class CourtReportValidator < ActiveModel::Validator
+    def validate(record)
+      # Check for no timestamp and a submission value other than 'not_submitted'.
+      if record.court_report_submitted_at.nil? && record.court_report_status != "not_submitted"
```
littleforest, line 73: Let's use the built-in enum methods to make it a bit cleaner:

```
if record.court_report_submitted_at.nil? && !record.court_report_not_submitted?
```

app/models/casa_case.rb
```diff
@@ -66,6 +66,22 @@ class CasaCase < ApplicationRecord
...
+        record.errors[:court_report_status] << "Court report submission date can't be nil if status is anything but not_submitted."
+      # Check for timestamp with a 'not_submitted' status.
+      elsif !record.court_report_submitted_at.nil? && record.court_report_status == "not_submitted"
```
littleforest, line 76: I think it's also easier to not read double-negatives:

```
elsif record.court_report_submitted_at && record.court_report_not_submitted?
```

Since the `court_report_submitted_at` is a timestamp, it can only be `nil` or have a value, so it's sufficient to just check for `record.court_report_submitted_at`.

spec/models/casa_case_spec.rb
```diff
@@ -229,4 +229,27 @@
       expect(casa_case.reload.assigned_volunteers).to eq [volunteer2]
     end
   end
+
+  describe "report submission" do
+    # Creating a case whith a status other than not_submitted and a nil submission date
+    it "rejects cases with a court report status, but no submission date" do
+      casa_org = create(:casa_org)
+      bad_case = create(:casa_case, casa_org: casa_org)
+      bad_case.court_report_status = 2
```
littleforest, line 238: For enums you should always use the string or symbol value:

```
bad_case.court_report_status = :in_review
```

spec/models/casa_case_spec.rb
```diff
@@ -229,4 +229,27 @@
...
+      bad_case.court_report_submitted_at = nil
+      bad_case.valid?
+
+      expect(bad_case.errors[:court_report_status]).to include("Court report submission date can't be nil if status is anything but not_submitted.")
+    end
+
+    it "rejects cases with a submission date, but no status" do
+      casa_org = create(:casa_org)
+      bad_case = create(:casa_case, casa_org: casa_org)
+      bad_case.court_report_status = 0
```
littleforest, line 248: Again, use string or symbol:

```
bad_case.court_report_status = :not_submitted
```

littleforest (changes requested):
@Rockenfels this is great! Thanks so much for working on this. Just some semantic change requests.

Also, can you make an `app/validators` directory and move your new validator class there?

Rockenfels:
@littleforest thank you for the feedback and semantic upgrades! All changes have been made and committed.

app/models/casa_case.rb
```diff
@@ -1,3 +1,5 @@
+require_relative "../validators/court_report_validator"
```
littleforest, line 1: @Rockenfels I don't think this line is necessary since everything in the `app` directory is loaded by default -- can you try it without this, or did you already hit an error without it?

Rockenfels: @littleforest you're absolutely right that the `require_relative` is unneeded-- I removed it and committed the change.

littleforest (approved):
🎉 💯 🚢

PR #1425, opened 2020-12-03
Safe multi-tenancy in case contacts report

### LINKED ISSUE #1340: [multi-tenancy bug] Case Contacts Report has leaky CASA org data
Report has wrong casa org data

Logged in as admin for casa 2, seeing casa 1 results

Screen Shot 2020-11-13 at 11 08 58 PM

**How to replicate bug**
1 - Log into QA as an admin belonging to `casa_org_2`
2 - Click on "generate reports" 
3 - do not change any of the conditions for the report
4 - click "download report"
5 - open the .csv, make note of the case number listed here. It does not match that of the only case number in this casa_org's instance.

Thrillberg:
Resolves #1340

### What changed, and why?
I've added a `casa_org` scope to `CaseContact`, with the hopes that we can filter out all `CaseContact`s from other CASA organizations. Unfortunately, it's not working yet (but I wanted to get this wip up in case anyone can give feedback or take over if I run out of time).

I think it shouldn't affect permissions

I wrote a test that is now passing, but it's broken when I test it in development. Anyone have any ideas why or what I'm doing wrong?

### Feelings gif (optional)

ambivalent dog with flag

### DIFF
--- a/app/models/case_contact_report.rb
+++ b/app/models/case_contact_report.rb
@@ -9,6 +9,7 @@ class CaseContactReport
     CaseContact
       .supervisors(args[:supervisor_ids])
       .creators(args[:creator_ids])
+      .casa_org(args[:casa_org_id])
       .occurred_between(args[:start_date], args[:end_date])
       .contact_made(args[:contact_made])
       .has_transitioned(args[:has_transitioned])
--- a/app/controllers/case_contact_reports_controller.rb
+++ b/app/controllers/case_contact_reports_controller.rb
@@ -18,12 +18,18 @@ class CaseContactReportsController < ApplicationController
   private
 
   def report_params
-    parameters = params.require(:report).permit(:start_date, :end_date,
-      :contact_made, :has_transitioned, :want_driving_reimbursement,
-      {contact_type_ids: []},
-      {contact_type_group_ids: []},
-      {creator_ids: []},
-      {supervisor_ids: []})
+    parameters = params.require(:report).permit(
+      :start_date,
+      :end_date,
+      :contact_made,
+      :has_transitioned,
+      :want_driving_reimbursement,
+      contact_type_ids: [],
+      contact_type_group_ids: [],
+      casa_org_id: current_organization.id,
+      creator_ids: [],
+      supervisor_ids: []
+    )
     convert_radio_options_to_boolean(parameters)
     parameters
   end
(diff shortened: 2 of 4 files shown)


app/models/case_contact_report.rb
```diff
@@ -9,6 +9,7 @@ def filtered_case_contacts(args)
     CaseContact
       .supervisors(args[:supervisor_ids])
       .creators(args[:creator_ids])
+      .casa_org(args[:casa_org_id])
```
littleforest, line 12: @Thrillberg thanks for taking a look at this! We actually don't want to be able to pass in a different casa org (an admin for a particular casa org shouldn't ever be able to get info for a different organization). Instead I think you want to take your casa_org scope and apply it, something like:

```
CaseContact
  .joins(:casa_case).where(casa_cases: { casa_org: current_organization })
  .supervisors(...)
  .creators(...)
 ...
```

Thrillberg: Thanks for your help on this @littleforest! I'm a little confused about how the `CaseContactReport` would have access to the `current_organization` variable. My understanding was that `current_organization` is available to controllers and from the controller would have to be passed into any collaborators like this.

If you look at the `CaseContactReportsController` change I made, the `casa_org_id` is always the current organization's id and is not changeable from the front end. (That is the intent of my code anyway. Whether it actually works that way may be questionable! 😆)

littleforest: Whoops, @Thrillberg you are totally right! Great. As long as the `casa_org_id` from the front-end is not changeable (or hackable) then all is good.

compwron:
This looks like it works! Does it not work?
You have a brand new spec merge conflict because I just merged another PR that touched nearby unrelated code, sorry. 
Thank you for working on this! 💪✨

Thrillberg:
@compwron Here's a screenshot of the report I get when I'm signed in as a CASA 2 admin. It's worth noting that in the report preparation screen, CASA org 1 volunteers and supervisors were listed in the form. So the multitenancy is broken elsewhere, too, I think.
Screen Shot 2020-12-05 at 7 12 38 PM
I've gone ahead and updated this branch against `main` so that if we want to merge it, it's ready to go!

compwron (approved):
Oy. Well, it's an improvement at least... we will do this! :)

PR #1428, opened 2020-12-07
added no js warning

FireLemons:
### What changed, and why?
All pages warn if javascript is disabled

### Screenshots please :)
image


compwron (approved):
:)

PR #1429, opened 2020-12-10
Add after party task to populate emancipation data

littleforest:
### What changed, and why?
* Added uniqueness for emancipation_category and name for emancipation_option at the app level instead of just the db level
* Modified the emancipation seeding data to be idempotent
* Added after party task to import the emancipation data


spec/models/emancipation_option_spec.rb
```diff
@@ -6,6 +6,12 @@
   it { is_expected.to have_many(:casa_cases).through(:casa_cases_emancipation_options) }
   it { is_expected.to validate_presence_of(:name) }
 
+  it "is unique across emancipation_category, name" do
+    eo = create(:emancipation_option)
+    eo_new = build(:emancipation_option, emancipation_category: eo.emancipation_category, name: eo.name)
+    expect(eo_new.valid?).to be false
+  end
+
```
FireLemons, line 14: I think I already wrote this starting at line 21. One of the tests needs to go

db/seeds/emancipation_data.rb
```diff
@@ -1,87 +1,87 @@
 # Emancipation Checklist Form Data
-category_housing = EmancipationCategory.create(name: "Youth has housing.", mutually_exclusive: true)
+category_housing = EmancipationCategory.where(name: "Youth has housing.").first_or_create(mutually_exclusive: true)
```
FireLemons, line 2: While you're editing stuff, I think we wanted mutually exclusive to be false for housing.

FireLemons (changes requested):

spec/models/emancipation_option_spec.rb
littleforest: Ah, yes. You are testing the error at the database level, but now I've stopped the validation at the Rails level, so I'll remove the db checking one.

db/seeds/emancipation_data.rb
FireLemons: If i'm interpreting this correctly image

littleforest: @compwron can a user have more than one housing option selected?

littleforest: @compwron I set the housing to mutually exclusive -- if only one option should be selected, then please flip it back to true before merging.

compwron: For now, what we have is great! Zoe is writing a new github issue for complexities we discovered today. 

Long term, we need to allow let's assume that multiple housing options can be selected at once, *and* that new options which were once selected can be de-selected. For example, a teenage youth who is not getting along with their foster family might spend half the week at a friend's place, and half the week at their foster family's place. And then later no longer be with their foster family, and not able to stay with their friends, and be staying with grandparents instead. So we are not yet sure about how do to the UI for that complicated situation.

compwron (approved):

PR #1454, opened 2020-12-12
Removes puts in view spec

scottolsen:
none

### What changed, and why?
Not sure if anyone is using this but was maybe added for debugging
purposes. It logs either `browser:` or `driver:`. I think we can remove
for now to continue cleaning up the test out.


littleforest (approved):
👍

PR #1455, opened 2020-12-12
Upgrade to Ruby 2.7.2

scottolsen:
none

### What changed, and why?
[Ruby2.7.2](https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/) has been out since October 
so we're probably good to upgrade. Also removes some 
warnings from the test output :)

```
warning: parser/current is loading parser/ruby27, which recognizes
warning: 2.7.2-compliant syntax, but you are running 2.7.1.
warning: please see
https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```


littleforest:
@scottolsen do you want to update the Dockerfile as well? Otherwise I can make a separate ticket and do that.

scottolsen:
@littleforest nice catch!

littleforest (approved):
🎉 💯

Issue #1449, opened 2020-12-12
Speed up test suite


compwron:
Our rspecs (and capybaras) are weirdly slow for how small our codebase is. Try to fix that!

Maybe it's factorybot creating too many objects? 

Or bullet, somehow?

Feel free to tune the CI scripts (CI uses Github Actions) if that will help. (but it's slow locally also, so maybe not)
https://github.com/rubyforgood/casa/actions?query=workflow%3Arspec

Or all of the above, or something else! 😅 
plz send halp

compwron:
@scottolsen fyi here's a github issue for the slow spec :)

scottolsen:
After a quick look around, I suspect this is mainly `FactoryBot` creating unneeded associations and Capybara in the system tests. 19 of the 20 slowest tests are system tests and they all take over 1.5 seconds. 

Also, one of the slower [user model specs](https://github.com/rubyforgood/casa/blob/75021c75b01caa2a0f3ddd610d5412a705c96340/spec/models/user_spec.rb#L35) I checked, creates around 50 database records.
```ruby
:casa_org=>{:create=>12, FactoryBot::Strategy::Create=>5}
:hearing_type=>{FactoryBot::Strategy::Create=>5}
:judge=>{FactoryBot::Strategy::Create=>5}
:casa_case=>{:create=>4, FactoryBot::Strategy::Create=>1} 
:case_assignment=>{:create=>5}
:volunteer=>{:create=>2}
:contact_type_group=>{FactoryBot::Strategy::Create=>4}
:contact_type=>{:create=>4}
:case_contact=>{:create=>4}
:user=>{FactoryBot::Strategy::Create=>1}
```

Hoping to dig in a little deeper this weekend.

littleforest:
@scottolsen what did you run to get this fancy report?

scottolsen:
@littleforest Pulled it from [factory_bot](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#activesupport-instrumentation). I need to tweak it so it only runs when a spec is tagged appropriately.

```ruby
  factory_bot_results = {}
  config.before(:suite) do
    ActiveSupport::Notifications.subscribe("factory_bot.run_factory") do |name, start, finish, id, payload|
      execution_time_in_seconds = finish - start

      if execution_time_in_seconds >= 0.5
        $stderr.puts "Slow factory: #{payload[:name]} using strategy #{payload[:strategy]}"
      end
    end

[...]

PR #1457, opened 2020-12-12
Use already created Org for factories

scottolsen:
#1449

### What changed, and why?
We were creating orgs in many of the factories and I think we can just
use the already created one. We were creating ~2600 orgs per test run
and with this change we are down to ~1300. Probably not a huge
difference in overall time but as things grow, small changes like this
will add up.

I also added some FactoryBot logging in `spec/support/factory_bot.rb`.
It's pretty noisy so it's commented out but will be useful when
debugging slow tests in the future.

### DIFF
--- a/spec/factories/case_assignment.rb
+++ b/spec/factories/case_assignment.rb
@@ -1,7 +1,7 @@
 FactoryBot.define do
   factory :case_assignment do
     transient do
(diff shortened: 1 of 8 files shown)


compwron (approved):
Wonderful!
Yeah @overrides is a remnant

PR #1459, opened 2020-12-12
Don't revert logo if non logo details are updated

### LINKED ISSUE #1434: changing organization details other than logo switches back to default logo
AS an admin,
I WANT my logo to remain the same when I edit non-logo details associated with my `casa_org`,
SO THAT the logo associated with my `casa_org` _only_ changes when _I_ (or another admin) changes it.

### Current behavior
When an admin clicks "submit" on the Edit CASA Org Details page, the logo reverts back to `default logo` (even without changing anything). 

Exception: When an admin clicks "submit" after uploading a new logo, the new logo appears instead of the `default logo`. 

### Expected behavior
The default logo should only appear for `casa_orgs` that haven't uploaded their own logo. 
The logo should only change when an admin changes it. Clicking "submit" without uploading a logo should not affect the current logo.

### Acceptance Criteria:
- [ ] Logo only changes when an admin changes it.

scottolsen:
Resolves #1434

### What changed, and why?
Fixes bug where if you upload a logo for an org and then later
update the org but don't include a logo, it will delete the previously
uploaded logo.

Added regression test to verify the logo is not deleted when the user submits
the org form with no logo.

### DIFF
--- a/app/controllers/casa_orgs_controller.rb
+++ b/app/controllers/casa_orgs_controller.rb
@@ -13,7 +13,6 @@ class CasaOrgsController < ApplicationController
   def update
     respond_to do |format|
       if @casa_org.update(casa_org_update_params)
-        @casa_org.logo.attach(casa_org_update_params[:logo])
         format.html { redirect_to edit_casa_org_path, notice: "CASA organization was successfully updated." }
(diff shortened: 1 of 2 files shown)


compwron (approved):
😅💪💖

PR #1462, opened 2020-12-13
IGNORE: just interested in screen output in CI environment

jhsu802701:
Ignore this pull request.  I'm just running the "cat /etc/os-release" command in the continuous integration environment.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -18,6 +18,8 @@ SimpleCov.start do
   add_group "Database", "/db"
 end
 
+system('cat /etc/os-release')
+
 RSpec.configure do |config|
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -18,6 +18,8 @@
   add_group "Database", "/db"
 end
 
+system('cat /etc/os-release')
```
compwron, line 21: Ubuntu "latest" whatever that github thinks that is today 😅 Set here - https://github.com/rubyforgood/casa/blob/c74097fab2ea5461c421f02b31fb6f5578b39978/.github/workflows/rspec.yml#L22

PR #1464, opened 2020-12-13
Set rack_test as default driver for system tests

scottolsen:
#1449

### What changed, and why?
All the system tests were using the `selenium_chrome_headless` driver.
Now the default is the `rack_test` driver and to use the selenium driver
you just have to tag the test with `js: true`.


littleforest (approved):
This is great, thanks! 🎉

PR #1476, opened 2020-12-15
Fixes Case Contact Report including wrong Org data

### LINKED ISSUE #1340: [multi-tenancy bug] Case Contacts Report has leaky CASA org data

scottolsen:
Resolves #1340

### What changed, and why?
Because `casa_org_id` isn't included in the form params from the view,
it wasn't being permitted in `report_params`. I moved it to be merged
in after we permit the params from the form.

Regression test added to verify `casa_org_id` is actually passed in to
`CaseContactReport`.
I also had to modify another test as it was passing due to the same bug.

### DIFF
--- a/app/controllers/case_contact_reports_controller.rb
+++ b/app/controllers/case_contact_reports_controller.rb
@@ -26,10 +26,9 @@ class CaseContactReportsController < ApplicationController
       :want_driving_reimbursement,
       contact_type_ids: [],
       contact_type_group_ids: [],
-      casa_org_id: current_organization.id,
       creator_ids: [],
       supervisor_ids: []
-    )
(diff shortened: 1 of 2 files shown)


compwron (approved):
YAY!

PR #1489, opened 2020-12-16
Punditize all things

### LINKED ISSUE #1237: Add "after_action :verify_authorized" to force authorization everywhere
all

**Description**
Add "after_action :verify_authorized" to force authorization everywhere
This is a safety measure
https://www.rubydoc.info/github/elabs/pundit/Pundit:verify_authorized
https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used

scottolsen:
Resolves #1237

### What changed, and why?
This uses Pundit in all controllers except `AllCasaAdminsController`. I may
create another issue for adding the `after_action :verify_authorized` to the
`ApplicationController` but I ran in to some issues with Devise and the
`AllCasaAdminsController`.

- [x] Make all policies inherit from `ApplicationPolicy`
- [x] Clean up duplicate methods
- [x] Verify all policies have matching spec for all actions
- [x] Add `authorize` helper to all controller actions
- [x] Add `before_action :authenticate_user!` to `ApplicationController`
- [x] Remove all `before_action :authenticate_user!` from controllers

I'd be more than happy to jump on call with someone to go over things 
since this is a fairly large PR.

This "shouldn't" affect any permissions.

Added policy specs for all controller actions.

### DIFF
--- a/app/controllers/casa_admins_controller.rb
+++ b/app/controllers/casa_admins_controller.rb
@@ -1,16 +1,19 @@
 class CasaAdminsController < ApplicationController
-  before_action :authenticate_user!, :must_be_admin
   before_action :set_admin, except: [:index, :new, :create]
   before_action :require_organization!
+  after_action :verify_authorized
 
   def index
+    authorize CasaAdmin
     @admins = policy_scope(current_organization.casa_admins)
   end
 
   def edit
+    authorize @casa_admin
   end
 
   def update
+    authorize @casa_admin
     if @casa_admin.update(update_casa_admin_params)
       redirect_to casa_admins_path, notice: "Admin was successfully updated."
     else
@@ -19,11 +22,13 @@ class CasaAdminsController < ApplicationController
   end
 
   def new
+    authorize CasaAdmin
     @casa_admin = CasaAdmin.new
   end
 
   def create
     @casa_admin = CasaAdmin.new(create_casa_admin_params)
+    authorize @casa_admin
 
     if @casa_admin.save
       @casa_admin.invite!
@@ -34,6 +39,7 @@ class CasaAdminsController < ApplicationController
   end
 
   def activate
+    authorize @casa_admin
     if @casa_admin.activate
       CasaAdminMailer.account_setup(@casa_admin).deliver
 
--- a/app/controllers/casa_admins_controller.rb
+++ b/app/controllers/casa_admins_controller.rb
@@ -9,11 +9,11 @@ class CasaAdminsController < ApplicationController
   end
 
   def edit
-    authorize @casa_admin, policy_class: CasaAdminPolicy
+    authorize @casa_admin
   end
 
   def update
-    authorize @casa_admin, policy_class: CasaAdminPolicy
+    authorize @casa_admin
     if @casa_admin.update(update_casa_admin_params)
       redirect_to casa_admins_path, notice: "Admin was successfully updated."
     else
@@ -52,7 +52,7 @@ class CasaAdminsController < ApplicationController
   end
 
   def deactivate
-    authorize @casa_admin, :deactivate?
+    authorize @casa_admin
     if @casa_admin.deactivate
       CasaAdminMailer.deactivation(@casa_admin).deliver
 
--- a/app/controllers/casa_admins_controller.rb
+++ b/app/controllers/casa_admins_controller.rb
@@ -9,11 +9,11 @@ class CasaAdminsController < ApplicationController
   end
 
   def edit
-    authorize @casa_admin
+    authorize @casa_admin, policy_class: CasaAdminPolicy
   end
 
   def update
-    authorize @casa_admin
+    authorize @casa_admin, policy_class: CasaAdminPolicy
     if @casa_admin.update(update_casa_admin_params)
       redirect_to casa_admins_path, notice: "Admin was successfully updated."
     else
--- a/app/controllers/case_court_reports_controller.rb
+++ b/app/controllers/case_court_reports_controller.rb
@@ -6,14 +6,14 @@ class CaseCourtReportsController < ApplicationController
 
   # GET /case_court_reports
   def index
-    authorize :case_court_reports
+    authorize CaseCourtReport
     @assigned_cases = CasaCase.actively_assigned_to(current_user)
       .select(:id, :case_number, :transition_aged_youth)
   end
 
   # GET /case_court_reports/:id
   def show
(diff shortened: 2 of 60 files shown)


app/policies/case_contact_policy.rb
```diff
@@ -1,15 +1,14 @@
...
-    # Everyone is allowed to create a case_contact
     true
   end
 
   def _is_creator_or_casa_admin?
-    _is_admin? || _is_creator?
+    is_admin? || _is_creator?
   end
 
   def _is_creator_or_supervisor_or_casa_admin?
```
noahc: Not that you should fix this, but this doesn't look like idiomatic ruby to me. Am I missing something that the `_method_name` implies?

spec/requests/casa_admins_spec.rb
```diff
@@ -27,7 +27,7 @@
       it "cannot access a casa admin edit page" do
         get edit_casa_admin_path(create(:casa_admin))
 
-        expect(response).to redirect_to new_user_session_path
+        expect(response).to redirect_to root_path
```
noahc: This seems like a change in behavior. Does the `new_user_session_path` redirect to the `root_path` and if so if the root path changes is that what we want to always redirect to if you aren't authorized?

app/policies/case_contact_policy.rb
scottolsen: @noahc I'm pretty sure I removed all of those. I think it's a python naming convention to indicate a private method.

spec/requests/casa_admins_spec.rb
scottolsen: @noahc This was reverted back by the final commit.

app/policies/supervisor_volunteer_policy.rb
```diff
@@ -0,0 +1,9 @@
+class SupervisorVolunteerPolicy < Struct.new(:user, :record)
+  def create?
+    user.casa_admin? || user.supervisor? || user.volunteer?
+  end
+
+  def unassign?
+    user.casa_admin? || user.supervisor?
```
noahc, line 7: Isn't there an admin_or_supervisor method you can use here?

app/policies/supervisor_volunteer_policy.rb
```diff
@@ -0,0 +1,9 @@
+class SupervisorVolunteerPolicy < Struct.new(:user, :record)
+  def create?
+    user.casa_admin? || user.supervisor? || user.volunteer?
```
noahc, line 3: This seems to be used in a few places. You could extract to method. I think this represents anyone that is a logged in user, but I don't enough about the domain and context to say for sure.

scottolsen: @noahc I have that on my list after this gets merged it to come back to. Since these headless policies don't inherit from the `ApplicationPolicy` I couldn't reuse those methods here. I plan to extract those next and use them in all the headless policies.

noahc (approved):
Thanks for doing this policy by policy. It made looking through each individual commit much easier. 

🚀

app/policies/supervisor_volunteer_policy.rb
noahc: Since I went commit by commit, I didn't see that it actually already exists. But obviously, you can't use it.

app/policies/case_contact_policy.rb
noahc: I went commit by commit. I saw you did that later on.

app/policies/supervisor_volunteer_policy.rb
noahc: This is headless so you can't.

app/controllers/casa_admins_controller.rb
```diff
@@ -1,16 +1,19 @@
...
   before_action :require_organization!
+  after_action :verify_authorized
 
   def index
+    authorize CasaAdmin
     @admins = policy_scope(current_organization.casa_admins)
   end
 
   def edit
+    authorize @casa_admin, policy_class: CasaAdminPolicy
```
littleforest, line 12: Just curious why you explicitly added the `policy_class` here -- shouldn't Pundit be able to infer it?

app/controllers/case_court_reports_controller.rb
```diff
@@ -1,17 +1,19 @@
 # frozen_string_literal: true
 
 class CaseCourtReportsController < ApplicationController
-  before_action :authenticate_user!
   before_action :set_casa_case, only: %i[show]
+  after_action :verify_authorized
 
   # GET /case_court_reports
   def index
+    authorize :case_court_reports
```
littleforest, line 9: In other index methods use are using `authorize CasaCase` for example. Can we keep these consistent? Or is there a subtle difference between the two?

app/controllers/casa_admins_controller.rb
littleforest: Oh, I see, it is probably using the UserPolicy by default. Never mind!

littleforest (approved):
This is amazing! Thanks so much for cleaning all of this up and for adding the policy specs. Just a minor question.

app/controllers/casa_admins_controller.rb
scottolsen: I thought it had to do with the polymorphic associations of users. Although after poking around a bit a more I see a method in the `User` model that is probably causing this. I'll delete that and see if I can delete all these `policy_class`s. Thanks!

```ruby
  def policy_class
    case type
    when Volunteer
      VolunteerPolicy
    else
      UserPolicy
    end
  end
```

app/controllers/case_court_reports_controller.rb
scottolsen: @littleforest I'll get that changed, nice catch!!

app/controllers/users_controller.rb
```diff
@@ -1,13 +1,15 @@
 class UsersController < ApplicationController
-  before_action :authenticate_user!
+  after_action :verify_authorized
 
   def edit
     @user = current_user
+    authorize @user, policy_class: UserPolicy
```
scottolsen, line 6: Had to specify the policy here since the logged in user could be a `Volunteer` and we still want them to be able to edit themselves.

compwron (approved):
This is a BIG improvement to the state of the art!
It looks good. :) Let's merge this before it gets any bigger. I'm happy for any cleanup we discover to go in a subsequent PR

PR #1490, opened 2020-12-16
made_contact_with_all_cases_in_days? no longer considers inactive and unassigned cases

### LINKED ISSUE #1439: Exclude inactive case and inactive case assignment from contact in days count
function `made_contact_with_all_cases_in_days?`
contains `total_cases_count = casa_cases.size`
called by `datatables/volunteer_datatable.rb` on line 24

**Acceptance criteria:**

- [ ] Do not include inactive cases (+ Write a test!)
- [ ] Do not include cases unassigned from the volunteer (+ Write a test!)

Thanks @FireLemons for finding this! https://rubyforgood.slack.com/archives/CVB0QJGVD/p1607574310223400

FireLemons:
Resolves #1439 

### What changed, and why?
`volunteer.rb` 's `made_contact_with_all_cases_in_days?` no longer considers inactive and unassigned cases

wrote a test where a volunteer has an inactive case checking that the inactive case does not factor into the output of the function
wrote a test where a volunteer has an unassigned case checking that the unassigned case does not factor into the output of the function

### DIFF
--- a/app/models/volunteer.rb
+++ b/app/models/volunteer.rb
@@ -78,10 +78,10 @@ class Volunteer < User
   # false if volunteer has any case with no contact in the past 30 days
   def made_contact_with_all_cases_in_days?(num_days = CONTACT_MADE_IN_DAYS_NUM)
     # TODO this should do the same thing as no_contact_for_two_weeks but for a volunteer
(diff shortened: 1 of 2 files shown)


compwron (approved):
🎉

PR #1493, opened 2020-12-16
Fix around action for rack attack specs

### LINKED ISSUE #1492: Rack Attack Specs in pending state
It looks like the [`freeze_time` change](https://github.com/rubyforgood/casa/pull/1475/commits/736c0fd5aa5105fd854eaa011955b9d2b5c54a11) in #1475 is causing the specs in `spec/config/initializers/rack_attack_spec.rb` to be in a pending state when running locally (not through Docker) and on [Github actions](https://github.com/rubyforgood/casa/pull/1489/checks?check_run_id=1563267566#step:7:22).

```
# around hook at ./spec/config/initializers/rack_attack_spec.rb:16 did not execute the example
```

littleforest:
Resolves #1492

### What changed, and why?
The around action was causing non-Docker specs to be marked as pending. Added proc.

- [x] Make sure GitHub Actions runs specs without marking them as pending

### DIFF
--- a/spec/config/initializers/rack_attack_spec.rb
+++ b/spec/config/initializers/rack_attack_spec.rb
@@ -13,8 +13,8 @@ RSpec.describe Rack::Attack do
     Rails.cache.clear
   end
 
-  around do
-    freeze_time
+  around do |example|
+    freeze_time(&example)
   end
 
   def app


scottolsen (approved):
🥳

PR #1511, opened 2020-12-19
I18n date formats

### LINKED ISSUE #1408: Move all date formats to il8n
Do this https://github.com/rubyforgood/casa/pull/1402#discussion_r531743219

Syntax we want to use: `format: :day_and_date`

**Why?**
- You can name the keys whatever you like using i18n
- One of the benefits of using i18n is that if you internationalize the application, then it's easy to have dates display in formats used by other countries/languages <- even if CASA doesn't need this, it's good for CASA contributors to get used to good ways of doing things. :)

bjthompson805:
Many dates were using a `DateFormat` module and are now using `i18n` instead.

Resolves #1408

### What changed, and why?
Any place that was using the `DateFormat` module to format dates has been changed to use the `I18n.l` function instead.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

I've written a test in `spec/decorators/casa_case_decorator_spec.rb` which checks for the correct date format of a case's updated_at time after it has been updated.

### Screenshots please :)
Here is an example of where a date has been switched from the `DateFormat` module to `i18n` and is working properly to display the date -- the text "Admin was deactivated on: 12-19-2020" shows this.
image

### Feelings gif (optional)
alt text


config/locales/en.yml
```diff
@@ -52,3 +52,8 @@ en:
   time:
     formats:
       day_and_date: "%A, %v"
+      mm_dd_yyyy: "%m-%d-%Y"
```
littleforest, line 55: Can we give this one an actual name, instead of one that describes the date format? While this app will probably not need to be localized, in the spirt of leaving that option open, we should use names that make sense in other languages. For example, if we were to deploy this in Europe, this format would probably look like:

```
de:
  time:
    formats:
      mm_dd_yyyy: "%d-%m-%Y"
```

in which case the label is less useful.  We could use something like "standard" or "default" for the label, or anything else that makes sense to you.

littleforest (changes requested):
@bjthompson805 thanks so much for working on this! Can you also fix the merge conflict?

compwron:
This is looking very promising! There are some test failures. they look like

```
      ActionView::Template::Error:
        uninitialized constant CaseContactDecorator::DateFormat
        Did you mean?  DateAndTime
```
Maybe a merge issue?

bjthompson805:
Yes I've fixed that one (a merged file needed to be migrated to i18n). I am having some other failures after merging as well that I can't figure out. I posted in the Slack casa channel.

littleforest (approved):
🎉 💯 🚢 Thanks for working on this!

Issue #1405, opened 2020-11-27
Fix bullet n+1 issues in tests


compwron:
Remove and fix each test with `:disable_bullet` one by one :)

bjthompson805:
@littleforest Are you working on this or can I take it?

littleforest:
@bjthompson805 you can take it!

bjthompson805:
@compwron I'm really new to Rails and wondering if you could help me figure out where this call is being made that I can add `.includes([:logo_attachment])` as it is suggesting. Here is what 99% of the errors are the first time I ran with `Bullet.raise = true`, any help is appreciated!:

      Bullet::Notification::UnoptimizedQueryError:
        user: root
        GET /casa_cases
        USE eager loading detected
          CasaOrg => [:logo_attachment]
          Add to your query: .includes([:logo_attachment])
        Call stack
          /usr/src/app/app/views/layouts/_sidebar.html.erb:3:in `_app_views_layouts__sidebar_html_erb__493316934298706686_51240'
          /usr/src/app/app/views/layouts/application.html.erb:38:in `_app_views_layouts_application_html_erb__199189227000836964_51180'
          /usr/src/app/spec/system/casa_cases/index_spec.rb:17:in `block (2 levels) in <top (required)>'

littleforest:
@bjthompson805 these can be tricky. In this case it should go on the `in_organization` scope in the users model:

```
  scope :in_organization, lambda { |org|
    where(casa_org_id: org.id).includes(casa_org: :logo_attachment)
  }
```

littleforest:
@bjthompson805 since I commented 3 days ago, shouldn't this not get marked as inactive?

bjthompson805:
Yeah it shouldn't. I'll look into that.

bjthompson805:
@littleforest I fixed the issue, let me know if you see anything else.

littleforest:
Great, thanks @bjthompson805 !

Luke-kb:
Has this issue been resolved?

littleforest:
Nope! There was a very weird issue I found with the logo, which maybe I can pair with you if you are interested in working on this. Since the logo is loaded on pretty much every page, it was affecting nearly every spec.

littleforest:
[...]

PR #1521, opened 2020-12-29
New workflow to unassign inactive issues.

### LINKED ISSUE #1409: Automate github issue weekly "is this still in progress?" checkins
Automate github issue weekly "is this still in progress?" checkins on "in progress" github issues without recent comments / commits. After two weeks (14 days) of checking-in and no comments or PRs, unassign the assignee and move the issue back to "to do" (automate as many parts of this as possible! Anything is better than nothing :) )

This is (hopefully) possible with github actions

bjthompson805:
The workflow will run once per hour and will post a warning message in
issues that are older than 5 days telling the user that the issue will
be unassigned after 7 days. After 7 days it will be unassigned. The
algorithm to determine the time of inactivity will account for comments
made but not events.

Resolves #1409

### What changed, and why?
A new workflow file was created in order to unassign inactive issues.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

I tested it manually in a test harness I created:

https://github.com/bjthompson805/hello-world-javascript-action

### Screenshots please :)
Here is what it looks like when the bot comments as well as unassigns the issue.
image

### Other comments
I suggest a code review of the action itself as well:

https://github.com/bjthompson805/unassign-issues

Also, you may want to fork this action and use it under the rubyforgood repo owner so it's not attached to my name. Let me know and I'll change the workflow file to use it.

### Feelings gif (optional)
How to embed:
alt text

### DIFF
--- a/.github/workflows/issue-auto-unassign.yml
+++ b/.github/workflows/issue-auto-unassign.yml
@@ -1,7 +1,7 @@
 on:
   schedule:
     # * is a special character in YAML so you have to quote this string
-    - cron:  '0 * * * *'
+    - cron:  '0 0 * * *'
 
 jobs:
   unassign_issues:
--- a/.github/workflows/issue-auto-unassign.yml
+++ b/.github/workflows/issue-auto-unassign.yml
@@ -1,0 +1,17 @@
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '0 * * * *'
+
+jobs:
+  unassign_issues:
+    runs-on: ubuntu-latest
+    name: Unassign issues
+    steps:
+      - name: Unassign issues
+        uses: bjthompson805/unassign-issues@v1
+        id: unassign-issues
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
+          unassign_inactive_in_hours: 168 # 7 days
+          warning_inactive_in_hours: 120 # 5 days


compwron:
Very cool! I will dig into this more tomorrow + over the holiday break :) 
If it runs once per hour, that could be a LOT of messages - maybe we should make it once a day or something (or maybe I'm not completely caught up yet... )
This looks super cool! 🙌🎉✨

bjthompson805:
> Very cool! I will dig into this more tomorrow + over the holiday break :)
> If it runs once per hour, that could be a LOT of messages - maybe we should make it once a day or something (or maybe I'm not completely caught up yet... )
> This looks super cool! 🙌🎉✨

Yeah I just realized that's true. I'll probably change the action so it doesn't post a new comment if the last one is from the bot.

bjthompson805:
Fixed the issue of comment spams, and another issue where if there were only bot comments, it would unassign the issue.

compwron (approved):
It looks great! This is so cool :) 

Eventually we will probably want to customize the comment somehow with instructions for what to do i.e. "make a comment to reset the clock" and add some nice language like "thanks so much! Please feel free to pick up more CASA work when you have time!" etc. :)

bjthompson805:
Looking back at my decision to ignore events... I think the action should look at some of them, at least. E.g. #1405 was unassigned even though it was recently self-assigned because there were no recent comments. It should not unassign it if it was recently assigned.

I think that might be the only event I should account for.

It should also comment before it unassigns it, explaining why.

bjthompson805:
I've fixed the issue where it unassigns even after a recently assigned issue. Also added a notification message to explain the unassign (see below).

It also was only checking the first 25 issues. The max to check is 100 so I've set it to that (I don't imagine we would ever need to check more the 100 most recent issues, but if we do we can implement pagination through the data and get the entire array of open issues).

I looked thru the 25 issues and did not notice the bot doing anything else that it shouldn't. Let me know if you notice anything else that is wrong.

New comment that is made during unassignment:
image

PR #1525, opened 2020-12-30
Filter contact type groups and contact types based on selected cases

### LINKED ISSUE #859: Contact Types and Groups display per casa_case when creating/editing a case_contact
Part of epic #856 // dependent on #858

As a volunteer/supervisor/admin, I want to see Contact Types and Groups relevant to the `casa_case` I am logging a `case_contact` for.

**How does it work now?**
All users see all Contact Types when logging a `case_contact` for any `casa_case`.

**How should it work?**
All users should **only** see Contact Types and Groups associated with the `casa_case` they are creating/editing a `case_contact` for – as determined by a Supervisor (see #858)

**Acceptance criteria**
When creating/editing a `case_contact` for a `casa_case`, only Contact Types and Groups selected by a supervisor for that `casa_case` should display.
For `casa_cases` that do not have customized Contact Types and Groups, all Contact Types and Groups available to the associated CASA Org instance (which have not been hidden by an Admin) should display when editing/creating a `case_contact` associated with that `casa_case`.  

So in this example, ONLY the two checked case_contact types should show up for the volunteer
Screen Shot 2020-12-09 at 7 49 20 PM

cliftonmcintosh:
Resolves #859

### Questions
What's the expected behavior of the case has not had contact types defined for it by a supervisor? Does that ever happen or is it never supposed to happen? If it does happen, what contact types should be shown? Show all contact types?

I have assumed that:

1. it is possible that a case will not have been set up by a supervisor so that only some contact types are defined for it
2. when that happens, we show all possible contact types for the current organization

If these assumptions are wrong, please let me know, and I will update this PR.

### What changed, and why?
Filter the contact type groups and the contact types based on the contact types that are assigned to a case.

No effect

I'm open to guidance on where to add tests for this. Please be specific about which spec you think it should be added to. The ones for the view right now assign empty data to the contact type groups that are in the view and, from what I can tell, do not set up contact types for the cases that are in the test. See [here](https://github.com/rubyforgood/casa/blob/main/spec/views/case_contacts/new.html.erb_spec.rb#L7-L12)

### Screenshots please :)
Here are screenshots for two cases. One case (with an id of 840) has no contact types assigned. For that case, all of the organization's contact types are displayed on the form. One case (with an id of 848) has two contact types assigned, both of which are in the "Family" group. For that case, only the assigned types are displayed on the form.

Here is the data for these two cases. 

```sql

casa_development=# select cc.id, cc.case_number, ct.name as "contact type", ctg.name as "group"
casa_development-# from casa_cases cc
casa_development-# left join casa_case_contact_types ccct
casa_development-# on cc.id = ccct.casa_case_id
casa_development-# left join contact_types ct
casa_development-# on ct.id = ccct.contact_type_id
casa_development-# left join contact_type_groups ctg
casa_development-# on ctg.id = ct.contact_type_group_id
casa_development-# where cc.id in (840, 848);
 id  |     case_number     | contact type | group  
-----+---------------------+--------------+--------
 840 | CINA-13-1002 org1MM |              | 
 848 | CINA-09-1010 org1MM | Parent       | Family
 848 | CINA-09-1010 org1MM | Grandparent  | Family
(3 rows)
```

Here are the screenshots

This one is for the case that has no contact types assigned to it, and so all types are displayed. In other words, this behaves like things currently behave.

all-contact-types-when-case-has-no-specific-ones-assigned

This one is for the case that has two contact types assigned to it, and so only those two types and their group are displayed. This is where the change is noticeable.

contact-types-filtered-when-case-has-specific-types

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,7 @@ group :test do
   gem "capybara-screenshot"
   gem "database_cleaner-active_record", "~> 1.8.0"
   gem "rake"
+  gem "rails-controller-testing"
   gem "selenium-webdriver"
   gem "simplecov", "~> 0.17.1", require: false # 0.17.1 pinned as a workaround for https://github.com/codeclimate/test-reporter/issues/418
   gem "webdrivers", require: false # Easy installation and use of web drivers to run system tests with browsers; do not initially require as causes conflict with Docker setup
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -242,6 +242,10 @@ GEM
       bundler (>= 1.15.0)
       railties (= 6.1.0)
       sprockets-rails (>= 2.0.0)
+    rails-controller-testing (1.0.5)
+      actionpack (>= 5.0.1.rc1)
+      actionview (>= 5.0.1.rc1)
+      activesupport (>= 5.0.1.rc1)
     rails-dom-testing (2.0.3)
       activesupport (>= 4.2.0)
       nokogiri (>= 1.6)
@@ -405,6 +409,7 @@ DEPENDENCIES
   pundit
   rack-attack
   rails (~> 6.1.0)
+  rails-controller-testing
   rake
   rspec-rails (~> 4.0.2)
   sablon
--- a/app/controllers/case_contacts_controller.rb
+++ b/app/controllers/case_contacts_controller.rb
@@ -30,7 +30,18 @@ class CaseContactsController < ApplicationController
     # By default the first case is selected
     @selected_cases = @casa_cases[0, 1]
 
-    @current_organization_groups = current_organization.contact_type_groups.joins(:contact_types).where(contact_types: {active: true}).uniq
+    @selected_case_contact_types = @selected_cases.flat_map(&:contact_types)
+
+    @current_organization_groups =
+      if @selected_case_contact_types.present?
+        @selected_case_contact_types.map(&:contact_type_group).uniq
+      else
+        current_organization
+          .contact_type_groups
+          .joins(:contact_types)
+          .where(contact_types: {active: true})
+          .uniq
+      end
   end
 
   def create
--- a/app/views/case_contacts/_form.html.erb
+++ b/app/views/case_contacts/_form.html.erb
@@ -29,7 +29,10 @@
         <% @current_organization_groups.each do |group| %>
           <td style="display:inline-block; vertical-align:top; padding-right:50px; padding-bottom:20px">
             <h5> <%= group.name %> </h5>
-            <% group.contact_types.each_with_index do |contact_type, index| %>
+            <% group.contact_types.filter do |ct|
+                @selected_case_contact_types.blank? ||
+                    @selected_case_contact_types.include?(ct)
+              end.each_with_index do |contact_type, index| %>
               <div class="form-check">
                 <%=
                   check_box_tag "case_contact[case_contact_contact_type_attributes][][contact_type_id]",
--- a/spec/controllers/case_contacts_controller_spec.rb
+++ b/spec/controllers/case_contacts_controller_spec.rb
@@ -1,0 +1,44 @@
+require "rails_helper"
+
+RSpec.describe CaseContactsController, type: :controller do
+  let(:organization) { create(:casa_org) }
+  let(:volunteer) { create(:volunteer, :with_casa_cases, casa_org: organization) }
+  let(:case_id) { volunteer.casa_cases.first.id }
+  let!(:contact_type_group_one) do
+    create(:contact_type_group, casa_org: organization).tap do |group|
+      create(:contact_type, contact_type_group: group, name: "Attorney")
+    end
+  end
+  let!(:contact_type_group_two) do
+    create(:contact_type_group, casa_org: organization).tap do |group|
+      create(:contact_type, contact_type_group: group, name: "Therapist")
+    end
+  end
+
+  before do
+    allow(controller).to receive(:authenticate_user!).and_return(true)
+    allow(controller).to receive(:current_user).and_return(volunteer)
+  end
+
+  describe "GET new" do
+    context "when the case has specific contact types assigned" do
+      before do
+        casa_case = volunteer.casa_cases.first
+        casa_case.contact_types = contact_type_group_one.contact_types
+        casa_case.save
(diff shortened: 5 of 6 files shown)


compwron:
This is looking good! I agree with your assumptions
> it is possible that a case will not have been set up by a supervisor so that only some contact types are defined for it
> when that happens, we show all possible contact types for the current organization

Thoughts on testing:
1. Most important - a new "system" (i.e. feature) spec to test the combination of the controller and view logic - this should go in `spec/system/case_contacts/new_spec.rb` 
1. Nice to have - a new spec file `spec/controllers/case_contacts_controller_spec.rb` to test the logic directly
1. (future PR) We might want to change behavior around **editing** a case contact if type=A was selected, but then type=A was edited to not apply to this case, and the volunteer goes back to edit the case contact - type=A should be visible, but no longer allowed to be elected (maybe) - let's clarify this later. That will be tested in `spec/system/case_contacts/edit_spec.rb` 
1. (optional) a erb spec

Example erb spec 

```
>  tree -f spec | grep erb_spec
    │   └── spec/views/casa_admins/admins_table.html.erb_spec.rb
    │   ├── spec/views/casa_cases/edit.html.erb_spec.rb
    │   ├── spec/views/casa_cases/index.html.erb_spec.rb
    │   ├── spec/views/casa_cases/new.html.erb_spec.rb
    │   └── spec/views/casa_cases/show.html.erb_spec.rb
    │   ├── spec/views/case_contacts/case_contact.html.erb_spec.rb
    │   ├── spec/views/case_contacts/edit.html.erb_spec.rb
    │   ├── spec/views/case_contacts/index.html.erb_spec.rb
    │   └── spec/views/case_contacts/new.html.erb_spec.rb
    │   └── spec/views/case_court_reports/index.html.erb_spec.rb
    │   └── spec/views/dashboard/show.html.erb_spec.rb
    │   ├── spec/views/layouts/application.html.erb_spec.rb
    │   ├── spec/views/layouts/footer.html.erb_spec.rb
    │   └── spec/views/layouts/sidebar.html.erb_spec.rb
    │   ├── spec/views/supervisors/edit.html.erb_spec.rb
    │   ├── spec/views/supervisors/index.html.erb_spec.rb
    │   └── spec/views/supervisors/new.html.erb_spec.rb
        ├── spec/views/volunteers/index.html.erb_spec.rb
        └── spec/views/volunteers/new.html.erb_spec.rb
```

cliftonmcintosh:
@compwron 

Thanks. For me, one of the biggest obstacles to contributing is when there is not already a spec in place for the current functionality. For example, a spec for `spec/system/case_contacts/new_spec.rb` for the testing of existing functionality doesn't already exist. Having to create the full spec before making alterations ups the level of effort for a relatively minor change.

That said, I will work on adding a spec for that.

compwron:
@cliftonmcintosh this project's tests are system-test heavy because those are the tests that are the best at making sure that functionality really really works, altho they are slower than lower-level tests. Our "testing pyramid" https://martinfowler.com/bliki/TestPyramid.html is (intentionally) more like an ice cream cone 😅 

I do really value lower-level "unit" tests! They increase the durability and understandability of the software. I consider the high-level tests more urgent in most cases, to provide the best safety net. 

Never be afraid to add a new testing file! :)

cliftonmcintosh:
@compwron 

Thank you for your guidance. I have updated the system spec in https://github.com/rubyforgood/casa/pull/1525/commits/5d2faaffd4bc352fc453e6c435a02b4e9fc39b8d to include tests that check whether or not specific contact types are displayed and added a controller spec in https://github.com/rubyforgood/casa/pull/1525/commits/feed1a9da2036e82477ce071793fd20925af62f7 to test the logic directly.

Gemfile
```diff
@@ -54,6 +54,7 @@ group :test do
   gem "capybara-screenshot"
   gem "database_cleaner-active_record", "~> 1.8.0"
   gem "rake"
+  gem "rails-controller-testing"
```
compwron, line 57: I'm not super thrilled with adding this extra gem but it's fairly maintained and easy to remove later if it gets obsolete :)
https://github.com/rails/rails-controller-testing
> This gem brings back assigns to your controller tests as well as assert_template to both controller and integration tests.
> Note: This gem is only useful once assigns and assert_template have been removed from Rails.

compwron (approved):
Very cool :)

Gemfile
cliftonmcintosh: @compwron 

Thanks for the review. Perhaps we might consider removing the controller test. Since writing the tests, I've taken a little time to try to learn when and why the ability to test with `assigns` was removed. The removal of the ability to test instance variables was because:

> Testing what instance variables are set by your controller is a bad idea. That's grossly overstepping the boundaries of what the test should know about. You can test what cookies are set, what HTTP code is returned, how the view looks, or what mutations happened to the DB, but testing the innards of the controller is just not a good idea.

See this Rails PR that removed it: https://github.com/rails/rails/issues/18950

If there is an alternative way to test the internal logic of the controller, I'm open to that.

Please let me know if the team thinks the controller tests here should be removed.

compwron: We can change the a little and remove the gem if we want to. Keep the tests for now. :)

littleforest: We are using all request specs now, and no controller tests. It would be great if this spec could get rewritten as a request spec.

cliftonmcintosh: @littleforest and @compwron 

Thank you for the information. 

I have a couple of things to ask about and note:

1. We currently have a system test for this change at https://github.com/rubyforgood/casa/blob/main/spec/system/case_contacts/new_spec.rb#L330-L362. I think this gets us what we want for testing this change (and I'm not sure the controller test a lot of value compared to that). Perhaps these tests cover what we really care about.
2. When looking through the request tests, what I see is verification of interactions in the request. I don't see any tests in the current requests test suite that verify what data are available to the user. Is there an example somewhere of testing the kind of logic that has changed here? Specifically, here that would be making sure that only certain contact types are available as choices for the user. Is there something similar that would be done in a requests test suite?

After the conversation here and looking over the test suite in general, I'm inclined to believe that the system tests cover what we want. If there is a need to add testing, I'm happy to do so. Having a sample of something similar tested in the requests test suite would help with that.

PR #1526, opened 2020-12-31
Add deliver_now to mailer method

littleforest:
### What changed, and why?

Supervisor weekly mailer wasn't going out since it was missing `deliver_now` or `deliver_later`


compwron (approved):
💖

PR #1534, opened 2021-01-01
Add azure-storage to the gemfile

pollygee:
We missed this step in the azure setup.  It wasn't working, and I'm hoping this last step will fix the problem.

Can we merge this @compwron  or @littleforest and then deploy to QA and test uploading a logo again?


littleforest:
@pollygee we installed the `azure-storage-blob` gem because it looked like the `azure-storage` gem was outdated, but now I'm not seeing that. @compwron? The edge guides say to install the `azure-storage-blob`, but the current guides say to use `azure-storage` like you have. But I don't think the issue is with the gem, I think there is some bigger authorization issue going on with Azure.

compwron:
Agreed, I don't expect that the gem is the problem. I'm happy to merge this and put it in QA anyway just to make sure tho!
Looks like we have a date-sensitive test failing your build as a new years present. Since you're the first one on the scene, please fix it 😭💖

compwron (approved):

littleforest:
@pollygee can you remove the `azure-storage-blob` gem out of the Gemfile so we don’t have both listed?

compwron:
Oh whups I thought this was my PR so I edited it... I will stop touching it sorry. Maybe I should sleep or something.

PR #1535, opened 2021-01-02
Collapsible emancipation checklist

### LINKED ISSUE #1494: make emancipation checklist dynamic
As a **volunteer** (or admin/supervisor, but mostly volunteer),
when I am using my Emancipation Checklist,
I want to only see items associated with each header category _after_ I check the box associated with that header category,
so that my checklist is more focused and less cluttered. 

### How it works now
Everything displays all the time, regardless of user input. This makes the list long and overwhelming.

If the box next to the header category is unchecked, only the header displays.
- [ ] **Youth has income to achieve self sufficiency.**

If the box next to the header category is checked, then the items associated with that category also display.
- [x] **Youth has income to achieve self sufficiency.**
   - [ ] Employment
   - [ ] Public Benefits/TCA
   - [ ] SSI 
   - [ ] SSDI
   - [ ] Inheritance/survivors benefits.

This should also apply to categories with _dropdown menus._ The items in the dropdown menu should function just like the checkbox items detailed above. (The "no" option currently in some dropdown menus _can be removed._ Not checking the box next to an item will serve as the equivalent of "no".)

FireLemons:
Resolves #1494

### What changed, and why?
Users can check emancipation categories
Checking emancipation categories expands emancipation options into view
Unchecking emancipation categories removes all emancipation category options from the case
When checkbox actions fail server side, the client side ui won't change.
Some refactoring

Users can now associate emancipation categories with cases

### Screenshots
image
image
image

### DIFF
--- a/app/controllers/emancipations_controller.rb
+++ b/app/controllers/emancipations_controller.rb
@@ -20,16 +20,16 @@ class EmancipationsController < ApplicationController
       return
     end
 
-    unless params.key?("option_action")
-      render json: {error: "Missing param option_action"}
+    unless params.key?("check_item_action")
+      render json: {error: "Missing param check_item_action"}
       return
     end
 
-    if !params.key?("option_id")
-      render json: {error: "Missing param option_id"}
+    if !params.key?("check_item_id")
+      render json: {error: "Missing param check_item_id"}
       return
-    elsif !/\A\d+\z/.match(params[:option_id])
-      render json: {error: "Param option_id must be a positive integer"}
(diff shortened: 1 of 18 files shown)


compwron:
Very cool! The failing test is one that I already fixed with https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time so I will merge this and I expect it to have a green build :)

compwron (approved):
✨🎉

PR #1542, opened 2021-01-03
Trying different config for azure storage

pollygee:
@littleforest @compwron trying to figure out why azure isn't working.  Found this difference in the config file.  Hoping this change will fix it?

### DIFF
--- a/config/storage.yml
+++ b/config/storage.yml
@@ -22,7 +22,7 @@ local:
 #   bucket: your_own_bucket
 
 # Use rails credentials:edit to set the Azure Storage secret (as azure_storage:storage_access_key)
-microsoft:
+azure:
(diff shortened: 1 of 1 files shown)


pollygee:
I verified the env vars in heroku they are correct.  If this isn't the problem I'm not sure what next troubleshooting steps should be.

compwron (commented):
lets try it! :)

littleforest:
This can be named anything we want, but it has to match what we put in the production environment file:

https://github.com/rubyforgood/casa/blob/main/config/environments/production.rb#L43

pollygee:
@littleforest oh ok I see.  I'll change it back.  Sorry about that.  
I don't know why it's not authenticating.

PR #1548, opened 2021-01-04
Heroku review app configuration

### LINKED ISSUE #1238: make per PR heroku review app
make per PR heroku review app
We need permissions for this- ask Sean/Polly
Screen Shot 2020-11-04 at 8 36 02 AM

littleforest:
Resolves #1238 

### What changed, and why?
- Use `db:migrate` instead of `schema:load` since `schema:load` is a destructive action on the database and requires an extra flag to run
- Running the seed file was causing an error since `production` was not available as an option for which dataset to use. Modified the seed code to use the `APP_ENVIRONMENT` environment variable to determine the dataset since we no longer have qa and staging environments. Set the review apps to have `APP_ENVIRONMENT` of "qa".

### DIFF
--- a/app.json
+++ b/app.json
@@ -22,7 +22,7 @@
     }
   },
   "scripts": {
-    "postdeploy": "bundle exec rake db:schema:load db:seed"
+    "postdeploy": "bundle exec rake db:migrate db:seed"
   },
   "stack": "heroku-18"
 }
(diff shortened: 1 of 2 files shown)


littleforest:
@compwron this is ready for review! You can view the review app with the "View deployment" link above.

compwron (approved):
VERY COOL

PR #1549, opened 2021-01-04
Fix help wanted label remover

### LINKED ISSUE #1362: Automatically remove Help Wanted issue label for issues in-progress/qa/later-columns
AS a maintainer of this project,
I WANT the `help wanted` label to automatically be removed from issues in the `in-progress` column or any columns following that one,
SO THAT I do not need to manually remove the `help wanted label` from these issues. 

### Proposed solutions & approach
This is probably a github actions thing

Maybe we could make a cronjob github action that runs every few hours and [finds all issues in project columns](https://docs.github.com/en/free-pro-team@latest/rest/reference/projects#list-project-cards) and then [remove label from issue](https://docs.github.com/en/free-pro-team@latest/rest/reference/issues#remove-a-label-from-an-issue)

bjthompson805:
Resolves #1362

### What changed, and why?
It was not removing the Help Wanted labels so I investigated and found it was using `label_to_add` instead of `label_to_remove`. I also noticed the original request wanted this done for Merged to QA and Done columns as well, so I've added those.

It will be tested when merged to production.

### Screenshots please :)
No screenshots, but I'm confident in this fix going through and working. If it doesn't work, it's only GitHub Actions, anyways :)


compwron (approved):
very nice :)

Issue #1495, opened 2020-12-17
Feedback loop between Volunteers & Supervisors


Zrrrpy:
### Context
This application seeks to replace outdated tactics for managing volunteers currently used by stakeholders with a more unified, efficient tool suite. 

One of those tactics involves the way volunteers and supervisors relay information in regards to `case_contacts`.  Currently, supervisors utilize a variety of communication tools (SMS, email, phone) to tell a volunteer that they need to follow up with their `case` or with a contact associated with their `case`. 

### Approach
The requirements below relate to new features and improvements which will:
- move these communications on-platform
- streamline these communications by reducing back & forth
- provide a more holistic view of outstanding items that require a volunteer or supervisor's attention
- close the loop on completed items

### Requirements
- [x] #1496 all users can flag a `case_contact` that requires followup
- [x] #1497 followups generated by `supervisors` and `admins` are distinguishable from those generated by `volunteers`
- [x] #1506 open followups can be closed by any user
- [x] #1507 `supervisors` and `admins` receive notifications when a `volunteer` closes a followup generated by that `supervisor` or `admin`
- [x] #1508 `volunteers` receive notifications when a followup has been generated for them by a `supervisor` or `admin`

scottolsen:
@Zrrrpy I'm curious if any of the below are valid scenarios?

- A case contact to have multiple followups open at once? Possibly a volunteer and admin both mark it?
- After a followup is resolved, should that case contact every allow another followup?

Zrrrpy:
Thank you for working on this, @scottolsen! 🙏🏼  Apologies for the delayed reply.
Once a followup is added to a case contact, it should not be possible to add another one – even across user types (volunteer/supervisor/admin). 
Yes, new followups should be able to be added to a case contact after an outstanding followup is resolved.

PR #1550, opened 2021-01-05
Case contact followups

### LINKED ISSUE #1496: All users can flag a case_contact that requires followup
Part of epic #1495

As **any user**,
I want the ability to flag a `case_contact` that requires followup,
so that the `volunteer` assigned to that case (even if that is myself) knows that further action is required.

### Acceptance Criteria
- [ ] a "Follow up" alert button displays on all `case_contacts` _logged within this quarter_
- [ ] this button displays to the left of the "Edit button" on desktop and underneath it on mobile
- [ ] clicking the "follow up" button affixes an alert symbol to the `case_contact` (see .png in attachments)
- [ ] once a followup has been added to a `case_contact`, the text on the "follow up" button changes to "resolve"

### Attachments
alert symbol (red would be better, but it should look roughly like this)
alert-logo-png-3-Transparent-Images

Thrillberg:
Resolves #1496 

### What changed, and why?
I added the concept of a `Followup` (and its controller with one action, a `create` action). A Followup has a `status` field that can currently be set to either `"requested"` or `"resolved"`. It's an enum, so should be easy enough to add to, if we want. We're also tracking who added the Followup since I think that will come in handy later. A Followup can currently be applied to a CaseContact.

A button as described in the issue was also added. Let me know if you want me to tweak it style-wise!

No effect on permissions.
- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

I wrote tests which I think cover everything. I've also been QAing it as I go (but would welcome more eyes on it!).

### Screenshots please :)
Screen Shot 2021-01-06 at 8 49 04 PM
Screen Shot 2021-01-06 at 8 50 52 PM

### Feelings gif (optional)

accomplished

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "bootstrap-select": "^1.13.18",
     "datatables.net-dt": "^1.10.20",
     "jquery": "^3.5.0",
+    "node-sass": "^4.14.1",
     "popper.js": "^1.16.1",
     "select2": "^4.0.13",
     "sweetalert2": "^9.10.11",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
     "bootstrap-select": "^1.13.18",
     "datatables.net-dt": "^1.10.20",
     "jquery": "^3.5.0",
-    "node-sass": "^4.14.1",
     "popper.js": "^1.16.1",
     "select2": "^4.0.13",
     "sweetalert2": "^9.10.11",
--- a/app/controllers/case_contact_followups_controller.rb
+++ b/app/controllers/case_contact_followups_controller.rb
@@ -1,0 +1,6 @@
+class CaseContactFollowupsController < ApplicationController
+  def create
+    case_contact = CaseContact.find(params[:contact_id])
+    case_contact.update(followup_created_at: Time.now)
+  end
+end
--- a/app/controllers/case_contact_followups_controller.rb
+++ b/app/controllers/case_contact_followups_controller.rb
@@ -1,6 +1,0 @@
-class CaseContactFollowupsController < ApplicationController
-  def create
-    case_contact = CaseContact.find(params[:contact_id])
-    case_contact.update(followup_created_at: Time.now)
-  end
-end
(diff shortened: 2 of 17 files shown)


littleforest:
I think a `followup_status` enum would be good. Enums are fields that are stored as integers in the database since it takes up less space and has fewer data integrity issues than strings, but you treat them like they are strings or symbols in your code. Plus you get some free methods that make it easy to set the status value. The [documentation](https://api.rubyonrails.org/v5.2.4.4/classes/ActiveRecord/Enum.html) is pretty straightforward.

scottolsen:
Reading through #1495, I'm wondering if a `FollowUp` table would work. If I'm reading it correctly we need not only the follow up status but also who created it. This would also make it easier if creating multiple follow ups for the same case contact becomes a thing. Thoughts?

| follow_ups      |         |
|-----------------|---------|
| id              | bigint  |
| case_contact_id | bigint  |
| status          | integer |
| created_by_id   | bigint  |

I also should have some time tonight after 4 or 5pm central time if you want to pair.

Thrillberg:
Thanks for your help @littleforest and @scottolsen! I've given it another shot here and I think it's a bit closer. @scottolsen I may indeed want to pair tonight but I'm not certain of it yet. I'll post in the Slack channel tonight if I'm up for pairing a bit.

compwron:
don't worry about the web lint build right now, it's having an issue

package.json
```diff
@@ -19,6 +19,7 @@
     "bootstrap-select": "^1.13.18",
     "datatables.net-dt": "^1.10.20",
     "jquery": "^3.5.0",
+    "node-sass": "^4.14.1",
```
compwron, line 22: We have sass now??

compwron:
I like where this is going! :)

package.json
Thrillberg: Not intentionally! Not sure how that got there but I will remove it.

Thrillberg:
Methinks this puppy is ready for review!

compwron (approved):
Very cool!!!

PR #1558, opened 2021-01-06
Fixes case_contact filter in reports

### LINKED ISSUE #1555: contact_type filter broken in report generator
### Impacted user types
Admins and Supervisors

### Environment
Staging and Production

### Current behavior
All `contact_types` within a specified time frame display in a generated report, even when filtering by only one `contact_type`. 
This filter [generated this report.](https://docs.google.com/spreadsheets/d/1HVl5f3WgqM8EGUXQhYJox7KSawqbO0nyarkcctqvUxA/edit?usp=sharing)
grandparents

### Expected behavior
When a supervisor/admin filters by a `contact_type` or `contact_types`, the report generated should only contain `case_contacts` with those `contact_types`.

### How to replicate
1 - Log into staging as a supervisor or admin.
2 - Identify a volunteer account and copy their email.
3 - Log out of staging.
4 - Log back in as that volunteer using the email you copied and the pw: 123456
5 - Log a couple `case_contacts` and take note of which `contact_types` you're selecting – you'll need to remember at least one.
6 - Log out of staging.
7 - Log back in as a supervisor or admin.
8 - Click on "generate reports" in the lefthand sidebar menu.
9 - Select the name of the volunteer you logged `case_contacts` as and a date range that at least includes the contacts you logged. Filter by one of the `contact_types` you logged with. Select "both" for the final three filtering items. 
10 - Generate report.

scottolsen:
Resolves #1555

### What changed, and why?
We were using `contact_type` inside of the `CaseContactReport` class but we
actually pass in `contact_type_ids` from the controller.

I only updated the tests which doesn't really gaurd against this happening in
the future. I'm wondering if we should have 1 system test that uses all these
attributes to generate a CSV and compare it to what we think it should be. I
don't have time right now but would be happy to look in to it.

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/models/case_contact_report.rb
+++ b/app/models/case_contact_report.rb
@@ -27,7 +27,7 @@ class CaseContactReport
       .contact_made(args[:contact_made])
       .has_transitioned(args[:has_transitioned])
       .want_driving_reimbursement(args[:want_driving_reimbursement])
-      .contact_type(args[:contact_type])
+      .contact_type(args[:contact_type_ids])
       .contact_type_groups(args[:contact_type_group_ids])
   end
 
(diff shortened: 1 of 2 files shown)


compwron:
> I'm wondering if we should have 1 system test that uses all these attributes to generate a CSV and compare it to what we think it should be. 

YES plz. Here is a github issue to track it so we don't forget :) https://github.com/rubyforgood/casa/issues/1565

compwron (approved):

PR #1563, opened 2021-01-07
redefined case/emancipation option join table to fix seed bug

FireLemons:
Fixes seed bug from tonight's office hours

Testing
I was able to recreate the error on my end
image
after I made the changes everything ran fine
image

### DIFF
--- a/db/migrate/20210107061732_drop_casa_cases_emancipation_options.rb
+++ b/db/migrate/20210107061732_drop_casa_cases_emancipation_options.rb
@@ -1,0 +1,5 @@
+class DropCasaCasesEmancipationOptions < ActiveRecord::Migration[6.1]
+  def change
+    drop_table :casa_cases_emancipation_options
+  end
+end
--- a/app/models/casa_cases_emancipation_option.rb
+++ b/app/models/casa_cases_emancipation_option.rb
@@ -7,12 +7,16 @@ end
 #
 # Table name: casa_cases_emancipation_options
 #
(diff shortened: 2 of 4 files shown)


db/migrate/20210107061732_drop_casa_cases_emancipation_options.rb
```diff
@@ -0,0 +1,5 @@
+class DropCasaCasesEmancipationOptions < ActiveRecord::Migration[6.1]
+  def change
+    drop_table :casa_cases_emancipation_options
```
compwron, line 3: This is not safe in prod IF any volunteers have edited their emancipation options- we would lose their data. So before we merge this, we have to check for data in prod- and once it's merged, we have to hurry it into prod before any data is created. You can see the count of casa_cases_emancipation_options here: https://data.heroku.com/dataclips/cnluraqwatwiupkkhkpueonwqcfz

I would prefer a safer migration path. @littleforest what do you think?

littleforest: Since it looks like we just want to add on the primary key that was left off, as well as the `created_at`, `updated_at` timestamps, I would just make those changes without destroying the entire table:

Add primary key:
https://stackoverflow.com/questions/5328528/how-do-i-retroactively-add-a-primary-key-to-my-table-in-rails

Add timestamps:
https://stackoverflow.com/a/43548750/1753903

PR #1568, opened 2021-01-08
join tables require unique two-id row combinations

### LINKED ISSUE #1564: Many join tables allow for duplicate records
Namely the tables are: `casa_case_contact_types`, `casa_case_emancipation_categories`, `casa_cases_emancipation_options`, and `case_contact_contact_types`

Steps to recreate:
Open the rails console with `rails console`
run `CasaCase.first.contact_types << ContactType.first` thrice
run `CasaCase.first.contact_types` to see the result which is  

    #<ActiveRecord::Associations::CollectionProxy [
      #<ContactType id: 45, contact_type_group_id: 15, name: "Youth", created_at: "2021-01-07 06:25:30.266025000 +0000", updated_at: "2021-01-07 06:25:30.266025000 +0000", active: true>,
      #<ContactType id: 45, contact_type_group_id: 15, name: "Youth", created_at: "2021-01-07 06:25:30.266025000 +0000", updated_at: "2021-01-07 06:25:30.266025000 +0000", active: true>,
      #<ContactType id: 45, contact_type_group_id: 15, name: "Youth", created_at: "2021-01-07 06:25:30.266025000 +0000", updated_at: "2021-01-07 06:25:30.266025000 +0000", active: true>
    ]>irb(main):008:0>

FireLemons:
Resolves #1564

### What changed, and why?
Added a unique constraint to some join tables because it didn't make sense  
 - to add the same contact type to a casa case twice and vice versa
 - to add the same emancipation option to a casa case twice and vice versa
 - to add the same emancipation category to a casa case twice and vice versa
 - to add the same case contact to a contact type twice and vice versa

### Testing
There are 2 existing constraints that are not tested in `supervisor_volunteer` and `case_assignment` so I did not add any tests. Let me know if tests are preferred.

### DIFF
--- a/app/models/casa_case_contact_type.rb
+++ b/app/models/casa_case_contact_type.rb
@@ -2,6 +2,8 @@ class CasaCaseContactType < ApplicationRecord
   has_paper_trail
   belongs_to :casa_case
   belongs_to :contact_type
+
+  validates :casa_case_id, uniqueness: {scope: :contact_type_id}
 end
 
 # == Schema Information
--- a/app/models/casa_case_emancipation_category.rb
+++ b/app/models/casa_case_emancipation_category.rb
@@ -1,6 +1,8 @@
 class CasaCaseEmancipationCategory < ApplicationRecord
   belongs_to :casa_case
   belongs_to :emancipation_category
+
+  validates :casa_case_id, uniqueness: {scope: :emancipation_category_id}
 end
 
 # == Schema Information
--- a/app/models/casa_cases_emancipation_option.rb
+++ b/app/models/casa_cases_emancipation_option.rb
@@ -1,6 +1,8 @@
 class CasaCasesEmancipationOption < ApplicationRecord
   belongs_to :casa_case
   belongs_to :emancipation_option
+
+  validates :casa_case_id, uniqueness: {scope: :emancipation_option_id}
 end
 
 # == Schema Information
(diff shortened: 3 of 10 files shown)


compwron:
This seems like a very good idea! We need tests to go with it

FireLemons:
@compwron should I add tests for all 6 join models?
  
casa_case_contact_types does not have any tests  
  
some of the model tests don't have shoulda matcher tests I see in tests like `casa_case_spec.rb` should I fill those out?

noahc:
Yes.

littleforest:
@FireLemons I think there may be a few places in the code where we are using `<<` to add new associations. Can you make sure there are specs that exist that check for those cases if an existing association already exists?

FireLemons:
@littleforest I was gonna do `it { should validate_uniqueness_of(:name).scoped_to(:category_id) }` Is that enough?

littleforest:
No, because if you add that constraint, but then we have code like:

```
casa_casa.contact_types << contact_type
```
but that contact_type association already exists for that casa_case, then the code will throw an error. So if there are places in the code that are using `<<` to add the associations then it would be good to add a spec that first sets that association before testing the code.

FireLemons:
how do you convert this to a draft?

littleforest:
Screen Shot 2021-01-08 at 4 29 37 PM

compwron (approved):
Very nice! 💪 tests :)

PR #1569, opened 2021-01-08
Adds system test for CaseContact Report download

### LINKED ISSUE #1565: Add a system test for reports with filters
I'm wondering if we should have 1 system test that uses all these
attributes to generate a CSV and compare it to what we think it should be. 
https://github.com/rubyforgood/casa/pull/1558

scottolsen:
Resolves #1565

### What changed, and why?
Adds system test for downloading the CaseContact Report. I only added one test
for now but it will give us a starting point to add others in the future.
I also changed the `CasaCase` and `ContactTypeGroup` factories to use the first
`casa_org` if it's available, or create a new one if it's not. This led to
needing to change a couple of tests.

### DIFF
--- a/spec/support/capybara.rb
+++ b/spec/support/capybara.rb
@@ -20,11 +20,16 @@ Capybara.register_driver :selenium_chrome_headless_in_container do |app|
 end
 
 Capybara.register_driver :selenium_chrome_headless do |app|
-  Capybara::Selenium::Driver.new app,
-    browser: :chrome,
-    options: Selenium::WebDriver::Chrome::Options.new(
-      args: %w[--headless --disable-gpu --disable-site-isolation-trials --window-size=1280,900]
-    )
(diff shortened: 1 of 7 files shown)


spec/support/capybara.rb
```diff
@@ -20,11 +20,16 @@
 end
 
 Capybara.register_driver :selenium_chrome_headless do |app|
-  Capybara::Selenium::Driver.new app,
-    browser: :chrome,
-    options: Selenium::WebDriver::Chrome::Options.new(
-      args: %w[--headless --disable-gpu --disable-site-isolation-trials --window-size=1280,900]
-    )
+  browser_options = ::Selenium::WebDriver::Chrome::Options.new.tap do |opts|
```
compwron, line 23: nice!

compwron (approved):
beautiful :)

PR #1571, opened 2021-01-09
Better UX for Generate Reports multi selects

scottolsen:
I couldn't find an issue for this.

### What changed, and why?
While I was fixing the report bugs I noticed it was a bit difficult to find and
select what I was looking for. This adds the `select2` class to the 4
multi-selects on this page. We're already using `select2` in some other forms.

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

There is already tests for this selects so adding more seemed like overkill.

### Screenshots please :)
Before:
image

After:
image

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/javascript/src/stylesheets/shared/select2.scss
+++ b/app/javascript/src/stylesheets/shared/select2.scss
@@ -1,0 +1,5 @@
+.select2 {
+  input {
+    min-width: 250px;
+  }
+}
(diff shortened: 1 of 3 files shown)


app/javascript/src/stylesheets/shared/select2.scss
```diff
@@ -0,0 +1,5 @@
+.select2 {
+  input {
+    min-width: 250px;
```
scottolsen, line 3: I had to add this so the placeholder text ("-Select Contact Type Groups-") for contact type groups would fully show. Not a huge fan of it but couldn't find another solution. Open to another idea if anyone has one!

compwron (approved):
Very nice!! :) :)

PR #1574, opened 2021-01-10
mailer issues

compwron:
paired


littleforest:
@compwron I added formatted email (and added fake domain for development if that's okay):

Without formatted email:

Screen Shot 2021-01-09 at 10 21 47 PM

With formatted email:

Screen Shot 2021-01-09 at 10 23 40 PM

pollygee (approved):
Looks great!

Issue #1573, opened 2021-01-10
Admin default view of all supervisors - sort


compwron:
**What type of user is this for? volunteer/supervisor/admin/all**
Admin

**Description**
Sort supervisors by name. Make all columns sortable. Do not add filters (we dont expect to have very many supervisors per org so there's no point)

**Screenshots of current behavior, if any**
![Screen Shot 2021-01-09 at 4 26 15 PM](https://user-images.githubusercontent.com/578159/104111508-6ac88e80-5297-11eb-89ba-0740d1112152.png)

_QA Login Details:_
url: https://casa-qa.herokuapp.com/
password for all users: 123456

Log in using any of the below emails, depending on the account type you'd like to log in as.
- volunteer1@example.com
- supervisor1@example.com
- casa_admin1@example.com

rpolley:
Going to take a look at this.

ATMartin:
Bug discovered during QA: https://github.com/rubyforgood/casa/issues/1573

_That said_, the sorting functionality is working otherwise. Proposing this as "working-in-qa".

PR #1584, opened 2021-01-13
add sorting to supervisors dashboard

rpolley:
Resolves #001

### What changed, and why?
Uses the DataTable function in dashboard.js to make the table used for displaying the supervisors index sortable. See #1573 for motivation.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: No change

Manually tested that sorting is applied to the table, and DataTable itself already has unit tests.

### Screenshots please :)
Capture

### DIFF
--- a/app/javascript/src/dashboard.js
+++ b/app/javascript/src/dashboard.js
@@ -9,6 +9,17 @@ var defineCaseContactsTable = function () {
   )
 }
 
+var defineSupervisorsDataTable = function () {
+  $('table#supervisors').DataTable(
+    {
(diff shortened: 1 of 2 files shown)


compwron (approved):
Yay :) 

I worry that we don't have enough js testing but I'm not sure what good testing would look like here- likely more capybara tests. The goal is to prevent early career contributors from unknowingly breaking things... and also experienced contributors like myself who are upgrading, for example datatables... from breaking things... 😅

PR #1605, opened 2021-01-21
Seed data fixes

littleforest:
### What changed, and why?
* Removed organization prefix from seed data since we fixed all the issues re multitenancy
* Used lorem ipsum text for case contact notes


compwron (approved):
very nice :)

PR #1657, opened 2021-02-01
Improve client side sync for emancipation checklist

FireLemons:
### What changed, and why?
 - Checking an emancipation option will disable the element until the request is complete. (Before it never disabled)
 - An emancipation option will only appear checked after the sent request returns with a successful response. (Before it would appear checked and uncheck itself on error responses)
 - Auto deleting emancipation options is now handled by a single request. (Before multiple requests would be sent quickly)


compwron (approved):
Very nice!
I would like more testing around this functionality and this change but I don't know exactly how to go about it. If I think of something I will post :)

PR #1661, opened 2021-02-02
All casa admin can see casa org overview key metrics #1086

rpolley:
### What github issue is this PR for, if any?
Resolves #1086

### What changed, and why?
Added stats to the casa org dashboard.

### How will this affect user permissions?
- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A
- All Casa Admin permissions: just collects information they already had access to.

### How is this tested? (please write tests!) 💖💪
Cappybara test that generates the page with a mocked organization and then tests that it displays the "correct" statistics for the mocked organization.

### Screenshots please :)
Capture

### Feelings gif (optional)

Capybara doesn't work on WSL

 (Capybara doesn't work on WSL 😢 )

PR #1671, opened 2021-02-03
Follow up for PR #1661

rpolley:
N/A

Follow up changes from code review of #1661

### What changed, and why?
- Added back trailing newline to all_casa_admin org dashboard view
- Removed redundant `cases` method from the `CasaOrg` model, and replaced uses of it with the `casa_cases` has_many relation
- Added `inactive` scope to `User` model and used the `active` and `inactive` scopes instead of  calling `where` directly in the view

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

See tests added in #1661


rpolley:
Any ideas what the deal with the pipeline here is? RSpec is failing here on a test unrelated to what I changed which passed when I ran locally.

compwron:
Hmm that's pretty weird. I will look at it during office hours tonight.

compwron (approved):
Very nice :)

PR #1717, opened 2021-02-14
Reduce repetition in tests

FireLemons:
### What changed, and why?
Some of the tests have repetitive lines so I just wrapped them in `before(:each)` or `let`  
Emancipation checklist request tests have been grouped by user type and is more similar to how casa case request tests are structured.

### DIFF
--- a/spec/controllers/emancipations_controller_spec.rb
+++ b/spec/controllers/emancipations_controller_spec.rb
@@ -12,15 +12,15 @@ RSpec.describe EmancipationsController, type: :controller do
   end
 
   it "raises Missing param casa_case_id error message" do
-    post :save, params:{casa_case_id: 'string'}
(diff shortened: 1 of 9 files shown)


FireLemons:
I think the controller tests were broken by https://github.com/rubyforgood/casa/commit/21850d46c7cc4cee1942c20b9c0b21a083a379d3  
  
I see `expect(response.body).to eq({"error": "Param casa_case_id must be a positive integer"}.to_json)` in the latest `spec/controller/emancipation...` and the commit above removed all the lines matching `Param casa_case_id must be a positive integer` from `app/controllers/emancipations_controller.rb`

compwron (approved):
Yay cleanup! You're right, it was. I was going to clean that up but, uh, I didn't. Thank you thank you thank you! :)

PR #1732, opened 2021-02-16
Supervisor weekly emails account for volunteers being unassinged from cases in part

### LINKED ISSUE #1438: Do not show inactive assignment or inactive case to volunteers. Do not include in supervisor weekly digest.
function `case_assignments_with_cases`
contains `case_assignments.includes(:casa_case)`

called by `views/volunteers/edit.html.erb` on line 80
and `views/supervisor_mailer/weekly_digest.html.erb `on line 15

**Acceptance criteria:**

- [ ] Volunteer Edit should NOT include casa cases from which the volunteer has been unassigned (inactive assignment)
- [ ] Write a test

- [ ] Volunteer Edit should NOT include **inactive** casa cases (inactive case)
- [ ] Write a test

- [ ] Supervisor email should NOT include cases which are inactive at the time that the mailer is generated (inactive case)
- [ ] Write a test

- [ ] Supervisor email should NOT include cases which have been unassigned from volunteers (inactive assignment)
- [ ] Write a test

- [ ] If a volunteer performs actions 
- - and is unassigned from a supervisor in the same week 
- - and **not** assigned to another supervisor
then the supervisor should see the activity in the email
and the email should indicate that the volunteer is unassigned

- [ ] If a volunteer performs actions 
- - and is unassigned from a supervisor in the same week 
- - and **is** assigned to another supervisor, then the volunteer's old supervisor should not see the volunteer's activity in the email 
- - and the volunteer's new supervisor should see the volunteer's activity in their email

Thanks @FireLemons for finding this! https://rubyforgood.slack.com/archives/CVB0QJGVD/p1607574310223400

FireLemons:
Resolves #1438
in part

#### Problem  
A volunteer is unassigned from a casa_case but the supervisor will still see the case in their weekly email every week after they've been unassigned. If the case is transferred to a different volunteer, and it starts receiving case contacts, it will appear as though the volunteer who the case was unassigned from has handled the new case contacts.  
  
#### Changes  
Supervisors will no longer see cases their volunteer have been unassigned from past the week they are unassigned  
Supervisors will see "This case was unassigned from VOLUNTEER on DATE" if the volunteer was unassigned from the case during the week.  
Supervisors will only see the latest case contact on a case that has been unassigned from a volunteer before the unassignment time.  
If a volunteer handled a case contact in the week before they were unassigned, there will be a disclaimer saying that the case activity only describes the portion of the week before the volunteer was unassigned.  

#### Further problems  
This PR does not solve:  
If a case has been reassigned to a new volunteer and they have not had any case contacts but the previous volunteer handled a case contact in the past week, it will appear as though the new volunteer handled the case contact of the previous volunteer.

### DIFF
--- a/app/decorators/casa_case_decorator.rb
+++ b/app/decorators/casa_case_decorator.rb
@@ -29,6 +29,10 @@ class CasaCaseDecorator < Draper::Decorator
     object.case_contacts.max_by(&:occurred_at)
   end
 
+  def case_contacts_latest_before(date)
+    object.case_contacts.where("occurred_at < ?", date).max_by(&:occurred_at)
+  end
+
   def successful_contacts_this_week
     this_week = Date.today - 7.days..Date.today
     object.case_contacts.where(occurred_at: this_week).where(contact_made: true).count
--- a/app/decorators/casa_case_decorator.rb
+++ b/app/decorators/casa_case_decorator.rb
@@ -38,9 +38,19 @@ class CasaCaseDecorator < Draper::Decorator
     object.case_contacts.where(occurred_at: this_week).where(contact_made: true).count
   end
 
+  def successful_contacts_this_week_before(date)
+    this_week_before_date = Date.today - 7.days..date
+    object.case_contacts.where(occurred_at: this_week_before_date).where(contact_made: true).count
+  end
+
   def unsuccessful_contacts_this_week
     this_week = Date.today - 7.days..Date.today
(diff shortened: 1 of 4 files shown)


FireLemons:
couldn't think of a way to shorten those lines for the linter

compwron (approved):
Very cool!

We will figure out how to fix the erb lint after this deploy :)

PR #1763, opened 2021-02-20
Imported data updates existing records

### LINKED ISSUE #1760: admins can bulk update data by re-importing it
AS an admin,
I WANT to be able to re-import data that I've _already_ imported containing new or corrected details,
SO THAT this data can be bulk updated. 

### Environments
QA, Staging, Prod

### How it works now
The import tool only accepts _new_ data. If an admin attempts to re-import existing system data, they get the error: _Validation failed: [`Case number` or `Email`] has already been taken._
[sample imports - Sheet2.pdf](https://github.com/rubyforgood/casa/files/6004999/sample.imports.-.Sheet2.pdf)
[sample imports - Sheet4.pdf](https://github.com/rubyforgood/casa/files/6005003/sample.imports.-.Sheet4.pdf)
[sample imports - Sheet6.pdf](https://github.com/rubyforgood/casa/files/6005008/sample.imports.-.Sheet6.pdf)

Emails and case numbers that already exist should be accepted by the import tool. If the data is re-imported with no changes, then no changes should be made to it. If the data is re-imported with new details (ex: same case number, different `birth_month_year_youth`), then the new details are updated.

FireLemons:
Resolves #1760

### What changed, and why?
Imported data attempts to update if primary keys of record found to match existing records

I ran into a strange error where `expect`ing the importer `to` `change` fields of records didn't work but I tested it all by console and it works fine.

#### Updating Volunteer Name
Updating volunteer name
#### Updating Supervisor Volunteer Assignment
Updating Supervisor Volunteer assignment
#### Updating Case `birth_month_year_youth`
Updating Case birth_month_year_youth

### DIFF
--- a/app/lib/importers/case_importer.rb
+++ b/app/lib/importers/case_importer.rb
@@ -12,7 +12,7 @@ class CaseImporter < FileImporter
   def import_cases
     failures = []
 
-    import_results = import { |row|
+    import { |row|
       casa_case_params = row.to_hash.slice(:case_number, :transition_aged_youth, :birth_month_year_youth).compact
 
       if !casa_case_params.key?(:case_number)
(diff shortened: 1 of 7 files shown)


compwron (approved):
✨💪🙌🏼💖

PR #1774, opened 2021-02-22
[#1759] Remove transition_aged_youth column from CSV file when importing a case

### LINKED ISSUE #1759: remove transition_aged_youth column from case import .csv
As an **admin**,
I WANT the `transition_aged_youth` column to be removed from the sample case import .csv,
SO THAT I don't accidentally do extra work that is now automated.

### Environments
QA, staging, prod

### How it works now
When an admin downloads a sample case .csv for imports, there is a column for `transiton_aged_youth`. This prompts admins to go through the whole spread sheet and mark `true` or `false` per case depending on their age. 
[Case sample import - Sheet1.pdf](https://github.com/rubyforgood/casa/files/6004863/Case.sample.import.-.Sheet1.pdf)

The `transition_aged_youth` column in this .csv should be removed since it is no longer needed. The platform now automatically calculates whether or not a youth is transition aged based off their birth month and year.

rhian-cs:
Resolves #1759

### What changed, and why?
When an admin imports Cases from a CSV file, they no longer have to (or can) specify the `transition_aged_youth` column.

When importing a case from a CSV file, `transition_aged_youth` is now calculated by checking if the youth is 14 years old or older.

The reference `casa_cases.csv` from the imports page has been updated and the column has been removed.

- Admin permissions: They can no longer specify a value for `transition_aged_youth` when importing a case from a CSV file, since it is now calculated automatically.

Since this is a minor change, I only updated the existing tests and fixtures. 

In the `casa_cases.csv` fixture I changed the `birth_month_year_youth` of one of the cases from 2014 to 2000, so we could test both values of `transition_aged_youth`.

### Screenshots please :)
Nothing in the UI was changed, but here is a screenshot of the sample CSV file before and after the change.

#### Before
before

#### After
after

### DIFF
--- a/spec/lib/importers/case_importer_spec.rb
+++ b/spec/lib/importers/case_importer_spec.rb
@@ -18,13 +18,13 @@ RSpec.describe CaseImporter do
       expect { case_importer.import_cases }.to change(CasaCase, :count).by(3)
 
       # correctly imports true/false transition_aged_youth
-      expect(CasaCase.find_by(case_number: "CINA-01-4347").transition_aged_youth).to be_truthy
-      expect(CasaCase.find_by(case_number: "CINA-01-4348").transition_aged_youth).to be_falsey
+      expect(CasaCase.find_by(case_number: "CINA-01-4347").transition_aged_youth).to be_falsey # birth_month_year_youth is nil
+      expect(CasaCase.find_by(case_number: "CINA-01-4348").transition_aged_youth).to be_truthy
       expect(CasaCase.find_by(case_number: "CINA-01-4349").transition_aged_youth).to be_falsey
 
       # correctly imports birth_month_year_youth
       expect(CasaCase.find_by(case_number: "CINA-01-4347").birth_month_year_youth).to be_nil
-      expect(CasaCase.find_by(case_number: "CINA-01-4348").birth_month_year_youth&.strftime("%Y-%m-%d")).to eql "2014-02-01"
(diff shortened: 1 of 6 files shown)


spec/lib/importers/case_importer_spec.rb
```diff
@@ -18,13 +18,13 @@
       expect { case_importer.import_cases }.to change(CasaCase, :count).by(3)
 
       # correctly imports true/false transition_aged_youth
-      expect(CasaCase.find_by(case_number: "CINA-01-4347").transition_aged_youth).to be_truthy
-      expect(CasaCase.find_by(case_number: "CINA-01-4348").transition_aged_youth).to be_falsey
+      expect(CasaCase.find_by(case_number: "CINA-01-4347").transition_aged_youth).to be_falsey # birth_month_year_youth is nil
```
abnersajr, line 21: If you don't expect to have this field anymore, you could remove this expect or change it to `be_nil` that corresponds better to the expectation.

abnersajr (changes requested):

compwron (approved):
Very cool! :) ✨🙌

spec/lib/importers/case_importer_spec.rb
compwron: I generally prefer that we follow https://www.betterspecs.org/ but as long as the linter likes it, I'm ok with it.

PR #1811, opened 2021-03-05
Allow supervisors to edit a volunteer's email address

### LINKED ISSUE #1805: supervisor can edit volunteer's email
AS a **supervisor**,
I WANT to be able to edit a volunteer's email,
SO THAT the volunteer knows to come to _me_ when their email changes, and I can then make this change in other systems we rely on where the volunteer's email would also need to be updated.

### How it works now
Only admins can edit a volunteer's email.

Supervisors _and_ admins should both be able to edit a volunteer's email.

isaacm:
Resolves #1805 

- Volunteer permissions:
- Supervisor permissions: can now edit a volunteer's email address
- Admin permissions:

See unit tests included in this pull request

### Screenshots please :)
Screen Shot 2021-03-04 at 9 04 11 PM

### DIFF
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -4,7 +4,7 @@ class UserPolicy < ApplicationPolicy
   end
 
   def update_volunteer_email?
-    is_admin?
+    admin_or_supervisor?
   end
 
   def unassign_case?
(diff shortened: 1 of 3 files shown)


compwron (approved):
Very nice :) Yay tests!
Lint is failing, but I'll lint after merging.

Issue #1259, opened 2020-11-05
Move js out of erb


compwron:
move js out of app/views/case_contacts/_case_contact.html.erb and into a js file

```
<script type="text/javascript">

  function moreText (event) {
    event.preventDefault()
    const table_cell = event.target.closest('td')

    table_cell.querySelectorAll('.full-notes')[0].style.display = 'block'
    table_cell.querySelectorAll('.limited-notes')[0].style.display = 'none'
  }

  function lessText (event) {
    event.preventDefault()
    const table_cell = event.target.closest('td')

    table_cell.querySelectorAll('.full-notes')[0].style.display = 'none'
    table_cell.querySelectorAll('.limited-notes')[0].style.display = 'block'
  }
</script>

```

kukyak2020:
hi team! I would like to work on this issue. Can you add to me this issue? @littleforest? thanks :)

kukyak2020:
Thanks 😎

ATMartin:
@kukyak2020 Are you still working on this ticket? I'd love to take it on if not.

kukyak2020:
@ATMartin Hi, yes I am working on it. Can you give me a couple of days for it? thank you! 🙏🏿

ATMartin:
@kukyak2020 Of course! It looks like a GitHub bot unassigned you - @compwron could we reassign so no one else grabs this ticket?

Zrrrpy:
@kukyak2020 just reassigned. Thank you for your help with this! 🙏🏼

PR #1816, opened 2021-03-08
Move read more JS code from partial

arku:
Resolves https://github.com/rubyforgood/casa/issues/1259

### What changed, and why?

Moves read more JS code from a partial which was being rendered N times. Uses vanilla JS event delegation to handle read more and read less link clicks. We can reuse this feature in other views by using the`.js-*` classes the script expects.
NA

### DIFF
--- a/app/javascript/src/readMore.js
+++ b/app/javascript/src/readMore.js
@@ -1,0 +1,27 @@
+document.addEventListener('DOMContentLoaded', () => {
+  document.addEventListener('click', (event) => {
+    if (event.target.matches('.js-read-more')) {
+      return handleReadMore(event)
+    }
+
+    if (event.target.matches('.js-read-less')) {
+      return handleReadLess(event)
+    }
+  })
+})
+
+const handleReadMore = (event) => {
+  event.preventDefault()
+
+  const wrapper = event.target.closest('.js-read-more-text-wrapper')
+  wrapper.querySelector('.js-full-text').style.display = 'block'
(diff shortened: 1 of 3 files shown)


app/views/case_contacts/_case_contact.html.erb
```diff
@@ -13,15 +13,15 @@
               <br>
               <%= contact.decorate.subheading %>
             </h6>
-            <div class="card-text">
+            <div class="card-text js-read-more-text-wrapper">
```
arku, line 16: Couldn't come up with a better name for this one. Let me know if you have any suggestions.

app/javascript/src/readMore.js
```diff
@@ -0,0 +1,27 @@
...
+  wrapper.querySelector('.js-full-text').style.display = 'block'
+  wrapper.querySelector('.js-truncated-text').style.display = 'none'
+}
+
+const handleReadLess = (event) => {
+  event.preventDefault()
+
+  const wrapper = event.target.closest('.js-read-more-text-wrapper')
+  wrapper.querySelector('.js-truncated-text').style.display = 'block'
+  wrapper.querySelector('.js-full-text').style.display = 'none'
```
arku, line 26: Looks like CodeClimate doesn't like the duplication here. Not sure if we should worry about it.

compwron (approved):
Yay!

compwron:
We should fix codeclimate but I'm happy to merge this now :)

PR #1817, opened 2021-03-08
Changes is_active column name

### LINKED ISSUE #1776: migrate column case_assignment.is_active to case_assignment.active to match other columns
migrate column case_assignment.is_active to case_assignment.active to match other columns

this is too annoying, kill it. 
and remember to do a SAFE column rename! :) 

we have a workaround in case-assignment scope, 
and in case assignment factory

efgalvao:
Resolves #1776

Changes the column case_assignment.is_active  to case_assignment.active  to match the others columns

This change won't affect user permissions

All tests which were working with .is_active now work as .active

### DIFF
--- a/app/controllers/case_assignments_controller.rb
+++ b/app/controllers/case_assignments_controller.rb
@@ -6,13 +6,13 @@ class CaseAssignmentsController < ApplicationController
     authorize CaseAssignment
     case_assignments = case_assignment_parent.case_assignments
     existing_case_assignment = if params[:volunteer_id]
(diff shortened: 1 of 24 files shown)


cam-benfield (approved):
Looks good, PR #1820 will need to be reworked in order to merge d/t renaming of columns but good work!

compwron (approved):
Very cool :) 
I am a little worried about the deploy but we'll be extra careful. We don't need zero-downtime deploys so that makes our lives easier. https://www.sandipmane.dev/zero-downtime-migrations-in-rails

compwron:
Thanks! :)

PR #1840, opened 2021-03-16
Pretty print FactoryBot object creation stats

arku:
### What changed, and why?

This is a super minor PR pretty-printing the stats from FactoryBot.

**Before**
image

**After**

image

NA


compwron (approved):
:)

PR #1843, opened 2021-03-16
Fix factory bot lint workflow and job names

arku:
### What changed, and why?

This is a minor PR fixing the factory bot lint workflow which uses the same workflow name and job name as the `rspec` workflow.

**Before**

image

**After**

image

NA


compwron (approved):
Thanks! :)

PR #1847, opened 2021-03-17
Fix error handling when importing casa cases, supervisors and volunteers

fabioxgn:
I noticed that there was not test coverage for missing case number and missing e-mail validations when importing cases, supervisors and volunteers. When I wrote the tests I figured out that the error handling for this validations were wrong. It was adding the error to an array that wasn't used and the process returned as success.

I've changed the method to raise errors, because when you raise an error for a row it is automatically handled by the `FileImporter`

### Screenshots please :)

Message after this change:

image

### DIFF
--- a/app/lib/importers/case_importer.rb
+++ b/app/lib/importers/case_importer.rb
@@ -10,35 +10,26 @@ class CaseImporter < FileImporter
   end
 
   def import_cases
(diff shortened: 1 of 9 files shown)


compwron (approved):
🎉💪✨

PR #1857, opened 2021-03-23
WIP - Fix for the mailer having the image and organization details hard cod…

seanmarcia:
…ed to PG Casa

This isn't ready to go because I couldn't get the letter opener to work locally. I also wasn't sure of the best approach here so I made a start if someone wants to take what I've done and use whatever is useful and discard the rest. 

I've:
- I updated the mailers to all have a @casa_organization -- I was going to set something in the application mailer like `@casa_org = current_user.casa_org` but wasn't sure of the logic in all the emails if that made sense. Plus one of the mailers was already using @casa_organization
- Added in a default logo for casa_orgs for orgs that don't immediately upload one.
- I attached the image to the email but I suspect we're going to need to add in a domain too?

Thanks to @pollygee for testing -- it looks like it works!

Org with no logo:
Weekly_summary_of_volunteers__activities_for_the_week_of_2021-03-16

Org with a logo:
Weekly_summary_of_volunteers__activities_for_the_week_of_2021-03-16_and_send_supervisor_digest_rake_—_casa

Obviously ignore the ugly logo used :)


compwron (approved):
This looks wonderful. Merge at will. :)

PR #1870, opened 2021-03-26
Enable bullet n+1 checking for all tests by default

compwron:
Next we need to fix all our bullet problems and remove each `:disable_bullet` one by one by one... 
https://github.com/flyerhzm/bullet


compwron:
It works on my machine 😭
Screen Shot 2021-04-02 at 7 44 03 PM

PR #1892, opened 2021-04-02
tiny refactor to decorator

compwron:


compwron:
I think that github actions is confused and this test will pass on master... I'll clean it up if not

PR #1926, opened 2021-04-07
Fix broken link in doc/CONTRIBUTING.md

dmarcoux:
Not much, but I've noticed this while reading the document.


doc/CONTRIBUTING.md
```diff
@@ -52,4 +52,4 @@ If you are wondering how to keep your fork in sync with the main [repo], follow
 [code of conduct]: https://github.com/rubyforgood/code-of-conduct
 [issues]: https://github.com/rubyforgood/casa/issues?q=is%3Aopen+is%3Aissue+label%3A%22Status%3A+Available%22
```
dmarcoux, line 53: This link seems also off since there's only one issue with the label `Status: Available`, but I don't know the project well enough to know if it is really wrong. If it is, the labels should be changed or perhaps linking to the issues is good enough.

compwron: Thanks for the fix! I think we use `Help Wanted` nowadays

compwron (approved):

PR #1928, opened 2021-04-07
Move erb_lint to development and test groups in Gemfile

dmarcoux:
Installing erb_lint is now unneeded in the erb_lint GitHub workflow since it's now part of the bundler cache, like all gems in the test group.

The erb_lint GitHub workflow should still pass.


compwron (approved):

compwron:
nice, thanks :)

PR #1936, opened 2021-04-08
show error on failed case cort report generation when template is not found

### LINKED ISSUE #1917: Show error on failed case court report generation
**What type of user is this for?**
volunteer

**Description**
The generate report button just hangs if the operation fails. It would be nice to show an error on failure instead
**Current behavior**
image
image

hairedfox:
… found

Resolves #1917 

### What changed, and why?
- show error when the template is not found

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Screenshots please :)
image

### Feelings gif (optional)
How to embed:
alt text

### DIFF
--- a/app/controllers/case_court_reports_controller.rb
+++ b/app/controllers/case_court_reports_controller.rb
@@ -38,14 +38,15 @@ class CaseCourtReportsController < ApplicationController
 
           render json: {link: case_court_report_path(casa_case.case_number, format: "docx"), status: :ok}
         else
-          flash[:alert] = "Report #{params[:case_number]} is not found."
-          error_messages = render_to_string partial: "layouts/flash_messages", formats: :html, layout: false, locals: flash
-          flash.discard
+          error_messages = generate_error(t(".error.report_not_found", case_number: params[:case_number]))
 
           render json: {link: "", status: :not_found, error_messages: error_messages}, status: :not_found
         end
       end
     end
+  rescue Zip::Error
+    error_messages = generate_error(t(".error.template_not_found"))
+    render json: {status: :not_found, error_messages: error_messages}, status: :not_found
   end
 
   private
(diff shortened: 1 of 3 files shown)


hairedfox:
@compwron

I'm not sure how to reproduce this error yet. So if you help me with that, I'd be so grateful. In the meantime, I just change the `path_to_template` to somewhere nonexistent, then the same error would come out.

 Actually, I don't have much time to investigate this to make it cleaner. But I'd love to hear from you first.

Btw, about the javascript part, if it goes for a long run. I think we shouldn't put all the javascript in the view as the case in `app/views/case_court_reports/index.html.erb`. I suggest using some other js framework to handle that, and `Stimulus` might be a good choice.

Hope to hear from you soon!!!

FireLemons:
The problem #1917 describes is that after generating a court report fails for any exception, the UI still says "Court Report Generating do not refresh or leave the page" after it fails. So the user doesn't have any ui telling them to stop waiting because the court report isn't coming. The court report generation might fail for more reasons than not being able to find the template. 

I would recommend changing `rescue Zip::Error` to rescue all errors or you could not rescue from failed generations and the generate report request returns a response with status code 500 that could be handled with javascript.

hairedfox:
> The problem #1917 describes is that after generating a court report fails for any exception, the UI still says "Court Report Generating do not refresh or leave the page" after it fails. So the user doesn't have any ui telling them to stop waiting because the court report isn't coming. The court report generation might fail for more reasons than not being able to find the template.
> 
> I would recommend changing `rescue Zip::Error` to rescue all errors
> or
> you could not rescue from failed generations and the generate report request returns a response with status code 500 that could be handled with javascript.

Yeah if that's the case then I'll think about capture the StandardError. I pushed this PR just to get more information first. Thank you for pointing it out.

compwron (approved):
Good progress! :)

PR #1938, opened 2021-04-10
#1392 Fix: Disable duplicated contact type

hiendinhngoc:
### What github issue is this PR for, if any?
Resolves #1392 

### What changed, and why?
- Add uniqueness validation with scope to contact type to avoid creating the duplicated record.

### How will this affect user permissions?
- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Screenshots please :)

Screenshot from 2021-04-11 00-47-01

PR #1939, opened 2021-04-12
Remove unneeded mimemagic dependency from Gemfile.lock

dmarcoux:
The changes were generated after running `bundle install` on the current `main` branch. This is reverting changes from #1938.

Mimemagic isn't needed since it was dropped from Rails' dependencies in Rails 6.1.3.1. Active Storage (and thus Rails) is now relying on Marcel for MIME type data. See upstream: https://weblog.rubyonrails.org/2021/3/26/marcel-upgrade-releases/


compwron (approved):
Thanks! :)

Issue #1257, opened 2020-11-05
Move more text to en.yml


compwron:
Move more text to en.yml
Someday we want to be able to translate

ErinClaudio:
Hello,
I would love to tackle this one, but I could use a little guidance.
I am currently reviewing the en.yml file.

Zrrrpy:
@compwron @littleforest any pointers for @ErinClaudio?

compwron:
@ErinClaudio there are strings in ERB files which are english text. We want to move these (one or a few at a time, not all at once) to en.yml so that we can have them in a central location and if we ever do translations to spanish, french etc then this is a pre-step for that also. 
So an example is - in file https://github.com/rubyforgood/casa/blob/main/app/views/judges/new.html.erb the string "New Judge" would move to be in `en.yml` under a key like `en.page.judge.title` (for example, title "CASA Volunteer Tracking" on line 50 is at key `en.page.meta.title`

compwron:
We're too small to worry about most of this stuff but https://phrase.com/blog/posts/rails-i18n-best-practices/ is something to read about the principles behind all this :)

hairedfox:
@compwron I'll take a look at this issue :D

PR #1941, opened 2021-04-12
Use localization for buttons/links

dmarcoux:
Related to #1257

See the commit descriptions for details. Each description has the command which was used to do this with GNU sed.


compwron (approved):
nice sed :)

PR #1956, opened 2021-04-16
default court report contains court mandates

### LINKED ISSUE #1948: Default court template supports court mandates
copy the mandates table from the montgomery template to the default template and preserve styling

FireLemons:
Resolves #1948

### What changed, and why?
I added the court mandates table from the montgomery template to the default template. 
I figure if the table is unwanted it would be very easy to delete

image


compwron (approved):
very cool!
I want to find a way to test the results of the doc create more someday...

PR #1957, opened 2021-04-16
Corrects Date written field

### LINKED ISSUE #1947: Created date is not showing up in court reports
### Impacted User Types
volunteers

Date written should not be blank
image

You may notice other fields are blank after generating the report. That's not because they're also broken. It's because the test data seeder leaves some fields blank. 
image

### How to Replicate
1. - log in to [QA](https://casa-qa.herokuapp.com/) or locally with `volunteer1@example.com`, `123456`
2. - Click on "Generate Reports" in the left sidebar menu.
3. - Generate a report

efgalvao:
Resolves #1947 Issue

### What changed, and why?
Added the correct command to create the Date. Because with the old command the field was looking in the wrong place for the "full" date format and the field become nil because the default was set to nil.

- Volunteer permissions: Nothing
- Supervisor permissions: Nothing
- Admin permissions: Nothing

I have created a simple test to fail if created_date is nil.

### Screenshots please :)
Screenshot from 2021-04-16 12-43-13

### Feelings gif (optional)
alt text

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -88,3 +88,6 @@ en:
     month: Month
     year: Year
     none: None
+  date:
+    formats:
+      full: "%B %-d, %Y"
--- a/spec/models/case_court_report_spec.rb
+++ b/spec/models/case_court_report_spec.rb
@@ -64,6 +64,10 @@ RSpec.describe CaseCourtReport, type: :model do
       it "must have Case Contacts as type Array" do
         expect(subject[:case_contacts]).to be_instance_of Array
       end
+
+      it "created_date is not nil" do
+        expect(subject[:created_date]).to_not be(nil)
+      end
     end
 
     describe "when generating report" do


config/locales/en.yml
```diff
@@ -88,3 +88,6 @@ en:
     month: Month
     year: Year
     none: None
+  date:
+    formats:
+      full: "%B %-d, %Y"
```
compwron, line 93: we already have date formats, including one named full. Is it reasonable in the code to reuse those, or move this to the section where they are?

compwron (approved):

config/locales/en.yml
efgalvao: I have seen only the time format named full. 
When is called 
`created_date: I18n.l(Date.today, format: :full, default: nil)`
He looks for that one I have inserted.

compwron:
:)

PR #1961, opened 2021-04-18
Fix: Load wrong casa cases when editing volunteer

### LINKED ISSUE #1959: Cases from wrong org shown on Edit Volunteer case assignment
### Impacted User Types
admins and supervisors

### Environment
all

### Current Behavior
shows cases not allowed

### Expected Behavior
show only cases for org

### How to Replicate
Screen Shot 2021-04-16 at 5 37 04 PM (1)

hiendinhngoc:
Resolves #1959 

### What changed, and why?
- Update `actively_assigned_excluding_volunteer` scope, add where condition to load only casa cases which have the same casa organization with the volunteer.

- Supervisor permissions:
- Admin permissions:


app/models/casa_case.rb
```diff
@@ -53,6 +53,7 @@ class CasaCase < ApplicationRecord
     joins(:case_assignments)
       .where(case_assignments: {active: true})
       .where.not(case_assignments: {volunteer: volunteer})
+      .where(casa_org: volunteer.casa_org)
```
compwron, line 56: This is great! Please remove the other changes (javascript and package.json) and add a test for this so we don't break it again :)

compwron (changes requested):

app/models/casa_case.rb
hiendinhngoc: sorry I missed the message, although the code was merged, I will add the test when I have time, thanks for reviewing.

PR #2004, opened 2021-05-03
Hide/Show unassigned volunteers on Supervisor Edit page

### LINKED ISSUE #1369: Hide unassigned volunteers from View Supervisor by default
sueprvisor/admin

**Description**
Supervisor/Admin viewing Supervisor should by default not see unassigned volunteers but be able to show them

librod89:
Resolves #1369 

### What changed, and why?

- Hide previously unassigned volunteers from supervisor edit page by default
- Added a new button to the supervisor edit page to show/hide previously unassigned volunteers

It will not affect user permissions.

Added controller tests and system tests

### Screenshots please :)

**New button to "Include unassigned" only if there are previously unassigned volunteers that are currently hiding**
image

**Button text changes when you click it**
image

**No button is shown if there are no previously unassigned volunteers for this supervisor**
image

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/controllers/supervisors_controller.rb
+++ b/app/controllers/supervisors_controller.rb
@@ -3,7 +3,8 @@
 class SupervisorsController < ApplicationController
   before_action :available_volunteers, only: [:edit, :update]
   before_action :set_supervisor, only: [:edit, :update, :activate, :deactivate]
-  before_action :all_volunteers_ever_assigned, only: [:edit, :update]
+  before_action :all_volunteers_ever_assigned, only: [:update]
+  before_action :supervisor_has_unassigned_volunteers, only: [:edit]
   after_action :verify_authorized
 
   def index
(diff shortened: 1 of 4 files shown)


librod89:
The downside to this solution is that if the current user has unsubmitted changes on the page, clicking this button will lose those changes. Do you think a Javascript only solution would be better @compwron?

compwron:
> The downside to this solution is that if the current user has unsubmitted changes on the page, clicking this button will lose those changes. Do you think a Javascript only solution would be better @compwron?

What unsubmitted changes could there be on this page? I can't think of any offhand but you're the one in the code right now-  let me know what I'm missing

In general CASA would like to keep our javascript to a minimum to keep our maintenance burden low, and especially because we don't have enough js tests yet.

librod89:
> What unsubmitted changes could there be on this page? I can't think of any offhand but you're the one in the code right now- let me know what I'm missing
> 
> In general CASA would like to keep our javascript to a minimum to keep our maintenance burden low, and especially because we don't have enough js tests yet.

That's what I thought which is why I went with a fully backend approach. The only unsubmitted changes on this page would be email and display name -- which is probably pretty edge casey.

compwron (approved):
Very cool!

PR #2021, opened 2021-05-08
Add test for CaseContactPopulator

### LINKED ISSUE #1256: add test for lib/tasks/data_post_processors/case_contact_populator.rb
add test for lib/tasks/data_post_processors/case_contact_populator.rb

librod89:
Resolves #1256

### What changed, and why?
Added a test for CaseContactPopulator and fixed a small error in the actual populator

It does not affect user permissions

New test

### DIFF
--- a/lib/tasks/data_post_processors/case_contact_populator.rb
+++ b/lib/tasks/data_post_processors/case_contact_populator.rb
@@ -9,7 +9,7 @@ module CaseContactPopulator
         unless ct
           if cts_by_name.any?
             ctg_name = cts_by_name.first.contact_type_group.name
-            org_ctg = ContactTypeGroup.where(casa_org: casa_org, name: ctg_name)
+            org_ctg = ContactTypeGroup.find_by(casa_org: casa_org, name: ctg_name)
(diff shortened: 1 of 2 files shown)


lib/tasks/data_post_processors/case_contact_populator.rb
```diff
@@ -9,7 +9,7 @@ def self.populate
         unless ct
           if cts_by_name.any?
             ctg_name = cts_by_name.first.contact_type_group.name
-            org_ctg = ContactTypeGroup.where(casa_org: casa_org, name: ctg_name)
+            org_ctg = ContactTypeGroup.find_by(casa_org: casa_org, name: ctg_name)
             if org_ctg
               ContactType.find_or_create_by!(contact_type_group: org_ctg, name: ct_name)
             else
```
librod89, line 15: I think the `else` on line 18 of this file is unreachable.

compwron: Hmm maybe... but find_by can always fail and return nil
It *should* tho...

Issue #1673, opened 2021-02-04
add more js tests


compwron:
add more js tests

FireLemons:
`app/javascript/src/case_contact.js` contains a lot of functions in a block. I think they still count as globals

PR #2042, opened 2021-05-13
Add cypress tests for volunteer actions

librod89:
#1673

### What changed, and why?
Added two new Cypress tests. One for a volunteer editing their profile and another for a volunteer generating a court report.

FYI, I'm learning Cypress as I go so I'm open to critique here.

It will not.


compwron (approved):
Very nice :)

PR #2064, opened 2021-05-19
cleaned up /spec/lib/importers/supervisor_importer_spec.rb

FireLemons:
Renamed file path variable
changed create(:admin) to build

removed a test testing for  volunteer import complete failure because there's already a test for partial failure


compwron (approved):
Very nice! :)

PR #2085, opened 2021-05-27
Create Problem validation

talya19:
### What changed, and why?
I created a new file to submit: problem validation because I thought we were lacking that type of file.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Problem validation
+++ b/.github/ISSUE_TEMPLATE/Problem validation
@@ -1,0 +1,10 @@
+Please use this template to document problems mentioned by CASA stakeholders so we make the best decisions for the solution.
+
+
+Which role has this problem (Admin, supervisor, volunteer)?
+
+What is the problem they are facing?(short sentence)
+
+When does it happen? At what moment in the workflow or after how long?
+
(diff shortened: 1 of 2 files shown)


compwron:
I don't understand what this means. Can you explain more?

talya19:
Hi Linda,
So I have created a new type of file when you click on « New issue ».
Basically you add this type of issue when it still needs information from
the stakeholders, when we need stakeholders and users to validate what type
of problem we are solving to offer the best solution.

Is that clearer?

compwron:
We have issue templates :)
Screen Shot 2021-06-01 at 9 55 19 PM
Screen Shot 2021-06-01 at 9 55 05 PM

talya19:
Hi @compwron, I know and I wanted to add this as a new issue template because it is neither documentation, nor a feature request, nor a chore nor a bug report.

compwron:
Oh, I thought you were trying to replace the existing ones sorry. this PR will not completely add the new one, you also need to add it to the config file. Let me know if you want help
https://github.com/rubyforgood/casa/blob/main/.github/ISSUE_TEMPLATE/config.yml
Please name files with underscores not spaces. :)

PR #2090, opened 2021-05-30
Handle error when generating court report fails

### LINKED ISSUE #2038: When generating court report fails for any reason, show an error
volunteer

**Description**
Currently, it only shows an error for errors of type `Zip::Error`
Please throw the error again so our error monitoring service bugsnag can give us a stack trace. 

see `app/controllers/case_court_reports_controller.rb`

_QA Login Details:_

- volunteer1@example.com
- supervisor1@example.com
- casa_admin1@example.com

hiendinhngoc:
Resolves #2038 

### What changed, and why?
- Rescue StandardError beside Zip::Error to handle error log when generating court report fails.

- Volunteer permissions:


compwron (approved):
:)
I would also like to see a test for this so we can't regress

PR #2143, opened 2021-06-12
tune search text for cases for court reports

compwron:
http://localhost:3000/case_court_reports
Screen Shot 2021-06-11 at 6 22 19 PM
Screen Shot 2021-06-11 at 6 22 11 PM

### DIFF
--- a/app/decorators/casa_case_decorator.rb
+++ b/app/decorators/casa_case_decorator.rb
@@ -54,12 +54,13 @@ class CasaCaseDecorator < Draper::Decorator
   end
 
   def court_report_select_option
+    volunteer_names = object.assigned_volunteers.map(&:display_name).join(",")
     [
-      "#{object.case_number} - #{object.has_transitioned? ? "transition" : "non-transition"}",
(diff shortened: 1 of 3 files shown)


compwron:
failing tests

rspec ./spec/system/case_court_reports/index_spec.rb:86 # case_court_reports/index 'Case Number' dropdown list has transition case option selected
rspec ./spec/system/case_court_reports/index_spec.rb:94 # case_court_reports/index 'Case Number' dropdown list has non-transitioned case option selected
rspec ./spec/system/case_court_reports/index_spec.rb:72 # case_court_reports/index when choosing the prompt option (value is empty) and click on 'Generate Report' button, nothing should happen Spinner does not become visible
rspec ./spec/system/case_court_reports/index_spec.rb:66 # case_court_reports/index when choosing the prompt option (value is empty) and click on 'Generate Report' button, nothing should happen 'Generate Report' button does not become disabled
rspec ./spec/system/case_court_reports/index_spec.rb:19 # case_court_reports/index when first arriving to 'Generate Court Report' page, by default shows a select element with default selection 'Select a case to generate report'
rspec ./spec/policies/supervisor_policy_spec.rb:43 # SupervisorPolicy update? does not allow supervisors to update other supervisors

PR #2153, opened 2021-06-13
replaced icon button with text button

FireLemons:
### What changed, and why?
Delete case mandate button is text now
image
used to look like
image


compwron (approved):
I like it :)

PR #2159, opened 2021-06-16
Refactor inline styles out of application layout

### LINKED ISSUE #2106: Refactor inline css out of app/views/layouts/application.html.erb
Refactor the styling into `.scss` files in `app/javascript/src/stylesheets`

for those unfamiliar with Sass, normal css works in `.scss` files

librod89:
Resolves #2106

### What changed, and why?
In order to move the inline styles from the application layout to a stylesheet, I needed webpacker to extract css files into the `public` folder since all of these files were not accessible within the `noscript` tag (i.e. without Javascript).

Now that the styles are available without Javascript, the site looks a lot better if a user navigates to the site with Javascript disabled, but they will still see a warning message on every page that Javascript is required (see screenshots below).

N/A

Existing tests

### Screenshots please :)
**Before**
image

**After**
image

### DIFF
--- a/app/javascript/src/stylesheets/application.scss
+++ b/app/javascript/src/stylesheets/application.scss
@@ -15,6 +15,7 @@
 @import "shared/sidebar";
 @import "shared/typography";
 @import "shared/utilities";
+@import "shared/noscript";
 
 @import "pages/casa_cases";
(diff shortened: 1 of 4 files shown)


compwron (approved):
Very cool! :)

PR #2220, opened 2021-07-01
revert styles to use inline for emails

FireLemons:
### What changed, and why?

Moved css from style tag back to inline to support more email platforms


compwron:
That one failing rspec test is "July 01" vs "July 1" something er fixed in office hours on Anthony's PR, ignore it for now

FireLemons:
ah shoot https://stackoverflow.com/a/11293350

compwron (approved):

PR #2239, opened 2021-07-07
add docker build+tests to CI so we stop breaking it unknowingly

compwron:
see r4g slack
https://rubyforgood.slack.com/archives/CVB0QJGVD/p1625501851139400
follow-up to https://github.com/rubyforgood/casa/pull/2238 so we will notice faster when we break docker, since so many people use it. Eventually we should remove the rspec build and just do the docker build, but let's have both at once for long enough to make sure that they reliably break at the same time.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,33 @@
+name: docker
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - 'doc/**'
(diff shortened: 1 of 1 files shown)


compwron:
split this into 2 PRs, they don't belong together

compwron:
Let's go with https://github.com/rubyforgood/casa/pull/2320 instead :)

PR #2251, opened 2021-07-09
Make sidebar height 100vh regardless of image size

### LINKED ISSUE #2152: Sidebar height set incorrectly for custom logos
**What types of user does this feature affect?**
 - volunteers
 - supervisors
 - admins

**Description**
If the logo is too short vertically, the sidebar doesn't extend all the way down. Maybe it overflows too much if the logo is too tall.

**Screenshots of current behavior**  
image

_QA Login Details:_  

Login Emails: 

aedwardg:
Resolves #2152 

### What changed, and why?
Change `.sidebar-wrapper` height to have a max-height of 100% of the
viewport and get rid of the calculated height on `.sidebar-container`.
This should ensure that the sidebar is always the size of the viewport,
no matter the image size and any extra height will scroll.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

I don't think we need a test for this.  If we do, let me know how and I'll write it up.

### Screenshots please :)
image

### Feelings gif (optional)
How to embed:
CSS hates me

### DIFF
--- a/app/javascript/src/stylesheets/shared/sidebar.scss
+++ b/app/javascript/src/stylesheets/shared/sidebar.scss
@@ -5,6 +5,7 @@
 .sidebar-wrapper {
   background: rgba(0,0,0,0.25);
   min-height: 100vh;
+  max-height: 100vh;
   position: fixed;
   width: 320px;
   overflow: auto;
@@ -44,7 +45,6 @@
   display: flex;
   flex-direction: column;
   justify-content: space-between;
-  height: calc(100vh - 144px);
   overflow-y: auto;
 }
 


aedwardg:
Hm...My change had nothing to do with this test failure, but that's interesting.
Looks like it was looking for "Denis O'Connell" but the html has "Denis O&#39;Connell"

compwron (approved):

FireLemons:
@aedwardg someone's fixing that already

FireLemons:
@aedwardg didn't quite get it.  
The rule you deleted for `.sidebar-container` tried to expand the sidebar when there wasn't enough menu items to fill it out.  
  
I'm not sure adding `max-height` does anything because `min-height` overrides `max-height` and `height`. See https://css-tricks.com/almanac/properties/m/min-height/  

image
image  
Even with the rule in place you can see the sidebar still doesn't touch the bottom all the way.  

I think flexbox might be the solution here. See https://css-tricks.com/snippets/css/a-guide-to-flexbox/

aedwardg:
@FireLemons `max-height` definitely does something here, it is triggered when the sidebar has enough items to expand past 100vh.  You can verify this by signing in as an admin and then getting rid of the `max-height` rule.  Without it, the sidebar will expand past the screen but will not be scrollable.   As far as `min-height` overriding it, I think that article is somewhat misleading.  My understanding is that `min-height` will always override `height` and `max-height` _if_ `min-height` is greater than them.  See https://developer.mozilla.org/en-US/docs/Web/CSS/min-height

You are totally right though about when the sidebar does not have enough items.  I did not check for that, so that's my mistake.  With that in mind, my solution might not be the one we need, so I'll keep looking into it.  I'm no CSS wizard, so it might take a bit of playing around 😂 .  Thanks for catching this!

Issue #1602, opened 2021-01-16
Fix 3 CodeQL security Code Scanning issues


compwron:
https://github.com/rubyforgood/casa/security/code-scanning
<img width="1184" alt="Screen Shot 2021-01-16 at 2 33 13 PM" src="https://user-images.githubusercontent.com/578159/104824581-d36bba00-5807-11eb-9ee9-81235eda66be.png">

chinmayjoshi:
Hi , @Zrrrpy @compwron I would like to work on this one

compwron:
Very cool! All yours :)

PR #2269, opened 2021-07-14
Fix DOM text reinterpreted as HTML

librod89:
Addresses #1602

### What changed, and why?
- Fix CodeQL workflow
- Fix CodeQL security warning of "DOM text reinterpreted as HTML" by using `textContent` property (which is text only) instead of `text()`
See explanation of this CodeQL error [here](https://codeql.github.com/codeql-query-help/javascript/js-xss-through-dom/)

N/A

New system test

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,22 +9,25 @@
 # the `language` matrix defined below to confirm you have the correct set of
 # supported CodeQL languages.
 #
-name: "CodeQL"
+name: CodeQL
 
 on:
   push:
-    branches: [ main ]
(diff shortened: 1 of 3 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -9,22 +9,24 @@
 # the `language` matrix defined below to confirm you have the correct set of
 # supported CodeQL languages.
 #
-name: "CodeQL"
+name: CodeQL
```
librod89, line 12: These changes will allow CodeQL to run for PRs now

FireLemons (approved):

PR #2273, opened 2021-07-15
close style tags

FireLemons:
close email style tags


compwron:
without this, emails appear to have no contents, which is bad

compwron (approved):

PR #2317, opened 2021-07-25
Adds "re-invite" button in weekly digest email

### LINKED ISSUE #2210: Refactor inline styling into a <style> element for app/views/supervisor_mailer/weekly_digest.html.erb
**Description**
Refactor all inline css inside a `<style>` element

efgalvao:
Resolves #2210

### What changed, and why?
Added a list to weekly digest if the supervisor has invited a volunteer and he not accepted it yet,  changed the resend invitation method so it could be called from the button in e-mail, also created a method to list all pending invitations for a specific supevisor.

It will not.

There are some tests that checks for the volunteer in the email if he not accepted and another one when the volunteer has accepted.

### Screenshots please :)
Screenshot from 2021-07-25 19-52-33

### Feelings gif (optional)
efgalvao

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -90,7 +90,7 @@ Rails.application.routes.draw do
     member do
       patch :activate
       patch :deactivate
-      patch :resend_invitation
+      get :resend_invitation
       patch :reminder
     end
   end
--- a/spec/mailers/supervisor_mailer_spec.rb
+++ b/spec/mailers/supervisor_mailer_spec.rb
@@ -61,6 +61,19 @@ RSpec.describe SupervisorMailer, type: :mailer do
       create(:case_assignment, casa_case: casa_case, volunteer: volunteer, active: false, updated_at: Date.today - 8.days)
       expect(mail.body.encoded).to_not match("Summary for #{volunteer.display_name}")
     end
+
+    context "when a supervisor has pending volunteer to accepts invitation" do
+      let!(:volunteer2) { create(:volunteer) }
+      let!(:invite) { volunteer2.invite!(supervisor) }
+
+      it "shows a summary of pending volunteers" do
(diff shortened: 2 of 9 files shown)


spec/mailers/supervisor_mailer_spec.rb
```diff
@@ -61,6 +61,26 @@
       create(:case_assignment, casa_case: casa_case, volunteer: volunteer, active: false, updated_at: Date.today - 8.days)
       expect(mail.body.encoded).to_not match("Summary for #{volunteer.display_name}")
     end
+
+    context "when a supervisor has pending volunteer to accepts invitation" do
+      let!(:volunteer2) { create(:volunteer) }
+      let!(:invite) { volunteer2.invite!(supervisor) }
```
wenderjean, line 67: It doesn't seem you're using `invite`, I mean, it'd probably fit better into a `before` block.

config/routes.rb
```diff
@@ -92,7 +92,7 @@
     member do
       patch :activate
       patch :deactivate
-      patch :resend_invitation
+      get :resend_invitation
```
compwron, line 95: I am very ambivalent about a modifying operation being a get. I would prefer patch or post. Why does this need to be a get? Does it have to do with calling it from an email?

compwron (approved):

config/routes.rb
efgalvao: Exactly, can call a patch from email.

compwron: Gotcha gotcha
Maybe someday we will find a better way, but at least it's authenticated 😅

compwron:
Very cool!! 🎉✨

config/routes.rb
efgalvao: Exactly.

PR #2339, opened 2021-07-30
Fix erblint github action so it will actually fail

librod89:
N/A

### What changed, and why?
- I noticed that a r[ecent successful erblint GitHub action](https://github.com/rubyforgood/casa/runs/3193964212) actually had failures. I have changed the GitHub action so it will fail the action when this happens.
- If you look at the first commit of this PR, you can see that the action failed correctly.

image

N/A

N/A


compwron (approved):
Excellent! Thank you :)

PR #2401, opened 2021-08-13
montgomery court report cleanup

FireLemons:
### What changed, and why?  
 - fixed afterparty task
 - alphabetized tests and methods
 - moved latest court date logic into casa case model  
 - writing test for casa_case method in casa_case model


compwron (approved):
:)

PR #2403, opened 2021-08-14
Fix file download tests in docker (locally)

### LINKED ISSUE #2246: Fix file download tests in docker
see https://github.com/SeleniumHQ/docker-selenium/issues/548

```
Failures:

  1) case_contact_reports/index filters report by date and selected contact type
     Failure/Error: sleep 20 until downloaded?

     Timeout::Error:
       execution expired

     [Screenshot Image]: /usr/src/app/tmp/screenshots/failures_r_spec_example_groups_case_contact_reports_index_filters_report_by_date_and_selected_contact_type_33.png


     # ./spec/support/download_helpers.rb:20:in `sleep'
     # ./spec/support/download_helpers.rb:20:in `block in wait_for_download'
     # ./spec/support/download_helpers.rb:19:in `wait_for_download'
     # ./spec/system/reports/case_contact_reports_spec.rb:25:in `block (2 levels) in <top (required)>'
     # ./spec/rails_helper.rb:69:in `block (2 levels) in <top (required)>'

```

librod89:
Addresses #2246

### What changed, and why?
- This is only the **first** step to fixing these errors. Now, these tests pass locally. The last step must be related to GitHub actions, but I'm having trouble figuring out the last piece here.
- By using `:selenium_chrome_in_container` and `vnc://localhost:5900`, I was able to spy on the test as it was running in docker. I could see that the download was working correctly. It was downloading to `/home/seluser/Downloads` so I created a shared volume from the host (/tmp/downloads/) to the container

N/A

The existing tests pass locally!

### Feelings gif (optional)
Alt Text

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -26,10 +26,6 @@ jobs:
         uses: docker/setup-buildx-action@v1
       - name: Build
         run: docker-compose build
-      - name: bundle install
-        run: docker-compose run web bundle install
-      - name: db:reset
-        run: docker-compose run web rails db:reset
       - name: docker UP
         run: docker-compose up -d
       - name: Test
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,8 +22,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2.3.4
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
       - name: Build
         run: docker-compose build
       - name: bundle install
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -22,15 +22,9 @@ jobs:
...
       - name: Checkout repository
         uses: actions/checkout@v2.3.4
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
-      - name: Build
-        run: docker-compose build
-      - name: bundle install
-        run: docker-compose run web bundle install
-      - name: db:reset
-        run: docker-compose run web rails db:reset
```
librod89, line 32: None of these other steps were needed so this should be up the action.

compwron (approved):
🎉🎉🎉💪

PR #2442, opened 2021-08-20
Correct node version and rename gh workflow

FireLemons:
### What changed, and why?

The workflow doesn't just lint. It also runs [jest](https://jestjs.io/) 
Our current node version is 12


compwron (approved):
Can you make a ticket to upgrade this? xD

FireLemons:
@compwron the node version? We already have a ticket to upgrade to node 14 LTS

PR #2454, opened 2021-08-21
Fix flaky seed test

### LINKED ISSUE #2430: Fix flaky seed test
This is probably (?) caused by factory run order - one of these (or all of them) `CasaOrg.first || create(:casa_org)`

```
  1) Seeds test development DB successfully populates all necessary tables
     Failure/Error: ActiveRecord::Tasks::DatabaseTasks.load_seed

     ActiveRecord::RecordInvalid:
       Validation failed: Volunteer and case must belong to the same organization
     # ./db/seeds/db_populator.rb:189:in `block in create_cases'
     # ./db/seeds/db_populator.rb:160:in `times'
     # ./db/seeds/db_populator.rb:160:in `create_cases'
     # ./db/seeds/db_populator.rb:47:in `create_org'
     # ./db/seeds.rb:35:in `block in seed'
     # ./db/seeds.rb:33:in `times'
     # ./db/seeds.rb:33:in `seed'
     # ./db/seeds.rb:102:in `<top (required)>'
     # ./spec/seeds/seeds_spec.rb:32:in `block (3 levels) in <top (required)>'
```

librod89:
Resolves #2430

### What changed, and why?
- In `create_cases`, if there was already a CasaCase that existed for a generated case_number, that case would be used to generate volunteers. But since that CasaCase was not just created with the script, the casa_org was not correct with what was given as a param to the volunteer. So I fixed this by using `new_casa_case.casa_org` as the param so it's always the correct organization.

N/A

I could reliably get `seed_spec.rb` to fail roughly 1 in 5 times that I ran it locally. I could never get it to fail after this change.

### DIFF
--- a/db/seeds/db_populator.rb
+++ b/db/seeds/db_populator.rb
@@ -178,10 +178,10 @@ class DbPopulator
           )
         end
 
-      volunteer = casa_org.volunteers.active.sample(random: rng) ||
-        casa_org.volunteers.active.first ||
+      volunteer = new_casa_case.casa_org.volunteers.active.sample(random: rng) ||
+        new_casa_case.casa_org.volunteers.active.first ||
         Volunteer.create!(
-          casa_org: casa_org,
+          casa_org: new_casa_case.casa_org,
           email: "#{SecureRandom.hex(10)}@example.com",
           password: SEED_PASSWORD,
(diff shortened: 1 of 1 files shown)


compwron (approved):
Excellent :)

PR #2489, opened 2021-08-29
fix emancipations not saving if case is of transitioning age

FireLemons:
### What changed, and why?
Emancipation checklist api calls check for transition age instead of the boolean before saving to the database

Updated existing tests


compwron (approved):
Looks good, some failing tests tho

FireLemons:
Something is missing a period in that thank you message test

PR #2563, opened 2021-09-10
Past notifications should reflect current name of user

### LINKED ISSUE #2471: Make Names Reflect Current Name in Past Notifications
 - volunteers
 - supervisors
 - admins

**Description**
Currently when someone updates their name, their name does not update in past notifications. We would like to change this so that the names appearing in notifications always reflect the current name of the person being referenced. 

Here we see Whitney Block who used to be named Volunteer 1 still has a bunch of notifications referencing their old name:
Names Should Update When Changed

_QA Login Details:_  

Login Emails: 

r-sierra:
Allows to get the users current name when showing the notification message

Resolves #2471

### What changed, and why?
Notifications should reflect the current name of the user that created it. To accomplish this instead of passing the name of the user as a parameter we pass the user instance.

No permission affected

with tests

### DIFF
--- a/app/notifications/followup_notification.rb
+++ b/app/notifications/followup_notification.rb
@@ -52,7 +52,7 @@ class FollowupNotification < Noticed::Base
   def created_by_name
     if params.key?(:created_by)
       params[:created_by][:display_name]
-    else  # keep backward compatibility with older notifications
+    else # keep backward compatibility with older notifications
       params[:created_by_name]
     end
   end
--- a/app/notifications/followup_notification.rb
+++ b/app/notifications/followup_notification.rb
@@ -3,7 +3,7 @@
 # FollowupNotification.with(followup: @followup).deliver_later(current_user)
 # FollowupNotification.with(followup: @followup).deliver(current_user)
 
-class FollowupNotification < Noticed::Base
+class FollowupNotification < BaseNotification
   # Add your delivery methods
   #
   deliver_by :database
@@ -17,10 +17,6 @@ class FollowupNotification < Noticed::Base
 
   # Define helper methods to make rendering easier.
   #
-  def title
-    t(".title")
-  end
-
   def message
     note = params[:followup][:note]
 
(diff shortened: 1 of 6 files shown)


app/notifications/followup_notification.rb
```diff
@@ -46,6 +46,14 @@ def message_without_note
...
 
   def message_heading
-    t(".message", created_by_name: params[:created_by_name])
+    t(".message", created_by_name: created_by_name)
+  end
+
+  def created_by_name
+    if params.key?(:created_by)
+      params[:created_by][:display_name]
+    else # keep backward compatibility with older notifications
```
compwron, line 55: wonderful :)

app/notifications/followup_resolved_notification.rb
```diff
@@ -22,10 +22,20 @@ def title
...
+    t(".message", created_by_name: created_by_name)
   end
 
   def url
     edit_case_contact_path(params[:followup][:case_contact_id], notification_id: record.id)
   end
+
+  private
+
+  def created_by_name
```
compwron, line 34: duplicate code block- can we combine them?

spec/system/notifications/index_spec.rb
```diff
@@ -12,19 +12,40 @@
...
+    it "lists my notifications" do
       sign_in admin
       visit notifications_path
 
       expect(page).to have_text(notification_message)
       expect(page).to have_text("Followup resolved")
       expect(page).not_to have_text(I18n.t(".notifications.index.no_notifications"))
     end
+
+    context "when volunteer changes its name" do
```
compwron, line 31: Please add a test for old notifications being displayable

compwron (approved):

app/notifications/followup_resolved_notification.rb
r-sierra: Sure

spec/system/notifications/index_spec.rb
r-sierra: Not sure we can do that, as notifications fail if required params are not filled on save. I'll see what i can do

compwron:
nice! :)

PR #2594, opened 2021-09-16
fix webpacker server issue

compwron:
https://docs.stimulusreflex.com/hello-world/setup#heads-up-issue-with-webpacker

This was not caught by ci because we only run docker in ci, not the webserver. Maybe we should change that...


cjilbert504 (approved):
🤞

PR #2595, opened 2021-09-16
remove old data migration that errors when running migrations

compwron:


db/schema.rb
```diff
@@ -312,7 +312,8 @@
   end
 
   create_table "versions", force: :cascade do |t|
-    t.string "item_type", null: false
+    t.string "item_type"
+    t.string "{:null=>false}"
```
littleforest, line 316: This actually seems really weird, are we sure this is right?

PR #2596, opened 2021-09-16
Spec refactor

littleforest:
### What changed, and why?
Refactor spec to build up test within `it` block for readability, easy of debugging, and helpful for newer developers. Going to work on some N+1 issues with the related controller, so refactoring this spec first. Also combined a few specs that were just testing a single thing to help speed up the test suite.


compwron (approved):
Nice! :)

PR #2607, opened 2021-09-17
Sort Case Court's Contact Dates

### LINKED ISSUE #2602: Fix Order of Case Contacts in Court Report
### Impacted User Types
 - volunteers
 - supervisors
 - admins

### Environment
Production

### Current Behavior
When a Court Report is generated Case Contacts appear in the order they were entered into the system instead of the order that they occurred in. In the below image see the rows for Medical Professional, Sibling, Psychiatric Practitioner, and Grandparent for examples.
Case Contact Dates Out of Order in Court Report

### Expected Behavior
When a Court Report is generated Case Contact dates should appear in the order they occurred regardless of the order they were entered into our system. 

### How to Replicate
1. - Log into qa as volunteer1@example.com password: 123456.
2. - Click on "Generate Court Reports" in the left sidebar menu.
3. - Select Case Number CINA-07-1152 from the dropdown.
4. - Click "Generate report" at the bottom of the page.
5. - Examine the generated report, specifically paying attention to the case contact dates in the rows for Medical Professional, Sibling, Psychiatric Practitioner, and Grandparent. 

see https://github.com/rubyforgood/casa/wiki/How-to-edit-docx-templates

josearmandojacq:
Resolves #2602

### What changed, and why?
Sort ascendant the Court Dates 

I tried to write a test but case_court_report just beat me and couldn't figure out how to 😞 

### Screenshots please :)

Screenshot 2021-09-17 at 18 38 58

### Feelings gif (optional)
Happy but Sad

### DIFF
--- a/app/models/case_court_report.rb
+++ b/app/models/case_court_report.rb
@@ -88,7 +88,15 @@ class CaseCourtReport
       contact_dates[contact_type] = [date_with_format]
     end
 
-    contact_dates
+    sort_dates(contact_dates)
+  end
+
+  def sort_dates(contact_dates)
+    contact_dates.each_value do |dates|
+      next if dates.size <= 1
+
+      dates.sort! { |first_date, second_date| first_date.delete("*") <=> second_date.delete("*") }
+    end
   end
 
   def prepare_case_details


app/models/case_court_report.rb
```diff
@@ -88,7 +88,15 @@ def aggregate_contact_dates(people)
...
-    contact_dates
+    sort_dates(contact_dates)
+  end
+
+  def sort_dates(contact_dates)
+    contact_dates.each_value do |dates|
+      next if dates.size <= 1
+
+      dates.sort! { |first_date, second_date| first_date.delete("*") <=> second_date.delete("*") }
+    end
```
littleforest, line 99: Can I help you write a spec for this? Did you have something that you tried but didn't work that you can push up?

app/models/case_court_report.rb
```diff
@@ -88,7 +88,15 @@ def aggregate_contact_dates(people)
       contact_dates[contact_type] = [date_with_format]
     end
 
-    contact_dates
+    sort_dates(contact_dates)
+  end
+
+  def sort_dates(contact_dates)
+    contact_dates.each_value do |dates|
```
littleforest, line 95: I did not know about the `each_value` method, very cool.

app/models/case_court_report.rb
josearmandojacq: Hi @littleforest thanks a lot. I tried to test/find in the spec withing the report, that the dates  appear sorted. But I got lost while searching. 😞

littleforest: @josearmandojacq I thought it wouldn't be too difficult to test this method, but I just noticed that this a private method, so it shouldn't be tested directly. The `case_court_report` spec is a little bit hard to follow - can you add an expectation after [this line](https://github.com/rubyforgood/casa/blob/main/spec/models/case_court_report_spec.rb#L79)? (Are the dates on line 73-75 what you are trying to sort, or am I looking at the wrong place?) I am not very familiar with this piece of the code, so I might not be able to be as helpful as I thought I could.

compwron: If there is a private method that we want to test directly, we could break it out to a service class and test that new class directly

josearmandojacq: @littleforest, @compwron  It is not about testing a private method, but instead testing that in the report the dates appear ordered ascendant. I tried to do it while looking into the string/xml created in the test, but it was a lot and couldn't find the proper way to test it. 😁

josearmandojacq: Happy you are commenting on it 💪🏽 💪🏽

josearmandojacq: I will try again in these days because I don't have too much time now. But feel free @compwron to merge it if you think it is good enough based in the priority :)

PR #2626, opened 2021-09-21
refactoring and testing for sorting case contact dates in court report

compwron:
refactoring and testing for https://github.com/rubyforgood/casa/pull/2607

PR #2629, opened 2021-09-21
refactor CaseContactsContactDates

compwron:
follow-up to https://github.com/rubyforgood/casa/pull/2626 which was a follow-up to https://github.com/rubyforgood/casa/pull/2607


app/services/court_report_format_contact_date.rb
```diff
@@ -0,0 +1,11 @@
+class CourtReportFormatContactDate
+  CONTACT_UNSUCCESSFUL_PREFIX = "*"
```
FireLemons, line 2: I think this would be suffix

PR #2637, opened 2021-09-22
Fix case prefix filter

### LINKED ISSUE #2618: Fix CASA Case Prefix Filter
### Impacted User Types
 - volunteers
 - supervisors
 - admins

### Environment
Production

### Current Behavior
When a prefix is selected from the CASA Case Prefix Filter on the Cases page, cases do not filter.
Screen Shot 2021-09-17 at 5 03 48 PM

### Expected Behavior
When a prefix is selected from the CASA Case Prefix Filter only cases with that prefix should be shown. 

### How to Replicate
1. - Log into qa or staging as an admin or supervisor.
2. - Click on "Cases" in the left sidebar menu.
3. - Attempt to Use the CASA Case Prefix Filter on the top right.

Beigelman:
Resolves #2618

### What changed, and why?

The javascript function that handles the table filter was changed to also filter based on the case number prefix

### Screenshots please :)

image
image

### DIFF
--- a/app/javascript/src/dashboard.js
+++ b/app/javascript/src/dashboard.js
@@ -68,11 +68,12 @@ $('document').ready(() => {
         }
       })
 
+      const possibleCaseNumberPrefixes = ['CINA', 'TPR']
       const status = data[3]
       const assignedToVolunteer = (data[5] !== '' && data[5].split(',').length >= 1) ? 'Yes' : 'No'
       const assignedToMoreThanOneVolunteer = (data[5] !== '' && data[5].split(',').length > 1) ? 'Yes' : 'No'
       const assignedToTransitionYouth = data[4]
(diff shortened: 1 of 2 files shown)


compwron:
This looks promising! You have some failing tests to fix, and please also add a test to make sure that this fixed functionality keeps working :)

app/javascript/src/dashboard.js
```diff
@@ -68,15 +68,18 @@ $('document').ready(() => {
         }
       })
 
+      const possibleCaseNumberPrefixes = ['CINA', 'TPR']
```
compwron, line 71: This is probably going to break with our new CASAs with new prefixes. It's fine for now as long as it doesn't break no-prefix case numbers

compwron (approved):
:) :) :)

PR #2639, opened 2021-09-23
Removes css-p from footer scss; reduces padding-top to move logo and …

### LINKED ISSUE #2353: Set Up Password Page Footer Is In The Way
### Impacted User Types
 - volunteers
 - supervisors
 - admins

### Environment
Production

### Current Behavior
When I try to set the password on a new user, the page draws in such a way as to obscure the set password fields.
Acceptin Invitation Update

### Expected Behavior
When a new user attempts to set their password, the password fields should be visible/accessible. 

### How to Replicate
1. - Log into your local environment as an admin.
2. - Create a user.
3. - Click the "Set Your Password" prompt in the email.

armahillo:
Resolves #2353 

### What changed, and why?
The issue was that the footer on the password page was set using css absolution positioning. Absolute positioning does weird stuff because it puts the content into a separate stacking index, meaning it doesn't sit the same with floats and z-index, among other things. It wasn't necessary, so I just struck it out entirely.

It felt like the footer content sat a little too far down the page so I reduced the padding-y from 5rem to 2rem. I think you could probably put some of teh footer content side by side too, but that felt out of scope for this

Visual change only; tested with DOM inspector manually.

### Screenshots please :)

Full-width -- the footer content can be seen by scrolling down. I think this could probably be made a little more responsive (that's a lot of whitespace above and below) but the behavior to correct in this was just the inability to edit the password fields because of coverup so I didn't touch that.

Screen Shot 2021-09-22 at 11 08 09 PM

Narrowest view -- this is what would be seen on mobile I presume. I couldn't make it any narrower. It's still visible and functional.

Screen Shot 2021-09-22 at 11 08 24 PM

### Feelings gif (optional)
What gif best describes your feeling working on this issue? 
dinosaur on bike

### DIFF
--- a/app/javascript/src/stylesheets/pages/password.scss
+++ b/app/javascript/src/stylesheets/pages/password.scss
@@ -32,9 +32,6 @@
 
 body.devise-invitations.devise-invitations-edit footer {
   width: 100%;
-  position: absolute;
-  bottom: 0;
-  left: 0;
 }
 
 body.devise-passwords.devise-passwords-edit div.password-box {
--- a/app/javascript/src/stylesheets/shared/footer.scss
+++ b/app/javascript/src/stylesheets/shared/footer.scss
@@ -1,5 +1,5 @@
 footer {
-  padding: 5rem 0;
+  padding: 2rem 0;
   text-align: center;
   background-color: $primary;
   color: #fff;


compwron (approved):
Yay!!!!!

PR #2568, opened 2021-09-13
[WIP] Refactor: Remove transition_aged_youth from casa_cases

hiendinhngoc:
### What github issue is this PR for, if any?
Resolves #2145

### How will this affect user permissions?
- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

### Feelings gif (optional)
What gif best describes your feeling working on this issue? https://giphy.com/
How to embed:
`alt text`

PR #2646, opened 2021-09-24
Feature/#2145 remove casa case transition aged youth

compwron:
Continuation of https://github.com/rubyforgood/casa/pull/2568


compwron:
Closing for now, remake it if the original doesn't revamp

PR #2675, opened 2021-10-01
Make Miles Driven Field Empty When Clicked On

### LINKED ISSUE #2433: Make Miles Driven Field Empty When Clicked On
 - volunteers
 - supervisors
 - admins

**Description**
Currently the Miles Driven field in a new Case Contact is set to zero by default. This is good. However, when someone clicks into the Miles Driven field the zero does not currently disappear to make way for the new data, so the data ends up looking like the second screenshot below. We would like the zero to disappear when the Miles Driven field is clicked into so that the field is empty for the new data.

Miles Driven
Miles Driven With Data

_QA Login Details:_  

Login Emails: 

AravindSelvamani:
Resolves #2433

### What changed, and why?
Currently, the Miles Driven field in a new Case Contact is set to zero by default. This is good. However, when someone clicks into the Miles Driven field the zero does not currently disappear to make way for the new data, so the data ends up looking like the second screenshot below. We would like the zero to disappear when the Miles Driven field is clicked so that the field is empty for the new data.

- Volunteer:
- Supervisor:
- Admin:

### Screen recording, please :)

https://user-images.githubusercontent.com/39373592/135666265-1e177de6-ceff-43fa-a043-435bdb396d38.mov


spec/system/case_contacts/new_spec.rb
```diff
@@ -177,6 +180,7 @@
 
         fill_in "case-contact-duration-hours", with: "1"
         fill_in "case-contact-duration-minutes", with: "45"
+        fill_in "Miles driven", with: "0"
```
compwron, line 183: This should not be needed- empty field should not make save invalid, that would be bad user experience. I think the right solution here is to make the controller smart enough to translate `''` to "no value" / nil / 0 and not try to set the value.

compwron (changes requested):
not setting a value should dofault to correct 0 and not impede saving

Issue #2647, opened 2021-09-24
API


FireLemons:
**Description**
We'd like to accomplish many things with mobile development like offline service workers and creating an iOS app but to do so we need our routes to work like an api.

 - For every controller action that changes the database, make sure it supports a json format.
 - Error check/sanitize parameters
 - return a status of 500 if an error was encountered
 - Write tests
 - Document the api please edit the api readme in `doc/`

Action list:
Work In Progress List
 - All Casa Admin
   - [x] create
   - [x] update
   - [x] update_password
 - Casa Admin
   - [x] create
   - [x] update
   - [x] deactivate
   - [x] activate
 - Casa Case   
   - [x] create
   - [x] update
   - [x] deactivate
   - [x] reactivate
 - Casa Org (`app/controllers/all_casa_admins/casa_orgs_controller.rb`)
   - [x] create
 - Casa Org (`app/controllers/casa_orgs_controller.rb`)
   - [x] update
 - Case Assignments
   - [ ] create
   - [ ] unassign

Please do not do this all in a single pull request

m-pereira:
I'm not sure about this idea, but I'd like to propose some solution. I'll start by `POST /casa_cases` route

m-pereira:
I'd like to know, why it is necessary, for example, can we use the `respond_to with format` helper, in those controllers?

FireLemons:
> can we use the `respond_to with format` helper  
  
That's originally what I had planned. I was just trying to get a list of all the routes that made changes to the database. Maybe I should look at each controller for a list.

m-pereira:
> > can we use the `respond_to with format` helper
> 
> That's originally what I had planned. I was just trying to get a list of all the routes that made changes to the database. Maybe I should look at each controller for a list.

Nice, then we do not need that implementation and go on with `respond_to` helper. I'll adjust the [pr](https://github.com/rubyforgood/casa/pull/2686)

FireLemons:
thanks I'll clean up the issue description too when I have a chance

m-pereira:
Hey guys, I'm working on Casa Cases endpoints 😃

efgalvao:
Hi, can I work on Casa Orgs endpoints ?

mdchaney:
I actually created an issue regarding this, but the /casa_orgs endpoint should be a singular resource.  It might make sense to fix that while working on it.  That's #2919

efgalvao:
Ok, first I will work on that one.

[...]

PR #2686, opened 2021-10-02
Add json response functionality Starting with 'POST /casa_cases' route

mauricio-prl:
References https://github.com/rubyforgood/casa/issues/2647

### What changed, and why?

For every route that changes the database, we need to make sure it supports rendering `json` when passed an `api: true` parameter.

with `rspec`, adding some integration request tests for the route chosen.


lib/core_extensions/string.rb
```diff
@@ -0,0 +1,7 @@
+class String
```
compwron, line 1: This seems like a bit much. I don't want monkeypatching in here, we shouldn't need it for any reason. Let's find a different way...

m-pereira: I agree, for some reason, it made sense in my mind when I was coding it.

spec/requests/casa_cases_spec.rb
```diff
@@ -111,6 +111,13 @@
           expect(casa_case.hearing_type).to eq hearing_type
           expect(casa_case.judge).to eq judge
         end
+
+        it "also respond to json", :aggregate_failures do
+          post casa_cases_url(format: :json), params: {casa_case: valid_attributes}
+
+          expect(response.content_type).to eq("application/json; charset=utf-8")
+          expect(response).to have_http_status(:created)
```
compwron, line 119: I would like to also see an assert of the body contents here to make sure the fields come through correctly.

compwron (approved):

PR #2689, opened 2021-10-02
Expanded tool installation for macs

acrosman:
### What changed, and why?

The Mac instructions were a bit thin, so I took a minute to expand to cover all the basic applications.  

n/a

Documentation only.


littleforest:
Thanks! I updated the ticket number in the description.

FireLemons (approved):
That's a lot better than what we have currently

Issue #2684, opened 2021-10-02
Refactor and add some tests for dashboard.js


m-pereira:
**Description**

We need to refactor `app/javascript/src/dashboard.js` file, to avoid codeclimate warnings like [this](https://codeclimate.com/github/rubyforgood/casa/pull/2652). And then write some tests :)

**Existing gem(s), file(s) needing updating, changing or deleting**

app/javascript/src/dashboard.js

compwron:
Closing as old

PR #2690, opened 2021-10-03
Improve mobile UI for volunteers table

### LINKED ISSUE #2624: Make Table Labelling Consistent on Mobile
 - supervisors
 - admins

**Description**
Currently when screen size is sufficiently small, the tables for volunteers and cases are hard to read because they have no labels. The supervisors table however has labels and is easy to interact with at any size. We would like to make the volunteers and cases tables look like the supervisors table with appropriate labels.

We want the table on the right to look more like the table on the left:
More Readable Collapsed Tables

_QA Login Details:_  

Login Emails: 

acrosman:
Resolves #2624

### What changed, and why?
Added mobile labels to volunteers table.

n/a

There didn't appear to be automated tests to extend for this, so it was manually tested. But it looks like #2684 suggests a lot of work is needed in this area, so it probably belongs in that process anyway.

### Screenshots please :)

image

### DIFF
--- a/app/javascript/src/dashboard.js
+++ b/app/javascript/src/dashboard.js
@@ -166,11 +166,11 @@ $('document').ready(() => {
             <a href="${casaCasePath(casaCase.id)}">${casaCase.case_number}</a>
             `
           })
-          var case_numbers = `
+          var caseNumbers = `
             <span class="mobile-label">Case Number(s)</span>
             ${links.join(', ')}
           `
-          return case_numbers
+          return caseNumbers
         },
         orderable: false
       },
--- a/app/javascript/src/dashboard.js
+++ b/app/javascript/src/dashboard.js
@@ -113,6 +113,7 @@ $('document').ready(() => {
         name: 'display_name',
         render: (data, type, row, meta) => {
           return `
+            <span class="mobile-label">Name</span>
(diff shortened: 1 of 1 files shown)


acrosman:
I'm looking at the code climate feedback, and while I don't disagree, given there is an issue to refactor this file in part to avoid warnings of that nature I'm not sure if/how it should be addressed.

compwron (approved):
Looking good!

Please add tests and either fix or mark ignored the code climate issues
And then we will merge! :)

acrosman:
Happy to make adjustments, but not clear how to make sure to align to the two asks since they are both also related to a larger need in issue #2684 and I don't want to conflict that effort.

Also, I tried and failed to tell code climate to ignore the repetitive code issue – are there directions for the right way to do that?

compwron:
> Happy to make adjustments, but not clear how to make sure to align to the two asks since they are both also related to a larger need in issue #2684 and I don't want to conflict that effort.
> 
> Also, I tried and failed to tell code climate to ignore the repetitive code issue – are there directions for the right way to do that?

Let's keep working on the codeclimate fix- 
https://docs.codeclimate.com/docs/advanced-configuration
https://stackoverflow.com/search?q=%5Bcode-climate%5D+ignore

PR #2724, opened 2021-10-10
Adds Request Spec verifying Judge Name in DOCX for past court dates

armahillo:

Closes #2517

### What changed, and why?
I added two new request specs to past_court_date_spec, similar to the
way they were done in the regular case report request spec. The DOCX
format is really hard to read and most of it is superfluous XML so I
wrote an additional helper method that strips out all the XML and
reveals just the underlying content. This spec was only needing to
verify that a specific content string was present, so it was enough.

That new helper is added alongside the existing (awesome!) helper for
this need.

The request specs themselves test both the affirmative and negative
cases of having a judge or not, since the judge's presence is optional.

ADDITIONAL CLEANUP:
I noticed that the PastCourtDate model had an extraneous presence
validation for the :casa_case_id field. As of Rails 5, belongs_to
implies a required presence for the association key unless you specify
optional: true. I modified the model spec that was verifying the
integrity of this association.

N/A

Manual testing to verify the feature worked normally (and also to see what it looked like).

Commit content is essentially just specs, so we're good here!

### Screenshots please :)
N/A

### Feelings gif (optional)
GREEN LIGHT`

### DIFF
--- a/app/models/past_court_date.rb
+++ b/app/models/past_court_date.rb
@@ -4,7 +4,6 @@ require "sablon"
 
 class PastCourtDate < ApplicationRecord
   belongs_to :casa_case
-  validates :casa_case_id, presence: true
   validate :date_must_be_past
 
   has_many :case_court_mandates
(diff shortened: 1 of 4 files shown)


FireLemons (approved):

FireLemons:
could you run `bundle exec standardrb --fix` and then commit on this branch?

FireLemons (changes requested):

compwron (approved):
Wonderful :)

armahillo:
ah!! `standardrb`

I ran `rubocop` before pushing and it flagged a bunch of issues so I figured it wasn't being used in CI -- I'll fix this tomorrow, thank you!

PR #2744, opened 2021-10-12
fixed court document date

### LINKED ISSUE #2370: Court Document Generated with Wrong Date
### Impacted User Types
 - volunteers
 - supervisors
 - admins

### Environment
Production

### Current Behavior
When I click Generate Report, the report is generated with the wrong date.
Dated to the Next Day Generated Court Report

### Expected Behavior
When I click Generate Report, the report should be generated for the date that it is when I click the button.

### How to Replicate
1. - Log into QA as an volunteer, supervisor or admin.
2. - Click on "Generate Reports" in the left sidebar menu.
3. - Filter by a volunteer who has logged at least one case contact.
4. - Click "Generate report" at the bottom of the page.
5. - Open the generated report and look at the date on the center top of the page.

Note: This bug likely has to do with UTC versus local time so it may only occur after midnight UTC and before midnight local time.

JelenaD1:
Resolves #2370

### What changed, and why?
Due to the TimeZone being set to UTC documents were displaying the wrong date when generating after certain times. 
This fix gets the TimeZone from the client and uses that in the document generation. 

image

### DIFF
--- a/app/models/case_court_report.rb
+++ b/app/models/case_court_report.rb
@@ -25,7 +25,7 @@ class CaseCourtReport
     latest_hearing_date = @casa_case.latest_past_court_date
 
     {
-      created_date: I18n.l(Date.today, format: :full, default: nil),
+      created_date: I18n.l(Time.now.in_time_zone(@time_zone).to_date, format: :full, default: nil),
       casa_case: prepare_case_details,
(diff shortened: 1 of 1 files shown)


FireLemons:
I see the user's time zone being sent to the backend. How is the timezone used to change the date written value?

PR #2749, opened 2021-10-12
Allow admins and supervisors to impersonate volunteers

ryanmrodriguez:
### What github issue is this PR for, if any?
Resolves #1397 

### What changed, and why?
Implements pretender gem
Allow admins and supervisors to impersonate volunteers

### How will this affect user permissions?
- Volunteer permissions: Not affected
- Supervisor permissions: Can now impersonate volunteers
- Admin permissions: Can now impersonate volunteers

### How is this tested? (please write tests!) 💖💪
Added specs for sidebar and for volunteers/impersonate route

### Screenshots please :)

image
image

Issue #1397, opened 2020-11-26
Add "impersonate" for supervisor/admin to see volunteer's view


compwron:
**What type of user is this for? volunteer/supervisor/admin/all**
supervisor/admin

**Description**
Add "impersonate" for supervisor/admin to see volunteer's view

Possibly using the ruby gem `pretender`

I'm not sure that we should do this- I think it would be better to allow supervisors and admins to *view* generated court reports for a volunteer/case. (Rather than supervisor/admin being able to generate the court report themselves)

FYI @littleforest

r-sierra:
Working on it

PR #2773, opened 2021-10-14
Bugfix - Impersonation Password Update

ryanmrodriguez:
Follow-up to https://github.com/rubyforgood/casa/pull/2749
https://github.com/rubyforgood/casa/issues/1397

### What changed, and why?
Bugfix for "When logged in as an Admin/Supervisor (A) and impersonating a User (B) and than updating user B's password User A's session becomes User B."

Normally, when a password is updated, Devise logs the user out. CASA forces a sign_in after the update so that the user doesn't have to enter it manually. We just need to make sure we only sign_in if we are the true user (not impersonating).

Not applicable

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -28,7 +28,7 @@ class UsersController < ApplicationController
       return render "edit"
     end
 
-    bypass_sign_in(@user)
+    bypass_sign_in(@user) if @user == true_user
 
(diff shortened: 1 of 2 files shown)


compwron (changes requested):

stephenagreer (approved):

compwron (approved):
:)

PR #2798, opened 2021-10-18
remove login requirement for timestamp of latest deploy

FireLemons:
### What changed, and why?
Removed the login requirement to see the timestamp of the latest deploy so the bot to be coded doesn't have to log in


compwron (approved):

compwron:
Add a test plz

PR #2824, opened 2021-10-21
Adjust sidebar spec to escape special characters

mauricio-prl:


FireLemons:
so html_safe marks a string as safe for html but both user name and email are input by users so this opens the site up for an xss attack

For example 
image

Issue #2698, opened 2021-10-05
Allow Users to Save Column Display Preferences


CovenantHuman:
**What type(s) of user does this feature affect?**
 - supervisors
 - admins

**Description**
On the combined cases and volunteers page (see: https://github.com/rubyforgood/casa/issues/2694) we would like to make it possible for users to save their preferences regarding which columns to show and which to hide in the table. 

**Screenshots of current behavior, if any**  
The Current Select Displayed Columns Pop-up:
![Select Columns Popup](https://user-images.githubusercontent.com/7007647/135951317-a004c994-4208-4f16-ad45-e637d6870f5a.PNG)

_QA Login Details:_  
[Link to QA site](https://casa-qa.herokuapp.com/)  

Login Emails: 
- volunteer1@example.com  view site as a volunteer
- supervisor1@example.com view site as a supervisor
- casa_admin1@example.com view site as an admin

password for all users: 12345678

compwron:
> When we say allow users to save preferences do we mean to the database? 
Yes

> If so is there currently a pattern for this in the application or would that be entirely new behavior?
New

ludamillion:
Cool, I can take a swing at implementing this.

JuanVqz:
I can work on this if still free to pick

GALTdea:
@dpaola2 Is this a good issues to work on? @compwron can i work on this issue?

GALTdea:
Thanks @JuanVqz! I'll definitely hit you up if I run into something!

dpaola2:
@GALTdea has been working hard on this, trying to figure out how to put it into In-Progress

GALTdea:
Still working on it!

GALTdea:
Need more time

GALTdea:
Need more time

littleforest:
Completed in:

1. PR1: https://github.com/rubyforgood/casa/pull/4769
2. PR2: https://github.com/rubyforgood/casa/pull/4833
3. PR3: https://github.com/rubyforgood/casa/pull/4840 

!

PR #2847, opened 2021-10-26
Add basic preference set model and associations

ludamillion:
Component of #2698

### What changed, and why?

Adds a basic preference set model to facilitate the saving of user preferences. The associated issue requires the saving of hidden columns. 

I have only implemented the basic model so far because the page the related issue references has not been created yet.

Unit test


compwron (approved):
This looks like a good start :)

PR #2899, opened 2021-11-03
Issue 2894: Bug: No Attempt (14 Days) Not Reflecting Correct Volunteer Numbers

### LINKED ISSUE #2894: Bug: No Attempt (14 Days) Not Reflecting Correct Volunteer Numbers
### Impacted User Types
 - supervisors
 - admins

### Environment
Production

### Current Behavior
Currently the number of Volunteers reflected under No Attempt (14 Days) is not correctly reflecting the number of volunteers who have attempted to make contact in the last 14 days. 
Screen Shot 2021-11-01 at 8 26 44 PM
howard-bug

### Expected Behavior
The number of volunteers listed under No Attempt (14 Days) should reflect the number of volunteers who have not attempted contact in the last 14 days. 

### How to Replicate
1. - Log into qa as an admin or supervisor.
2. - Click on "Supervisors" in the left sidebar menu.
3. - Compare the numbers found with the contact attempts visible on the Volunteers Table.

### QA Login Details:

Login Emails: 

### Questions? Join Slack!

jesselasalle:
Resolves #2894 

### What changed, and why?
The previous query was returning any volunteer under the supervisor who had any open case assignments. I've replaced the query with some short circuited logic, and a less heavy query structure. 

N/A

This is unit tested in `/spec/models/user_spec.rb`

### Screenshots please :)
Screenshot showing correct numbers with the initial seed data. 
image

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -92,12 +92,11 @@ class User < ApplicationRecord
     no_attempt_count = 0
     volunteers.map do |volunteer|
       if volunteer.supervisor.active? &&
-         volunteer.case_assignments.any? { |assignment| assignment.active? } &&
-         ( volunteer.case_contacts.none? ||
-           volunteer.case_contacts.maximum(:occurred_at) < (Time.zone.now - 14.days)
-         )
-      then
-        no_attempt_count +=1
+          volunteer.case_assignments.any? { |assignment| assignment.active? } &&
+          (volunteer.case_contacts.none? ||
+          volunteer.case_contacts.maximum(:occurred_at) < (Time.zone.now - 14.days))
+
+        no_attempt_count += 1
       end
     end
     no_attempt_count
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -89,14 +89,18 @@ class User < ApplicationRecord
     # Get ACTIVE volunteers that have ACTIVE supervisor assignments with at least one ACTIVE case
     # 1st condition: Volunteer has not created a contact AT ALL within the past 14 days
 
-    volunteers
-      .includes(:case_assignments)
-      .joins("LEFT JOIN case_contacts cc on cc.creator_id = users.id AND cc.occurred_at > (CURRENT_DATE - INTERVAL '14 days')")
-      .group("users.id, supervisor_volunteers_users.id, case_assignments.id")
-      .where(active: true)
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -89,14 +89,17 @@ def no_attempt_for_two_weeks
...
-      .joins("LEFT JOIN case_contacts cc on cc.creator_id = users.id AND cc.occurred_at > (CURRENT_DATE - INTERVAL '14 days')")
-      .group("users.id, supervisor_volunteers_users.id, case_assignments.id")
-      .where(active: true)
-      .where(supervisor_volunteers: {is_active: true})
-      .where(case_assignments: {active: true})
-      .length
+    no_attempt_count = 0
+    volunteers.map do |volunteer|
+      if volunteer.supervisor.active? &&
+          volunteer.case_assignments.any? { |assignment| assignment.active? } &&
```
compwron, line 95: style thoughts
`volunteer.case_assignments.any? { |assignment| assignment.active? }` could be `volunteer.case_assignments.active.any?`

`(volunteer.case_contacts.none? ||
          volunteer.case_contacts.maximum(:occurred_at) < (Time.zone.now - 14.days))`
could be just `volunteer.case_contacts.maximum(:occurred_at) < (Time.zone.now - 14.days))`

We might need some `preload` or `includes` on the query to keep from doing n+1 db queries... the rubygem bullet is supposed to detect this for us but sometimes it has a hard time with tests without lots of test data

compwron (approved):
It works and I love it. Yay tests!!

app/models/user.rb
jesselasalle: `volunteer.case_assignments.active.any?`

Yeah, duh on my part hah. 

`(volunteer.case_contacts.none? || volunteer.case_contacts.maximum(:occurred_at) < (Time.zone.now - 14.days))`

My concern here is if a volunteer doesn't have any case_contact records, this throws an `undefined for NilClass` error on `<`. I was also thinking in the strictest sense, that if a volunteer has 0 `case_contact` records, then they haven't done anything in 2 weeks. If that's not the intent, then I can just switch to a safe navigator to prevent raising the exception.

Love to hear your thoughts.

compwron: Good point about the nil error. I was thinking that because it's an array it would just come back nil, but .maximum is not an array. My bad.

PR #2910, opened 2021-11-04
Added "accept" attribute to org logo field.

### LINKED ISSUE #2909: Add "accept" attribute to corporate logo file field
As a basic measure the logo field needs to use the "accept" attribute to filter the types of files accepted to jpeg, gif, and png.

(pull request forthcoming)

mdchaney:
By restricting the file field to png, gif, and jpeg we should be able to
avoid issues with problematic file types being uploaded as logos.  This
needs to be supplemented with a check on the back end, possibly using a
gem such as activestorage-validator.

Resolves #2909

### DIFF
--- a/app/views/casa_orgs/edit.html.erb
+++ b/app/views/casa_orgs/edit.html.erb
@@ -18,7 +18,7 @@
       </div>
       <div class="custom-file mb-5">
         <%= form.label :logo %>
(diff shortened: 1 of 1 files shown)


compwron (approved):
I would like to add a test for this also - could you do that in a following PR?

compwron (approved):
:)

PR #2914, opened 2021-11-05
Adds basic Bootstrap 4 styling for all check boxes.

mdchaney:
This change moves all checkbox + label pairs to the standard Bootstrap 4 layout - checkbox before label and each with "form-check-input" or "form-check-label" class.  Additionally, it is embedded in a div with "form-check" class.


compwron (approved):
:)

PR #2916, opened 2021-11-05
Standardized Bootstrap 4 html for radio buttons.

mdchaney:
This is similar to the prior update to checkbox HTML.  There was only one form affected.


compwron (approved):
nice :)

PR #2921, opened 2021-11-07
Fixes supervisor weekly digest send on Monday spec.

### LINKED ISSUE #2920: Failing test
```
Failures:

  1) SupervisorWeeklyDigest#send! on monday with active and deactivated supervisor only sends to active supervisor
     Failure/Error: expect { subject }.to change { ActionMailer::Base.deliveries.count }.by(1)
       expected `ActionMailer::Base.deliveries.count` to have changed by 1, but was changed by 0
     # ./spec/lib/tasks/supervisor_weekly_digest_spec.rb:17:in `block (5 levels) in <main>'
```

This test:

```
  describe "#send!" do
    subject { described_class.new.send! }

    context "on monday" do
      context "with active and deactivated supervisor" do
        before do
          travel_to Date.new(2021, 9, 27) # monday
          create(:supervisor, active: true)
          create(:supervisor, active: false)
        end

        it "only sends to active supervisor" do
----> expect { subject }.to change { ActionMailer::Base.deliveries.count }.by(1)
          expect(ActionMailer::Base.deliveries.last.subject).to eq("Weekly summary of volunteers' activities for the week of 2021-09-20")
        end
      end
    end
```

mdchaney:
The issue here is that the spec uses "Date.new" in a travel to, but it
doesn't take time zone into account.  They were changed to
"Time.zone.local" and moved to noon.  This will ensure that it really is
Monday in the spec.

Resolves #2920

### DIFF
--- a/spec/lib/tasks/supervisor_weekly_digest_spec.rb
+++ b/spec/lib/tasks/supervisor_weekly_digest_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe SupervisorWeeklyDigest do
     context "on monday" do
       context "with active and deactivated supervisor" do
         before do
-          travel_to Date.new(2021, 9, 27) # monday
+          travel_to Time.zone.local(2021, 9, 27, 12, 0, 0) # monday noon
           create(:supervisor, active: true)
           create(:supervisor, active: false)
         end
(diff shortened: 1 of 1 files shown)


compwron (approved):
excellent :)

PR #2922, opened 2021-11-07
Issues/2415 Add Colored Bars to Supervisor's Screen

jesselasalle:
### What github issue is this PR for, if any?
Resolves #2415

### What changed, and why?
The three columns representing total volunteers, volunteers without contact in 14 days, and transition aged youth have been replaced with a bar showing how many active and non-attempt volunteers, per the issue.

### How will this affect user permissions?
N/A

### How is this tested? (please write tests!) 💖💪
`spec/views/supervisors/index.html.erb_spec.rb`

### Screenshots please :)
image

PR #2931, opened 2021-11-11
Revert "Issues/2415 Add Colored Bars to Supervisor's Screen"

compwron:
Reverts rubyforgood/casa#2922

Reverting this now because it doesn't look quite right in QA
Screen Shot 2021-11-10 at 6 27 18 PM


compwron:
This is not needed, fixed with one space

PR #2934, opened 2021-11-11
fix css syntax error

FireLemons:


compwron:
Very nice fix, works in QA :) 

Screen Shot 2021-11-10 at 6 49 56 PM

PR #2936, opened 2021-11-11
fix weird t.string `"{:null=>false}"` bug

aedwardg:
Seems to have fixed itself when running migrations locally.
Hopefully it doesn't go back to the other way.

No official issue, but here's the issue that was opened up on paper trail for the weirdness:
https://github.com/paper-trail-gem/paper_trail/issues/1347

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -346,8 +346,7 @@ ActiveRecord::Schema.define(version: 2021_10_29_033530) do
   end
 
   create_table "versions", force: :cascade do |t|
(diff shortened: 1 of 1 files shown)


compwron (approved):

pawlik:
Hey @aedwardg we just spotted it, just in case you're curious: it's ruby-2/ruby-3 related (works fine in 2, creates this "weird" thing in ruby-3). I bet you switched to ruby-3?

PR #2964, opened 2021-11-18
wrapped reserved words in quotes

### LINKED ISSUE #2956: Add missing translations
Follow-up to https://github.com/rubyforgood/casa/pull/2924 
Part of this unquoted yes and no in the yml so we're getting `translation-missing` in some places on the site. Fix and add tests
image

FireLemons:
Resolves #2956


compwron (approved):
Nice :)
We could use some tests for what broke also

PR #3068, opened 2021-12-30
Volunteer should be able to delete court order 

compwron:
Volunteer can currently create court order and see the delete button but gets an error when pressing the delete button because they are not allowed to... Fix this somehow.


compwron:
https://github.com/rubyforgood/casa/issues/3120

Issue #2760, opened 2021-10-13
Add optional Additional Expense to case contact form


DirkCionci:
**What type(s) of user does this feature affect?**
 - volunteers

**Description**
As a Volunteer, I would like the New Case Contact form to be organized in a way that is easier for me to understand so that I can input the data that is needed for each case.

**Screenshots of current behavior, if any**  
You can paste images on the clipboard here

**How Should it Work**
- Create New Contact form should look like [mock-up] in link below
  - https://www.figma.com/proto/HBcfH4G9W2pD8psX67orhg/CASA-Redesigns?page-id=0%3A1&node-id=5%3A10&viewport=345%2C48%2C0.5&scaling=min-zoom&starting-point-node-id=5%3A10&show-proto-sidebar=1
  - Other expenses should be associated with feature toggle in System Settings

**Acceptance Criteria**
- Create New Contact form looks like mockup
- Other expenses is displayed when feature toggle is enabled and NOT displayed when it is disbaled

_QA Login Details:_  
[Link to QA site](https://casa-qa.herokuapp.com/)  

Login Emails: 
- volunteer1@example.com  view site as a volunteer
- supervisor1@example.com view site as a supervisor
- casa_admin1@example.com view site as an admin

password for all users: 123456  

### Questions? Join Slack!

We highly recommend that you join us in slack https://rubyforgood.herokuapp.com/ #casa channel to ask questions quickly and hear about office hours (currently Wednesday 6-8pm Pacific), stakeholder news, and upcoming new issues.

DrewAPeterson7671:
I am working on it

DrewAPeterson7671:
Question on this expense amount.  I need to make a column for it.  I'm thinking to use :decimal for the column type with :precision => 4 and :unit =>2.  But it occurred to me.  The end customer will likely have a maximum amount.  For validation, do we know or can we ask what the maximum expense is?  Would a single line item expense of $1,000 pop eyes out of sockets?  I should think I need to set a validation for a max amount of a line item.

compwron:
Leave out validations for now, the human judgement component will cover it 
We can ask about validations in stakeholder sync next Weds or over email before then, but not urgent.

PR #3075, opened 2022-01-05
Add feature flags and merge Additional Expenses WIP PR

compwron:
Related to #2760 but does not finish it

### What changed, and why?
Add the bones of additional expense. Add feature flagging (no library, just the db table and service- to be used from rails console for now, eventually from All Casa Admin UI)

What are feature flags? https://martinfowler.com/articles/feature-toggles.html https://en.wikipedia.org/wiki/Feature_toggle

no

feature flags are tested with rspec. Additional Expenses are not thouroughly tested yet. 

### Screenshots please :)
Screen Shot 2022-01-04 at 7 45 30 PM
Screen Shot 2022-01-04 at 7 45 27 PM


app/controllers/case_contacts_controller.rb
```diff
@@ -88,12 +88,16 @@ def edit
...
   def update
+    # binding.pry
     authorize @case_contact
     @casa_cases = [@case_contact.casa_case]
     @selected_cases = @casa_cases
     @current_organization_groups = current_organization.contact_type_groups
 
     if @case_contact.update_cleaning_contact_types(update_case_contact_params)
+      additional_expense_params.each { |single_additional_expense_params|
+        @case_contact.additional_expenses.create(single_additional_expense_params)
```
compwron, line 99: this is the cool new bug we added where it creates a new additional expense every time you try to EDIT it. Let's fix this 😂

compwron: we can make `additional_expense_params` allow `id` and then use `update_or_create_by`

DrewAPeterson7671: Ok, I see the issue and I remember we talked about it last week.  This isn't working so far, but my initial approach is something like:
```
  def create_case_contact_for_every_selected_casa_case(selected_cases)
    selected_cases.map { |casa_case|
      ActiveRecord::Base.transaction do
        case_contact = casa_case.case_contacts.create(create_case_contact_params)
        if case_contact.persisted? && additional_expense_params&.any?
          additional_expense_params.each { |single_additional_expense_params|
            additional_expense_params = AdditionalExpense.exists?(id: additional_expense_params.dig("id")) ? AdditionalExpense.where(id: additional_expense_params.dig("id")).update(single_additional_expense_params) : AdditionalExpense.create(single_additional_expense_params)
            # case_contact.additional_expenses.create(single_additional_expense_params)
          }
        end
        case_contact
      end
    }
  end
```
Please let me know if you see something that makes it a bad idea (and, of course, it isn't currently working :-) )

compwron:
```
rspec ./spec/controllers/case_contacts_controller_spec.rb:156 # CaseContactsController POST #create with valid params with additional expense creates additional expense
rspec ./spec/requests/case_contacts_spec.rb:110 # /case_contacts logged in as volunteer PATCH /update with valid parameters updates the requested case_contact and redirects to the root path
rspec ./spec/system/case_contacts/edit_spec.rb:11 # case_contacts/edit when admin admin successfully edits case contact
rspec ./spec/system/case_contacts/edit_spec.rb:45 # case_contacts/edit volunteer user is successful
```

app/controllers/case_contacts_controller.rb
compwron: Deeply nested but functional, let's try it
And let's write tests :)

DrewAPeterson7671: I'm not finding an update_or_create method.  I did find a first_or_create that can be chained to an update.  However (just FYI) I put a binding.pry in that each block and it never triggers.  It seems its not going through that block at all.  So I'm scratching my head on that and poking around with it.

compwron: @DrewAPeterson7671 the method is https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by

DrewAPeterson7671 (approved):

PR #3086, opened 2022-01-10
Fix build

compwron:


compwron:
@FireLemons When you have time, can you fix the spec in spec/views/supervisors/index.html.erb_spec.rb to match the new UI?

FireLemons:
I didn't expect that pr to get merged

PR #3097, opened 2022-01-12
Add current user info to bugsnag events

compwron:
Add current user info to bugsnag events when it is available because it is more useful than IP


compwron:
Screen Shot 2022-01-11 at 6 45 45 PM

PR #3100, opened 2022-01-12
update tests for supervisors bar graph

FireLemons:
- updated tests to match new code
 - labeled supervisor stats as count of volunteers
 - use nokogiri instead of html string comparison
 - added test for when a supervisor has no volunteers

### DIFF
--- a/spec/views/supervisors/index.html.erb_spec.rb
+++ b/spec/views/supervisors/index.html.erb_spec.rb
@@ -20,14 +20,6 @@ RSpec.describe "supervisors/index", type: :view do
       expect(rendered).to have_link("New Supervisor", href: new_supervisor_path)
     end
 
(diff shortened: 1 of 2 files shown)


spec/views/supervisors/index.html.erb_spec.rb
```diff
@@ -20,49 +20,74 @@
...
+      }
+      let!(:volunteer_without_recently_created_contacts) {
+        create(:volunteer, :with_casa_cases, supervisor: supervisor)
+      }
+
+      it "shows positive and negative numbers" do
+        assign :supervisors, [supervisor]
+        render template: "supervisors/index"
+
+        parsed_html = Nokogiri.HTML5(rendered)
```
compwron, line 36: :)

compwron (approved):
:)

PR #3107, opened 2022-01-14
test made more DRY with localization text lookup

FireLemons:
### What changed, and why?
Test checks localization files instead of hard coded string

### DIFF
--- a/spec/views/casa_cases/edit.html.erb_spec.rb
+++ b/spec/views/casa_cases/edit.html.erb_spec.rb
@@ -51,7 +51,7 @@ RSpec.describe "casa_cases/edit", type: :view do
 
       render template: "casa_cases/edit"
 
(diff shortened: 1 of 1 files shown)


compwron:
The point of a test is to test the result, not to call the code twice and assert they're the same. I am not enthused, but.. maybe? 
https://stackoverflow.com/a/6453275/307743
Damp = 'Descriptive And Meaningful Phrases' - your unit tests should be able to be 'read':
Readability is more important than avoiding redundant code.
http://blog.jayfields.com/2006/05/dry-code-damp-dsls.html
https://codeshelter.wordpress.com/2011/04/07/dry-and-damp-principles-when-developing-and-unit-testing/
https://blog.shaunfinglas.co.uk/2015/04/dry-vs-damp-in-tests.html

compwron (approved):

PR #3113, opened 2022-01-16
test looks for element id instead of text within element

FireLemons:
### What changed, and why?

Tests search for html element instead of text


compwron (approved):
:)

PR #3133, opened 2022-01-19
make hearing type a casa case factory trait

compwron:


compwron:
failing in CI
```
rspec ./spec/system/volunteers/index_spec.rb:9 # view all volunteers admin user when no logo_url can see volunteers and navigate to their cases
rspec ./spec/system/casa_cases/edit_spec.rb:416 # Edit CASA Case logged in as supervisor with an available hearing type is able to assign another hearing type to the case
rspec ./spec/system/casa_cases/edit_spec.rb:431 # Edit CASA Case logged in as supervisor with an available hearing type is able to unassign a hearing type from the case
```

PR #3168, opened 2022-01-29
Nonexistant translation

compwron:
Screen Shot 2022-01-29 at 2 47 25 PM
Screen Shot 2022-01-29 at 2 44 10 PM


FireLemons:
We could use some kind of linter for this localization stuff

PR #3194, opened 2022-02-08
Added some emojis to the different types of issues. 

seanmarcia:
This doesn't resolve anything. All it does is add some more emojis into the issue template, you know, becase we always need more emojis.


compwron (approved):
🎉

PR #3240, opened 2022-02-26
Raise error in rspec for missing translations

### LINKED ISSUE #3230: erb spec translation check
add this to every erb spec
`expect(rendered).not_to have_css("span.translation_missing")`

amygurski:
Resolves #3230

### What changed, and why?
Enables the config option in the test environment only to raise an error for missing translations. 

For example, if I add `t('.umm')` to a view (and the key `umm` doesn't exist), rspec now fails with a helpful "missing translation error":
expected_failures

Once this option was enabled, there were 132 rspec failures due to missing translations that were resolved.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

No new tests since only a rails-default config option was added.

I had to work through 132 missing translation errors. Most of these were "silent" errors. If rails can't find a translation, it'll fall back on a default translation, for example `no_court_dates` just becomes "No Court Dates" which is what was wanted anyway. Here are screenshots of the text that did "visibly" change.

### Before fixing missing translations
emancipation_checklists_before

### After fixing missing translations
emancipation_checklists_after

## New Court Case Form Change

There is also a slight styling change for the New Court Case form for the drop-down selection fields.

### Left - before fixing; Right - after fixing

new court date form - before and after

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/views/court_dates/_form.html.erb
+++ b/app/views/court_dates/_form.html.erb
@@ -12,7 +12,7 @@
       </p>
 
       <div class="field form-group">
-        <%= form.label :date, t(".add_cout_date") %>
+        <%= form.label :date, t(".add_court_date") %>
         <%= form.text_field :date,
                         value: court_date.date&.to_date || Time.zone.now,
                         data: {provide: "datepicker", date_format: "yyyy/mm/dd"},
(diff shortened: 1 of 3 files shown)


compwron (approved):
This is wonderful! :)

PR #3279, opened 2022-03-10
Resolves #3270 - Redirect better for non-edit links

### LINKED ISSUE #3270: Redirect better for non-edit links
When a supervisor or admin is logged in and visits a page like https://casa-qa.herokuapp.com/volunteers/2478 they see (screenshot)
but we would like them to instead be redirected to either https://casa-qa.herokuapp.com/volunteers/2478/edit (if they have permission to view) 
If they don't have permission to view/edit that page, then they can keep seeing the screen

This issue surprised Prince George CASA (old link back when we had view volunteer page)

Screen Shot 2022-03-08 at 5 28 38 PM

mattwd7:
Resolves #3270

### What changed, and why?
A show action was added to the volunteers_controller which redirects to the edit action in the same controller.

The same level of authorization permissions for this new action are copied from the :index action in the same controller.

A new request spec was written which confirms that the :show action redirects to the :edit action on the volunteers_controller.
The volunteer_policy spec was updated to account for the new :show action.

### Screenshots please :)
n/a

### DIFF
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -6,6 +6,11 @@ class VolunteersController < ApplicationController
     authorize Volunteer
   end
 
+  def show
+    authorize @volunteer
+    redirect_to action: :edit
+  end
+
   def datatable
     authorize Volunteer
     volunteers = policy_scope current_organization.volunteers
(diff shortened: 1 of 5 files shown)


compwron (approved):
yay! :)

PR #3282, opened 2022-03-11
Resolve #3226 - deprecate casa_case.court_date

### LINKED ISSUE #3226: deprecate casa_case.court_date and wholly use court_dates table instead
It looks like there are two places for the concept of a court_date to be persisted to the database.

The court_date field on the casa_cases table, and the court_dates table.

Where should this court_date be persisted? Should a new court_dates record be created, or should the casa_cases court_date field be populated?

mattwd7:
Resolves #3226

### What changed, and why?
CasaCase#court_date was previously replaced by a new table of court_cases. This commit removes the defunct CasaCase#court_date column.

n/a

Existing test coverage covers this change

### Screenshots please :)
n/a


db/migrate/20220310194756_remove_court_date_from_casa_cases.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveCourtDateFromCasaCases < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured { remove_column :casa_cases, :court_date }
```
compwron, line 3: This is dangerous... let me think about it for a minute :)

mattwd7: Ya.. I suppose we can leave it in in case it's used anywhere downstream and just eliminate calls to the attribute. We can also raise if a dev tries to call casa_case.court_date moving forward

compwron: Let's use https://www.bigbinary.com/blog/rails-5-adds-active-record-ignored-columns first and then do the migration deleting in a separate later PR

compwron (approved):
Very nice! :)

PR #3290, opened 2022-03-16
Remove cypress because it is flaky and disabled

compwron:
none

### What changed, and why?
Remove cypress because it is flaky and disabled, as discussed with @FireLemons in slack https://rubyforgood.slack.com/archives/CVB0QJGVD/p1647116916273879


FireLemons:
Also we have other tests that do what cypress is trying to achieve

PR #3303, opened 2022-03-19
Deprecate get_docx_subfile_contents

FireLemons:
### What changed, and why?
A while ago I made a function to test if text exists in a docx file. It's hard to use. There's another function that tries to do the same but it's incomplete. I'm replacing both with a class.

### Screenshots please :)


compwron:
Seems promising :)

spec/models/case_court_report_spec.rb
```diff
@@ -122,6 +122,8 @@
           report_top_header = get_docx_subfile_contents(report_as_raw_docx, "word/header3.xml")
           report_body = get_docx_subfile_contents(report_as_raw_docx, "word/document.xml")
 
+          DocxInspector.new(docx_path: path_to_template)
```
compwron, line 125: This doesn't assign any variables, how does that work?

compwron (approved):

PR #3336, opened 2022-04-03
Remove never text from most recent contact type dates

FireLemons:
### What changed, and why?
Removed "never" text from appearing

### Screenshots please :)
Before  
image
After  
image


compwron:
Why do you want to remove the never text?

FireLemons:
@compwron I think it clutters the page and is confusing if a volunteer has never put in a case contact

compwron:
Let's talk to the stakeholders before doing this, I'm not sure it fits what they asked for

compwron:
Let's not for now

PR #3358, opened 2022-04-10
always run bot script for debug

FireLemons:
### What changed, and why?
Remove code to prevent bot script from being run if no recent deploy occurred for debug

### DIFF
--- a/.github/workflows/after-deploy.yml
+++ b/.github/workflows/after-deploy.yml
@@ -9,7 +9,7 @@ jobs:
     name: On Deploy
     steps:
       - name: After Deploy
-        uses: Firelemons/on-deploy@v2.1.1
+        uses: Firelemons/on-deploy@v2.1.2
         with:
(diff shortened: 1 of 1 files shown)


FireLemons:
archiving lists are correct
something must be wrong with archiving requests

FireLemons:
invert boolean in if statement and return instead of wrapping all the script code for readability
https://github.com/FireLemons/on-deploy/blob/62a547e375787176117dd49afecf64729dd2a4ff/index.ts#L413

FireLemons:
this is a bug
https://github.com/FireLemons/on-deploy/blob/62a547e375787176117dd49afecf64729dd2a4ff/index.ts#L63

because this already gets the complete list of cards to be archived
https://github.com/FireLemons/on-deploy/blob/62a547e375787176117dd49afecf64729dd2a4ff/index.ts#L55

PR #3431, opened 2022-05-13
3118: Remove PaperTrail

### LINKED ISSUE #3118: Remove papertrail and related table since we are not using it

scantisani:
Resolves #3118 

### What changed, and why?
Removes the PaperTrail gem, DB tables and associated config.

The associated issue mentions that it isn't in use, but I did find one place in the `User` model where it was used to show a custom error message when a self-deactivated admin tries to log in. I removed that logic, which I hope is OK!

It shouldn't affect any permissions :)

Adds a few system specs testing error messages when inactive users log in, since that was the only real behaviour change.

### Screenshots please :)
What the 'inactive' error message will look like for everyone now, including self-deactivated admins:
image

### Feelings gif (optional)
Woman furiously typing then giving a thumbs up with Software Engineering superimposed in rainbow font


compwron (approved):

compwron:
This feels scary but it's what we want for now. We haven't used this and it has been fine for years.

PR #3518, opened 2022-05-18
Replace localization method calls with string literals in the volunteer assignment view

brodyf42:
Resolves #[3542](https://github.com/rubyforgood/casa/issues/3452) (RailsConf 14)

### What changed, and why?
Replace localization method calls with string literals in app/views/casa_cases/_volunteer_assignment.html.erb

No effect on permissions

### DIFF
--- a/app/views/casa_cases/_volunteer_assignment.html.erb
+++ b/app/views/casa_cases/_volunteer_assignment.html.erb
@@ -1,20 +1,20 @@
-<h1 class="pt-5"><%= t(".title") %></h1>
+<h1 class="pt-5">Manage Volunteers</h1>
 
 <div id="volunteer-assignment" class="card card-container">
   <div class="card-body">
     <% if @casa_case.volunteers.present? %>
       <br>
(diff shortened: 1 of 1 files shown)


app/views/casa_cases/_volunteer_assignment.html.erb
```diff
@@ -41,20 +41,20 @@
...
+                  <%= link_to "Activate Volunteer",
                               activate_volunteer_path(assignment.volunteer,
                                                       redirect_to_path: 'casa_case',
                                                       casa_case_id: @casa_case.id
                               ),
                               method: :patch,
                               class: "btn btn-outline-success" %>
                 <% else %>
-                  <%= t("common.none") %>
+                  "None"
```
brodyf42, line 57: I think this one was not supposed to have quotes around it.  However, i am having issues making an update. I can try to address this later.

brodyf42: OK, I got this updated.

compwron (approved):

PR #3543, opened 2022-05-20
Feature/1885 Give user the ability to customise which SMS notifications they want to receive

### LINKED ISSUE #1885: Give user the ability to customise which SMS notifications they want to receive
**What type of user is this for? **
Volunteer and supervisor

**Description**
User can control the notifications that they want to receive via SMS from the app.

User Acceptance criteria:

- [x] User sees events for which they can receive SMS notifications
- [x] User selects the events for which they receive SMS notifications
- [x] User can disable all SMS notifications only if email notifications are enabled

_QA Login Details:_

- volunteer1@example.com
- supervisor1@example.com
- casa_admin1@example.com

harsohailB:

Resolves #1885 

### What changed, and why?

* Created `sms_notification_events` table to store events that users can opt into for SMS notifications
    
    *  This table has a column for user type which allows users to see the events that corresponds to them

* Created `user_sms_notification_events` table to create many:many relationship between `users` table and `sms_notification_events` table
* Rendered the events below "Text Me" option of edit profile

    * Event options are disabled if "Text Me" is not checked

* Create an `sms_notification_event_populator` for seeding and prod db

    * Added an afterparty task to run this populator upon deployment 

#### Events Added

Volunteer

* CASA case youth has birthday

Supervisor

* Volunteer made case contact
* Volunteer edited case (case details, court order, court dates)

CASA Admin

* New entry in reimbursement queue

> 💡 Feel free to add more events here and I can commit them into the code!

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

* Wrote requests tests to check if `user_sms_notification_events` table is updated correctly upon user selecting an event
* Wrote view tests to check if the different user types can only see events that are meant for them

### Screenshots please :)

Screen Shot 2022-05-19 at 6 12 38 PM

### DIFF
--- a/app/javascript/src/require_communication_preference.js
+++ b/app/javascript/src/require_communication_preference.js
@@ -1,54 +1,54 @@
 /* eslint-env jquery */
 
-import Swal from 'sweetalert2';
+import Swal from 'sweetalert2'
 
-import { disableBtn, enableBtn } from './casa_case';
+import { disableBtn, enableBtn } from './casa_case'
 
-const EMAIL_TOGGLE_CLASS = 'toggle-email-notifications';
-const SMS_TOGGLE_CLASS = 'toggle-sms-notifications';
-const SAVE_BUTTON_CLASS = 'save-preference';
-const SMS_NOTIFICATION_EVENT = 'toggle-sms-notification-event';
+const EMAIL_TOGGLE_CLASS = 'toggle-email-notifications'
+const SMS_TOGGLE_CLASS = 'toggle-sms-notifications'
+const SAVE_BUTTON_CLASS = 'save-preference'
+const SMS_NOTIFICATION_EVENT = 'toggle-sms-notification-event'
 
-function displayPopUpIfPreferencesIsInvalid(receiveEmail, receiveSMS, triggerPopup = false) {
-  const emailNotificationState = $('#user_receive_email_notifications').prop('checked');
-  const smsNotificationState = $('#user_receive_sms_notifications').prop('checked');
-  receiveSMS = smsNotificationState;
-  receiveEmail = emailNotificationState;
+function displayPopUpIfPreferencesIsInvalid (receiveEmail, receiveSMS, triggerPopup = false) {
+  const emailNotificationState = $('#user_receive_email_notifications').prop('checked')
+  const smsNotificationState = $('#user_receive_sms_notifications').prop('checked')
+  receiveSMS = smsNotificationState
+  receiveEmail = emailNotificationState
 
   if (receiveSMS === false && receiveEmail === false) {
-    disableBtn($(`.${SAVE_BUTTON_CLASS}`)[0]);
+    disableBtn($(`.${SAVE_BUTTON_CLASS}`)[0])
     if (triggerPopup) {
       Swal.fire({
         icon: 'error',
         title: 'Preference Error',
-        text: 'At least one communication preference required',
-      });
+        text: 'At least one communication preference required'
+      })
     }
   } else {
-    enableBtn($(`.${SAVE_BUTTON_CLASS}`)[0]);
+    enableBtn($(`.${SAVE_BUTTON_CLASS}`)[0])
   }
 }
 
 $('document').ready(() => {
   if ($(`.${SAVE_BUTTON_CLASS}`).length > 0) {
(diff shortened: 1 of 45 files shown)


xihai01:
what's up with all those model file changes?

harsohailB:
> what's up with all those model file changes?

Just ran the migrations after pulling in `main`. Not sure why that was not committed before 🤔

7riumph (approved):
Awesome, this is great! Always increasing my RoR arsenal when reviewing your PRs. Also wanted an extra explanation on `users` and `sms_notification_events` `many:many` relationship. I read https://medium.com/@pk60905/many-to-many-relationship-in-rails-cf86e12db8b0 but want to understand more on why it helps for this specific case

app/javascript/src/require_communication_preference.js
```diff
@@ -41,8 +42,13 @@ $('document').ready(() => {
       displayPopUpIfPreferencesIsInvalid(receiveEmail, receiveSMS, true)
     })
   }
+
+  const smsToggle = document.getElementById(SMS_TOGGLE_CLASS)
```
7riumph, line 46: `document.getElementById(CLASS)` throws null errors attempting to find the element on any given site page (inspect element console error). `$(.${CLASS})[0]` should fix this while also avoiding weird occurrences like allowing an array of values to pass.

app/views/users/edit.html.erb
```diff
@@ -79,9 +79,18 @@ minlength: User.password_length.min %>
      <%= form.label :receive_email_notifications, t("communication_preferences.email_me") %>
    </div>
    <div>
-     <%= form.check_box :receive_sms_notifications, class: "toggle-sms-notifications" %>
+     <%= form.check_box :receive_sms_notifications, id: "toggle-sms-notifications", class: "toggle-sms-notifications" %>
```
7riumph, line 82: Some things are dependent on the automatically generated id  `user_receive_sms_notifications`. Don't think it's best to override this.

app/views/users/edit.html.erb
```diff
@@ -79,9 +79,18 @@ minlength: User.password_length.min %>
...
      <%= form.label :receive_sms_notifications, t("communication_preferences.text_me") %>
    <div>
+   <div class="pl-4 pb-4">
+      <%= form.collection_check_boxes("sms_notification_event_ids", SmsNotificationEvent.where(user_type: @user.type),
+:id, :name) do |event| %>
+        <div class="form-check">
+          <%= event.check_box(class: "form-check-input", id: "toggle-sms-notification-event") %>
```
7riumph, line 89: This is looking ahead but as the number of notification events grow adding a drop-down/collapse menu could be good at some point. Probably in the same vein as password confirms dropdown. Checkbox is ticked → menus drops down with event checkboxes. checkbox is unticked → menu collapses

7riumph:
Events aren't displaying on the edit profile page on my end even after ticking "Text Me". Still unsure why. I'm on the clone branch, `7riumph:harsohail_sms_event` for reference. Is anyone else getting this? Only ran `rake db:migrate`

harsohailB:
> Events aren't displaying on the edit profile page on my end even after ticking "Text Me". Still unsure why. I'm on the clone branch, `7riumph:harsohail_sms_event` for reference. Is anyone else getting this? Only ran `rake db:migrate`

I think you will have to re-seed the database as well to populate the events data.

app/models/supervisor_volunteer.rb
```diff
@@ -24,8 +24,8 @@ def ensure_supervisor_and_volunteer_belong_to_same_casa_org
 #  is_active     :boolean          default(TRUE)
 #  created_at    :datetime         not null
 #  updated_at    :datetime         not null
-#  supervisor_id :integer          not null
-#  volunteer_id  :integer          not null
+#  supervisor_id :bigint           not null
```
xihai01, line 27: so confused why they changed from int to bigint.

compwron: I think it's a local version difference with postgres maybe - bigint is arguably more correct.

compwron (approved):

app/views/users/edit.html.erb
xihai01: good idea

xihai01 (approved):
late to the party, but great stuff!

PR #3380, opened 2022-04-18
#1790 - SMS opt-in to system imports with mobile numbers

harsohailB:
### What github issue is this PR for, if any?
Resolves #1790

### What changed, and why?

- [x] Write function to detect mobile numbers exist
- [x] Create opt-in modal
- [x] Pause import if opt-in required and show modal
- [x] Temporarily store CSV in local storage
- [x] Delete CSV from local storage upon successful import
- [x] Resume import upon confirming opt-in 
- [x] Turn on SMS notifications for imported users

### How will this affect user permissions?
- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

### How is this tested? (please write tests!) 💖💪

- [x] Update import request tests
- [x] Update volunteer and supervisor importer tests
- [x] Add system tests for opt in modal

### Screenshots please :)

image

PR #3550, opened 2022-05-22
fix import js to only run when import elements are present

### LINKED ISSUE #3547: Import javascript has null values on non-import pages
**Description**

* As import elements do not exist on non-import pages, adding an event listener through javascript results in an undefined error to be thrown in the console

Screen Shot 2022-05-21 at 1 02 23 PM

**Existing gem(s), file(s) needing updating, changing or deleting**

N/A

**New file(s) needing creating**

`app/javascript/src/import.js`

* Add condition to only run the respective code when the elements exist

harsohailB:

Resolves #3547 

### What changed, and why?

* Changed `import.js` to only add event listeners when on import page to avoid undefined errors in console on other pages

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

* Tests for `import.js` were already written in #3380 

### Screenshots please :)

N/A

### DIFF
--- a/app/javascript/src/import.js
+++ b/app/javascript/src/import.js
@@ -55,17 +55,21 @@ function populateFileInput (inputId) {
 $('document').ready(() => {
   ['volunteer', 'supervisor'].forEach((importType) => {
     const inputFileElementId = `${importType}-file`
+    const inputFileElement = $(`#${inputFileElementId}`)[0]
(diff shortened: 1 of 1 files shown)


7riumph (approved):
Awesome! 🔥

xihai01 (approved):
lgtm

compwron (approved):
Nice :)

PR #3664, opened 2022-06-21
Fixing case contacts date suggestion (also for the entire system)

### LINKED ISSUE #3174: Case Contacts Suggest Tomorrow as Date of Contact After Midnight GMT
### Impacted User Types
 - volunteers
 - supervisors
 - admins

### Environment
Production, staging, QA, probably local? unconfirmed on local yet

### Current Behavior
When I try to make a case contact for the current day after 5pm Pacific Time, the default date offered is for the next day and I can save the case contact as having occurred the next day. See the difference between my system date and time and the date of the most recently saved case contact in the provided screenshot below.
Datetime Error on Case Contact Creation

### Expected Behavior
It should not be possible to make a case contact for a date in the future, nor should the suggested date on the case contact form be in the future. 

### How to Replicate
1. - Log into qa as a volunteer after midnight GMT.
2. - Select a Case by clicking on its Case Number.
3. - Press the New Case Contact Button at the top of the page.
4. - Check the date that the Case Contact Form is suggesting.

### QA Login Details:

Login Emails: 

### Questions? Join Slack!

arthur1041:
Resolves #3174

### What changed, and why?
This bug was happening because of different time zones. The default time zone in Rails is UTC. When 
the user tried registering a new case contact after 5 pm in the Pacific time zone, it was already the next day in UTC. The change I did was to change Rails timezone to the user's system timezone, now the system will adjust to your timezone no matter in what part of the world you are.

- Volunteer permissions: No impacts
- Supervisor permissions: No impacts
- Admin permissions: No impacts

Just run RSpec. As my modification was at the settings level, no new tests were written, but some of the existing tests were adjusted.

### Feelings gif (optional)
How to embed:
thinking a lot


arthur1041:
I'm working on a fix for the RSpec failure.

@compwron let me know if you have a better approach to fix this bug. The one that I did affects the entire system but is the most correct solution I can think of.

compwron:
> The change I did was to change Rails timezone to the user's system timezone,
this is not ok because we have different users in different timezones. Server and database timezones should always always be in UTC. The correct solution is to translate in the view layer to the user's timezone

config/application.rb
```diff
@@ -25,5 +25,6 @@ class Application < Rails::Application
     config.active_storage.variant_processor = :mini_magick
     config.active_storage.content_types_to_serve_as_binary.delete("image/svg+xml")
     config.serve_static_assets = true
+    config.time_zone = ActiveSupport::TimeZone[Time.now.strftime("%z").delete("0").to_i]
```
compwron, line 28: Given that it will be running on heroku servers which are themselves in UTC, I do not expect this to solve our problem. But also, let's not override server timezone.

arthur1041:
> > The change I did was to change Rails timezone to the user's system timezone,
> > this is not ok because we have different users in different timezones. Server and database timezones should always always be in UTC. The correct solution is to translate in the view layer to the user's timezone

Okay, got it. I'll apply this solution you suggested.

PR #3703, opened 2022-06-28
Added policy scope for hearing types

Learningstuff98:
### What changed, and why?

Did a small code refactor.

n/a

I ran tests and tested it in localhost.

### DIFF
--- a/app/controllers/checklist_items_controller.rb
+++ b/app/controllers/checklist_items_controller.rb
@@ -48,9 +48,9 @@ class ChecklistItemsController < ApplicationController
   def authorize_checklist_item
     authorize ChecklistItem
   end
-
+ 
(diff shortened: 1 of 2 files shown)


app/controllers/checklist_items_controller.rb
```diff
@@ -50,7 +50,7 @@ def authorize_checklist_item
   end
 
   def set_hearing_type
-    @hearing_type ||= HearingType.find(params[:hearing_type_id])
+    @hearing_type ||= policy_scope(HearingType).find(params[:hearing_type_id])
```
compwron, line 53: :)

compwron: I would also like to see a test for this

PR #3727, opened 2022-06-29
Added test for hearing type policy scope

Learningstuff98:
### What changed, and why?

Added a test for the hearing type policy scope to increase test coverage.

n/a

Made sure that the test passes and fails when its supposed to.


compwron (approved):

compwron:
yay!

PR #3732, opened 2022-07-01
Fix flaky data report spec

marmitoTH:
### What changed, and why?
I made this test, but it was causing the test suite to fail sometimes.

Just changed it a little bit so the order of the output doesn't matter anymore.


compwron (approved):
:)

PR #3735, opened 2022-07-02
#1796: "try again"! reminder for volunteers after no contact made

harsohailB:
### What github issue is this PR for, if any?

Resolves #1796 

### What changed, and why?

* Updated `user_case_contact_types_reminder` table to be `user_reminder_time` where each column now represents time a reminder was last sent (simpler this way)
* Created service and task for sending out no contact made reminders (tasks runs daily)
* Abstracted out `sms_reminder_service` to be used by other reminder services such as this `no_contact_reminder` and previously created `court_date_report_due_reminder`

### How will this affect user permissions?
- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

### How is this tested? (please write tests!) 💖💪

* Wrote tests for model, service, and task added

### Screenshots please :)

N/A

PR #3757, opened 2022-07-16
Feature 1792 - users can choose to recieve password resets via SMS

### LINKED ISSUE #1792: users can choose to receive password reset via SMS
Part of epic #1017, dependent on #1785, #1790, #1791, #1780, and #1782

AS a **volunteer/supervisor/admin**,
I WANT the option to receive a password reset link via SMS,
SO THAT I am not restricted by receiving it only via email.

### How it works now
Right now, all users only have the option to receive a password reset link by email.

All users should have the option to receive a password reset link via SMS. 

### Acceptance criteria
- [ ] on the "forgot password" page, there should be a field for mobile number
- [ ] the field should only accept numbers
- [ ] field should only accept mobile numbers that are the appropriate length of a US mobile number (ten digits without +1 in the front, 11 digits if 1 is included as the first digit)
- [ ] the text "or" should separate the email and mobile number fields so it is obvious the user only needs to enter **one**
- [ ] the phone number entered must already be associated with an existing user account
- [ ] clicking "send me reset password instructions" after entering an existing mobile number triggers an SMS notification containing a reset password link
- [ ] that SMS should read something like:
_Click here to reset your password on casavolunteetracking.org [link]_
- [ ] clicking "send me password instructions" after entering a mobile number overrides previous opt-out preferences (so that the user can receive the one-time notification they are requesting)

xihai01:
Resolves #1792 

### What changed, and why?
CASA users can now send password resets via SMS in addition to email.
Users can have the flexible options of getting a reset link via:
- Just email
- Just phone number
- Both email and phone number

To achieve this:
- A phone number field is added in the front end
- A passwords controller with a create action is implemented to override Devise's passwords controller. This is needed to add in the functionality of sending via SMS.

- A view spec is created to test front end aspect (i.e additional text and phone number field is present on page)
- A system spec is created to test user flow (forgot password -> enter email and/or phone number -> click send -> redirect with a notice) and rendering of any form validation errors
- A request spec is created to test our custom route that it calls short io and twilio services
 and redirects if no form errors occur

### Screenshots please :)
demo

#### SMS received via my phone (short link was visible thanks to @7riumph 's branded domain)
SMS

### Feelings gif (optional)
excitment
I really enjoyed this issue - it allowed me to dive deeper into the internals of Devise

### DIFF
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -1,0 +1,23 @@
+class Users::PasswordsController < Devise::PasswordsController
+  include ApplicationHelper
+  include PhoneNumberHelper
+
+  def create
+    email = params[resource_name][:email]
+    phone_number = params[resource_name][:phone_number]
+    # try to find user by email
+    if !User.find_by(email: email)
+      resource.errors.add(:base, "Email not found")
+    end
+    # validate and add any errors
+    validation = valid_phone_number(phone_number)
+    if validation[0]
+      User.find_by(phone_number: phone_number) ? "" : resource.errors.add(:base, "Phone number not found")
+    else
+      resource.errors.add(:phone_number, validation[1])
+    end
+    # re-render and display errors
+    respond_with(resource)
+    # otherwise, send reset email and sms
+  end
+end
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -7,33 +7,18 @@ class Users::PasswordsController < Devise::PasswordsController
     email = params[resource_name][:email]
     phone_number = params[resource_name][:phone_number]
     reset_token = nil
-
-    if email.blank? && phone_number.blank?
-      resource.errors.add(:base, "Please enter at least one field.")
-    end
-
-    if !email.blank? && !User.find_by(email: email)
-      resource.errors.add(:base, "Email not found")
-    end
-
-    validation = valid_phone_number(phone_number)
-    if validation[0]
-      User.find_by(phone_number: phone_number) ? "" : resource.errors.add(:base, "Phone number not found")
-    else
-      resource.errors.add(:phone_number, validation[1])
-    end
+    @resource = email.blank? ? User.find_by(phone_number: phone_number) : User.find_by(email: email)
 
     # re-render and display any errors
-    if resource.errors.any?
-      respond_with(resource)
+    params_is_valid, error_resource = password_params_is_valid(resource, email, phone_number)
+    if !params_is_valid
+      respond_with(error_resource)
       return
     end
 
-    # otherwise, send reset email and sms
-    @resource = email.blank? ? User.find_by(phone_number: phone_number) : User.find_by(email: email)
+    # generate a reset token and
+    # call devise mailer
     if !email.blank?
-      # generate a reset token and
-      # call devise mailer
       reset_token = @resource.send_reset_password_instructions
     end
 
@@ -58,5 +43,28 @@ class Users::PasswordsController < Devise::PasswordsController
     end
 
     redirect_to after_sending_reset_password_instructions_path_for(resource_name), notice: "You will receive an email or SMS with instructions on how to reset your password in a few minutes."
+  end
+
+  private
+
+  def password_params_is_valid(resource, email, phone_number)
+    if email.blank? && phone_number.blank? && resource
+      resource.errors.add(:base, "Please enter at least one field.")
+      return [false, resource]
+    end
+
+    phone_number_is_valid, error_message = valid_phone_number(phone_number)
+    if !phone_number_is_valid && resource
+      resource.errors.add(:phone_number, error_message)
+      return [false, resource]
+    end
+
+    if resource.email != email || resource.phone_number != phone_number
+      # A new, empty resource is returned (see application helper)
+      # so to check for nil, we need to check its email/phone fields
+      resource.errors.add(:base, "User does not exist.")
+      return [false, resource]
+    end
+    return [true, nil]
   end
 end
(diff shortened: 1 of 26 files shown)


spec/system/devise/passwords/new_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe "users/passwords/new", type: :system do
+  before do
+    visit root_path
+    click_on "Forgot your password?"
+  end
+
+  it "displays text above form fields" do
+    expect(page).to have_text("Please enter email or phone number to recieve reset instructions.")
```
xihai01, line 10: TO DO: move this test to view spec

xihai01:
@compwron @harsohailB @7riumph rspec is failing because the db is missing a "user reminder times" table and the lib/tasks tests depend on it. Looking at the schema, it seems to me someone refactored the "users reminder times" to "user case contact types reminders". Any clue to what's going on here?
Screenshot from 2022-07-16 11-18-39
Screenshot from 2022-07-16 11-25-42

xihai01:
I'll work on fixing the code climate issues. Please don't merge yet or until everything is green. :)

harsohailB:
> it seems to me someone refactored the "users reminder times" to "user case contact types reminders"

@xihai01 It should be the opposite. I refactored "user case contact types reminders" table to "users reminder times" to make it re-usable for all reminder times. #3735

app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    if email.blank? && phone_number.blank?
+      resource.errors.add(:base, "Please enter at least one field.")
+    end
+
+    if !email.blank? && !User.find_by(email: email)
+      resource.errors.add(:base, "Email not found")
+    end
+
+    validation = valid_phone_number(phone_number)
+    if validation[0]
```
harsohailB, line 20: Can we do something like this instead so its more readable than indexing:
```ruby
phone_number_is_valid, error_message = valid_phone_number(phone_number)

if phone_number_is_valid
    error_message = "Phone number not found" if User.find_by(phone_number: phone_number)
end

if error_message
    resource.errors.add(:phone_number, error_message)
end
```

app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    # re-render and display any errors
+    if resource.errors.any?
+      respond_with(resource)
+      return
+    end
+
+    # otherwise, send reset email and sms
+    @resource = email.blank? ? User.find_by(phone_number: phone_number) : User.find_by(email: email)
```
harsohailB, line 33: Seems like you are already querying `User.find_by(phone_number: phone_number)` and `User.find_by(email: email)` above in the file to add error if the queries doesn't find anything. Maybe put these in a variable so you only call them once and reuse them if the query does find something?

app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      sms_params = {
+        From: @resource.casa_org.twilio_phone_number,
+        Body: body_msg,
+        To: phone_number
+      }
+      twilio_service.send_sms(sms_params)
+    end
+
+    redirect_to after_sending_reset_password_instructions_path_for(resource_name), notice: "You will receive an email or SMS with instructions on how to reset your password in a few minutes."
```
harsohailB, line 60: Just a question: What happens if they put in both email and phone number? Do they get notified through email and sms?

app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+  include ApplicationHelper
+  include PhoneNumberHelper
+  include SmsBodyHelper
+
+  def create
+    email = params[resource_name][:email]
+    phone_number = params[resource_name][:phone_number]
+    reset_token = nil
+
+    if email.blank? && phone_number.blank?
```
harsohailB, line 11: I think you should have 2 clear paths for email and sms based on what is filled out. Right now, its hard to follow the logic as we have a lot `if` statements that keep checking `email.blank?` and `phone_number.blank?`. Maybe:

```ruby
if !email.blank? && valid_email
   # run this code which  only has to do with the email
end

if !phone_number.blank? && valid_phone_number
   # run this code which  only has to do with the sms
end
```

harsohailB (commented):
Really nice job! Glad you enjoyed this because I was having a really hard time wrapping my head around `Devise`. Just a few comments about code separation and reusability in the passwords controller.

app/controllers/users/passwords_controller.rb
xihai01: yes, that is intended. If user supplies both email and phone number, they get both an email and SMS

xihai01:
TO DO STILL: 
- create action is apparently still too complex for a human to understand -> need refactoring
- migrate request spec to a controller spec to pass that spec checker :hankey: 
- fix error with failing lib/tasks tests

compwron:
This is exciting!
CI is failing on lint and spec

compwron (approved):

app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,71 @@
+class Users::PasswordsController < Devise::PasswordsController
```
compwron, line 1: eek

app/helpers/application_helper.rb
```diff
@@ -44,4 +44,16 @@ def flash_class(level)
   def og_tag(type, options = {})
     tag.meta(property: "og:#{type}", **options)
   end
+
+  def resource_name
```
compwron, line 48: for devise ?

compwron (approved):

7riumph (commented):
🔥🔥🔥

app/helpers/application_helper.rb
xihai01: yes, for devise. The devise view calls some internal methods like resource and resource_name. So we need to map these to use them in our controllers and specs

PR #3784, opened 2022-07-27
See what breaks when removing il8n translations

compwron:


compwron:
That is a lot of failing tests :/

PR #3792, opened 2022-07-29
Address flaky system specs by disabling Rack::Attack in the test environment

kcdragon:
### What changed, and why?

This commit disables Rack::Attack in the test environment except for a single spec that needs it.

The problem we noticed was that system specs occasionally fail in the CI and the resulting screenshot at the failure is a blank page that just says "Retry later". This seems similar to an issue posted at https://github.com/thoughtbot/capybara-webkit/issues/628 and the fix for that issue was to disable Rack::Attack in the test environment.

It's difficult to be certain that this will fix the specs since we aren't able to reproduce locally but this seems like a reasonable thing to try.

See more conversation in the Ruby for Good Slack https://rubyforgood.slack.com/archives/C04PM108S/p1658887094904429

- Volunteer permissions: No changes
- Supervisor permissions: No changes
- Admin permissions: No changes

No new tests since its a change to existing test

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/spec/config/initializers/rack_attack_spec.rb
+++ b/spec/config/initializers/rack_attack_spec.rb
@@ -9,6 +9,7 @@ RSpec.describe Rack::Attack do
   let(:cache) { Rails.cache }
 
   before do
+    Rack::Attack.enabled = true
     ActionController::Base.perform_caching = true
     allow(Rails).to receive(:cache).and_return(memory_store)
     Rails.cache.clear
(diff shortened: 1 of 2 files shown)


spec/config/initializers/rack_attack_spec.rb
```diff
@@ -9,6 +9,7 @@
   let(:cache) { Rails.cache }
 
   before do
+    Rack::Attack.enabled = true
```
kcdragon, line 12: I confirmed locally that these tests do fail when rack attack is disabled (which is what we'd expect) so I added this to enable them.

compwron (approved):
very cool!

PR #3793, opened 2022-07-30
add .env gem in project

italomatos:
### What changed, and why?
This PR has goal to allow contributors use .env gem to setup environment in local machine.

- run ``` bin/setup ```
- edit .env file that was created with database config
- run project


compwron (approved):
this does look like an improvement! Thanks :)

PR #3795, opened 2022-07-31
Add controller tests for all_casa_admins/casa_admins_controller.rb

### LINKED ISSUE #3672: Add test for controllers/all_casa_admins/casa_admins_controller.rb
Thank you for working on CASA!

To complete this issue:
1. open `controllers/all_casa_admins/casa_admins_controller.rb`
1. make a new test file: controllers/all_casa_admins/casa_admins_controller_spec.rb
1. add at least one test for the functionality in controllers/all_casa_admins/casa_admins_controller.rb!
1. remove the line controllers/all_casa_admins/casa_admins_controller.rb from `.allow_skipping_tests`

This will improve our test coverage and make our code safer to modify and easier to understand.

Example:

```
Before:
require "rails_helper"

RSpec.describe CasaAdminsController do
  it "adds the numbers" do
    expect(described_class.new(1, 1)).to eq(2)
  end
end
```

### Questions? Join Slack!

Learningstuff98:
Resolves #3672

### What changed, and why?

I added controller tests to increase test coverage.

n/a

I made sure that the tests pass and fail when they need to.


spec/controllers/all_casa_admins/casa_admins_controller_spec.rb
```diff
@@ -0,0 +1,170 @@
+require "rails_helper"
+
+RSpec.describe AllCasaAdmins::CasaAdminsController, type: :controller do
+  let(:all_casa_admin) { build(:all_casa_admin) }
+  let(:casa_org) { create(:casa_org) }
+
+  before do
+    all_casa_admin.update_attribute(:id, 5)
```
Learningstuff98, line 8: The id of the all_casa_admin instances are nil, so I had to give them a number value so the sign in process would work.

compwron: This seems very weird, that shouldn't happen...

spec/controllers/all_casa_admins/casa_admins_controller_spec.rb
```diff
@@ -0,0 +1,170 @@
+require "rails_helper"
+
+RSpec.describe AllCasaAdmins::CasaAdminsController, type: :controller do
+  let(:all_casa_admin) { build(:all_casa_admin) }
```
compwron, line 4: the id is nil because you are using build not create. It would be a nicer style to use create and not manually set the id.

Learningstuff98: Changing it to create fixed the issue. Thanks for pointing this out.

compwron (approved):
yay tests!

PR #3799, opened 2022-08-01
replace class with method in class << self to modules

italomatos:
### What changed, and why?
Replace class with only class methods with module ``` extend self ```

"Reason is that Class is too heavy for creating only class methods. As we discussed earlier a class is supposed to have both data and behavior. If the only thing you care about is behavior then ruby suggests to implement it as a module."

more details: [https://www.bigbinary.com/blog/extend-self-in-ruby](https://www.bigbinary.com/blog/extend-self-in-ruby)


compwron (approved):
:)

PR #3830, opened 2022-08-11
Updated layouts/footer.html.erb_spec.rb

7riumph:
### What changed, and why?
The "Ruby For Good" and "SMS Terms & Conditions" `footer links` now have tests to ensure...
- they redirect to the expected webpage  
- the links text displays accurately

- Volunteer permissions: n/a
- Supervisor permissions: n/a 
- Admin permissions: n/a


compwron (approved):
very nice :)

PR #3838, opened 2022-08-13
fix webmock version

italomatos:
### What changed, and why?
This PR has the goal to fix web mock version. Because current version can no longer be found in that source

link: 
```
Fetching gem metadata from https://rubygems.org/.........
[215](https://github.com/rubyforgood/casa/runs/7778570716?check_suite_focus=true#step:4:216)
Your bundle is locked to webmock (3.16.0) from rubygems repository
[216](https://github.com/rubyforgood/casa/runs/7778570716?check_suite_focus=true#step:4:217)
https://rubygems.org/ or installed locally, but that version can no longer be
```

image

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -461,7 +461,7 @@ GEM
       nokogiri (~> 1.6)
       rubyzip (>= 1.3.0)
       selenium-webdriver (~> 4.0)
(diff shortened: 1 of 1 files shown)


compwron:
thanks! :)

PR #3849, opened 2022-08-17
Issue #1865: Copy Court Orders from One Case into Another

### LINKED ISSUE #1865: Copy Court Orders from One Case into Another
volunteer, supervisor, admin

**Description**

Currently we have no way to copy court orders that have been entered once to sibling cases that they also apply to. We would like to add a way to do that to the Edit Case Details page. We want this feature to add court orders from a selected case to another existing case's court orders, not overwrite court orders from one case with court orders from another. Under the current "Court Orders - Please check that you didn't enter any youth names" text we would like to add text that says "Copy all orders from case:" and then a dropdown with all the cases the current user has access to. (For a volunteer this would be their own cases, for a supervisor or admin this would be all cases within a given CASA. As noted on the image below this new section of the page should only be visible to volunteers with more than one case, supervisors, and admins.) Next to this dropdown we would like to add a submit button that only impacts this new section of the form. This submit button would add all the court orders from the case selected in the dropdown to the case being edited (leaving any existing court orders on the case being edited as they were). 

copy-court-orders

_QA Login Details:_

- volunteer1@example.com
- supervisor1@example.com
- casa_admin1@example.com

mussajoop:
Resolves #1865 

### What changed, and why?
We added a way for volunteers, supervisors, and admins to copy court orders from one case to another, to be able to copy court orders that have been entered once to sibling cases that they also apply to.

- Volunteer permissions: n/a
- Supervisor permissions: n/a
- Admin permissions: n/a

Added some system tests to test some use cases like element visibility, court orders copy, sibling cases ...
Tests are based on volunteers and Admin (Since admin and Supervisor are expected to have the same behaviour for this case) :)
### Screenshots please :)
image

image


app/controllers/casa_cases_controller.rb
```diff
@@ -35,6 +35,11 @@ def new
   end
 
   def edit
+    if current_role == "Volunteer"
+      @siblings_casa_cases = current_user.casa_cases.excluding(@casa_case)
+    elsif ["Supervisor", "Casa Admin"].include? current_role
```
compwron, line 40: ideally we would find a way to use our pundit policies for permissions work like this

app/controllers/casa_cases_controller.rb
```diff
@@ -124,6 +129,15 @@ def reactivate
     end
   end
 
+  def copy_court_orders
+    authorize @casa_case, :update_court_orders?
+    CasaCase.find_by_case_number(params[:case_number_cp]).case_court_orders.each do |court_order|
+      dup_court_order = court_order.dup
+      dup_court_order.save
```
compwron, line 136: save can return true or false
this should probably be wrapped in a transaction

app/views/shared/_court_order_list.erb
```diff
@@ -1,5 +1,14 @@
-<%= form.label :case_court_orders, t("global.court_orders") %>
+<%= form.label :case_court_orders, "Court Orders - Please check that you didn't enter any youth names" %>
```
compwron, line 1: nice inlining :)

compwron (approved):
Yay tests!

app/controllers/casa_cases_controller.rb
mussajoop: absolutely you're right. I'll commit the changes :)

app/controllers/casa_cases_controller.rb
```diff
@@ -35,6 +35,11 @@ def new
   end
 
   def edit
+    if current_role == "Volunteer"
```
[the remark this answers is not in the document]
mussajoop: will do thanks for the review! :)

app/controllers/casa_cases_controller.rb
mussajoop: Yeah it's surely not the best way to do :)
I'm still looking for a way to do it with pundit

app/views/shared/_court_order_list.erb
mussajoop: thanks 👍🏾

compwron (approved):
nice!

PR #3881, opened 2022-09-01
System tests for checklist item functionality

Learningstuff98:
### What changed, and why?

I wrote some tests to improve test coverage.

n/a

I made sure that the tests pass and fail when they need to.

### DIFF
--- a/spec/system/checklist_items/destroy_spec.rb
+++ b/spec/system/checklist_items/destroy_spec.rb
@@ -2,11 +2,10 @@ require "rails_helper"
 
 RSpec.describe "checklist_items/destroy", type: :system do
   let(:casa_admin) { create(:casa_admin) }
-  let(:hearing_type) { create(:hearing_type) }
   let(:checklist_item) { create(:checklist_item) }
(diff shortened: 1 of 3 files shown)


spec/system/checklist_items/destroy_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe "checklist_items/destroy", type: :system do
+  let(:casa_admin) { create(:casa_admin) }
+  let(:hearing_type) { create(:hearing_type) }
+  let(:checklist_item) { create(:checklist_item) }
+
+  before do
+    hearing_type.checklist_items.push(checklist_item)
```
Learningstuff98, line 9: I tried to set the checklist item's hearing type the same way that the edit spec is set up, but it didn't work. This is what I came up with. If there's a better way, I'm open to ideas, because this smells to me.

Learningstuff98: I figured out a better way.

FireLemons:
:1st_place_medal:

Issue #2942, opened 2021-11-15
fix ignored spec "can see Add to Calendar buttons" - this depends on https://github.com/rubyforgood/casa/issues/2889


compwron:
fix ignored spec "can see Add to Calendar buttons" - this depends on https://github.com/rubyforgood/casa/issues/2889

### Questions? Join Slack!

We highly recommend that you join us in slack https://rubyforgood.herokuapp.com/ #casa channel to ask questions quickly and hear about office hours (currently Wednesday 6-8pm Pacific), stakeholder news, and upcoming new issues.

littleforest:
Fixed in #4025

PR #4025, opened 2022-10-05
update failing spec

metamoni:
Fixes an ignored spec in the file mentioned in #2942. While the issue doesn't explicitly mention this spec, I was looking through the file and noticed we had one ignored spec that was failing.

### What changed, and why?
Changed the mocked date. For some reason, dates like `1.year.from_now` or `2.months.from_now` get formatted with an extra space, causing the spec to fail. The date rendered on the page appears to be formatted properly, without the extra space.

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

- updated `next_court_date` for `casa_case`


compwron (approved):
cool, thanks! :)

PR #4043, opened 2022-10-09
remove translations in views

### LINKED ISSUE #4042: Remove remaining translations - part 2 (erb files)
## Description
This is a follow-up to #3930, #3930, #3957 and #4040. There are a few more translations in the codebase that need to be removed. This issue should cover all translations in `erb` files (views)

metamoni:
Resolves #4042 

### What changed, and why?
Translations have been replaced with strings

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

Should be covered by existing tests

### DIFF
--- a/app/views/volunteers/_notes.html.erb
+++ b/app/views/volunteers/_notes.html.erb
@@ -19,7 +19,7 @@
               <tr>
                 <td><%= note.content %></td>
                 <td><%= note.creator.display_name %></td>
-                <td><%= note.created_at.strftime("%m-%d-%Y") %></td>
+                <td><%= l(note.created_at, format: :standard) %></td>
                 <td>
                   <%= link_to "Edit", edit_volunteer_note_path(@volunteer, note), class: "btn btn-primary" %>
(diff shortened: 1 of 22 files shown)


app/views/volunteers/_notes.html.erb
```diff
@@ -19,7 +19,7 @@
               <tr>
                 <td><%= note.content %></td>
                 <td><%= note.creator.display_name %></td>
-                <td><%= note.created_at.strftime(t("time.formats.standard")) %></td>
+                <td><%= note.created_at.strftime("%m-%d-%Y") %></td>
```
littleforest, line 22: Could we actually take advantage of the translation file for this? That way if we change the way we want to display the date, we only have to change it in one place:
```suggestion
                <td><%= l(note.created_at, format: :standard) %></td>
```

Ref: https://guides.rubyonrails.org/i18n.html#adding-date-time-formats

littleforest (commented):
Looks great! Just one minor request to use the locale file to display the date, and to modify it to use the `locale` syntax.

compwron:
CI is failing on erb lint

PR #4049, opened 2022-10-10
remove all uses of transition_aged_youth DB field

### LINKED ISSUE #2145: refactor: remove transition_aged_youth and only use casa_case.birth_month_year_youth
refactor: remove all usages of `casa_case.transition_aged_youth` and only use `casa_case.birth_month_year_youth` and also make a migration to remove `casa_case.transition_aged_youth`. Make sure to use "Rails Ignored Columns".

We want to do this because having two different flags for this is having extra state which is unnecessary. 

Also remove this rake task
and the heroku job that runs it in qa, staging, and prod. 
```
desc "Transition youth who are now 14, run by heroku scheduler"
task transition_youth: :environment do
  puts "Updating casa cases..."
  CasaCase.should_transition.update_all(transition_aged_youth: true)
  puts "done."
end
```

librod89:
Addresses #2145 

This DB column can be completely removed in a future PR.

### What changed, and why?
- Add `transition_aged_youth` to list of `ignored_columns` in CasaCase model
- Remove `transition_aged_youth` references where necessary **while leaving references to this field on the frontend for clarity**

N/A

Updated existing tests

### Screenshots please :)
N/A

### Feelings gif (optional)
How to embed:
`alt text`


app/models/user.rb
```diff
@@ -92,7 +92,7 @@ def volunteers_serving_transition_aged_youth
     volunteers.includes(
       case_assignments: :casa_case
     ).where(
-      case_assignments: {active: true}, casa_cases: {active: true, transition_aged_youth: true}
+      case_assignments: {active: true}, casa_cases: {active: true, birth_month_year_youth: ..14.years.ago}
```
littleforest, line 95: @librod89 could we use a constant for 14, maybe put it in the CasaCase model?

librod89: Yes, good idea

lib/tasks/scheduler.rake
```diff
@@ -1,10 +1,3 @@
-desc "Transition youth who are now 14, run by heroku scheduler"
-task transition_youth: :environment do
```
compwron, line 2: @FireLemons this will make qa and prod scheduler fail, we should remove that when this is deployed

compwron (approved):
wow wow wow!!!

PR #4054, opened 2022-10-12
test scheduler

compwron:
let's see what this does xD


FireLemons:
maybe we can just install cron on heroku

FireLemons:
Nevermind it looks like it'll cost money.

PR #4064, opened 2022-10-13
Add models for UserLanguages backfill values

### LINKED ISSUE #3953: Add a unique constraint for volunteer, langauge pairs in the join table
**Description**
Add a unique constraint for languages a volunteer can have

image
As seen on the edit profile page for a volunteer

Login Emails: 

### Questions? Join Slack!

ludamillion:
Resolves #3953

### What changed, and why?
Added a `UserLanguage` model and new database table. Added migration to backfill the values from the existing `languages_users` join table.

New rspec model and request specs for the affected code. Manual test for the backfill migration.

### DIFF
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -1,6 +1,7 @@
 class Language < ApplicationRecord
   belongs_to :casa_org
-  has_and_belongs_to_many :users
+  has_many :user_languages
+  has_many :users, through: :user_languages
 
   validates :name, presence: true
 end
--- a/app/controllers/languages_controller.rb
+++ b/app/controllers/languages_controller.rb
@@ -36,10 +36,10 @@ class LanguagesController < ApplicationController
 
   def add_to_volunteer
     authorize @language
-    current_user.languages << @language
-    if current_user.save
+    begin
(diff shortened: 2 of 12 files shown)


app/models/language.rb
```diff
@@ -1,6 +1,7 @@
 class Language < ApplicationRecord
   belongs_to :casa_org
-  has_and_belongs_to_many :users
+  has_many :user_languages
+  has_many :users, through: :user_languages
```
compwron, line 4: scope this to casa org?

ludamillion: Will do. Still getting familiar with the data models and relationships.

ludamillion: How do you mean? Each language already belongs to a `casa_org` with a distinct record in the database. The uniqueness validation only cares about not associating the same `Language` record to the same `User` record more than once. On the _edit profile_ page the only languages available to select are those belonging to same organization as the user. So they can't select a language outside the scope of their organization.

compwron: that makes sense, works for me :)

compwron (approved):

PR #4073, opened 2022-10-18
Use friendly_id to have slugs CasaCase URLs instead of ids

### LINKED ISSUE #2416: Put Case Number and CASA Org in Case URLs
 - volunteers
 - supervisors
 - admins

**Description**
We would like to make it so that links to cases have the case number in the url. In addition, because multiple casa might have the same case number, we would like to add a short identifier based on the name of the casa to the url as well. An example of what a new url using this system would look like is `https://localhost:3000/casa_org/prince-george/casa_cases/CINA-10-1234`
The casa org saved identifier should not change even if the name of the casa org changes. 

Consider using the [FriendlyId gem](https://github.com/norman/friendly_id).

Casa case URL

_QA Login Details:_  

Login Emails: 

Thrillberg:
Resolves #2416 

### What changed, and why?
This PR introduces the `friendly_id` gem, which we will apply to the `CasaCase` model so that we can use URLs like this: `/casa_cases/CINA-15-1002` instead of `/casa_cases/42`. The advantage of the former is that it is recognizable by users and the disadvantage of the latter is that it reveals little bits of information about the data in our database... not desirable.

You may notice I had to call `parameterize` all over the place. This is basically just to downcase the slug since `friendly_id` expects parameterized slugs.

Should have no effect here.

Tests are updated and passing. No new tests were necessary.

### Screenshots please :)
Screen Shot 2022-10-20 at 1 16 28 PM

### Feelings gif (optional)
My approach to testing in gif form

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ gem "twilio-ruby" # twilio helper functions
 gem "draper" # adds decorators for cleaner presentation logic
 gem "faker" # creates realistic seed data, valuable for staging and demos
 gem "filterrific" # filtering and sorting of models
+gem 'friendly_id', '~> 5.4.0' # allows us to use a slug instead of casa case ids in their URLs
 gem "image_processing", "~> 1.12" # Set of higher-level helper methods for image processing.
 gem "jbuilder" # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem "jsbundling-rails"
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem "twilio-ruby" # twilio helper functions
 gem "draper" # adds decorators for cleaner presentation logic
 gem "faker" # creates realistic seed data, valuable for staging and demos
 gem "filterrific" # filtering and sorting of models
-gem 'friendly_id', '~> 5.4.0' # allows us to use a slug instead of casa case ids in their URLs
+gem "friendly_id", "~> 5.4.0" # allows us to use a slug instead of casa case ids in their URLs
 gem "image_processing", "~> 1.12" # Set of higher-level helper methods for image processing.
 gem "jbuilder" # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem "jsbundling-rails"
--- a/app/controllers/court_dates_controller.rb
+++ b/app/controllers/court_dates_controller.rb
@@ -66,7 +66,7 @@ class CourtDatesController < ApplicationController
   private
 
   def set_casa_case
-    @casa_case = current_organization.casa_cases.find(params[:casa_case_id])
+    @casa_case = current_organization.casa_cases.friendly.find(params[:casa_case_id])
   end
 
   def set_court_date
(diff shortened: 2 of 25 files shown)


app/models/casa_case.rb
```diff
@@ -20,7 +21,7 @@ class CasaCase < ApplicationRecord
   TRANSITION_AGE_YOUTH_ICON = "🦋".freeze
   NON_TRANSITION_AGE_YOUTH_ICON = "🐛".freeze
 
-  before_create :set_slug
+  friendly_id :case_number, :use => :scoped, :scope => :casa_org
```
littleforest, line 24: Colon syntax preferred over hash rocket style:

```suggestion
  friendly_id :case_number, use: :scoped, scope: :casa_org
```

Thrillberg: Thanks! Looks like this was picked up by the linter too, which is good. :)

app/javascript/src/require_communication_preference.js
```diff
@@ -33,7 +33,7 @@ $('document').ready(() => {
     displayPopUpIfPreferencesIsInvalid()
   })
 
-  smsToggle.addEventListener('change', () => {
+  smsToggle?.addEventListener('change', () => {
```
Thrillberg, line 36: This is unrelated to my current code here but was popping up an error in the console.

app/controllers/court_dates_controller.rb
```diff
@@ -78,6 +78,10 @@ def sanitized_params
       p[:case_court_orders_attributes]&.reject! do |k, _|
         p[:case_court_orders_attributes][k][:text].blank? && p[:case_court_orders_attributes][k][:implementation_status].blank?
       end
+
+      p[:case_court_orders_attributes]&.each do |k, _|
+        p[:case_court_orders_attributes][k][:casa_case_id] = @casa_case.id
+      end
```
Thrillberg, line 84: The above is some unfortunate gobbledygook that was necessary for court dates.

Thrillberg:
I see some yarn lint errors but I don't think I introduced them? Are other branches not having this? `yarn lint` runs clean on my machine.

Gemfile
```diff
@@ -20,6 +20,7 @@ gem "twilio-ruby" # twilio helper functions
 gem "draper" # adds decorators for cleaner presentation logic
 gem "faker" # creates realistic seed data, valuable for staging and demos
 gem "filterrific" # filtering and sorting of models
+gem "friendly_id", "~> 5.4.0" # allows us to use a slug instead of casa case ids in their URLs
```
compwron, line 23: we may want to unpin this at some point but this seems reasonable for now

app/javascript/src/court_order_list.js
```diff
@@ -12,10 +12,11 @@ function replaceNumberWithDecrement (str, num) {
 module.exports = class CourtOrderList {
   // @param {object} courtOrdersWidget The div containing the list of court orders
   constructor (courtOrdersWidget) {
-    const urlMatch = window.location.pathname.match(/^\/([a-z_]+)s\/(\d+)(\/(([a-z_]+)s))?/).filter(match => match !== undefined)
-
+    // The following regex is intended for pathnames such as "/casa_cases/CINA-19-1004/court_dates/new"
+    const urlMatch = window.location.pathname.match(/^\/([a-z_]+)s\/(\w+-+\d+)(\/(([a-z_]+)s))?/).filter(match => match !== undefined)
```
compwron, line 16: 😅

compwron (approved):
this is BIG!! :)

Issue #2627, opened 2021-09-21
Decouple Selection and Expansion on Emancipation Checklist


CovenantHuman:
**What type(s) of user does this feature affect?**
 - volunteers
 - supervisors
 - admins

**Description**
Currently when you expand a section of the emancipation checklist it automatically selects itself, and when you contract a section of the emancipation checklist it automatically deselects itself. We would like to make it so that sections can be expanded or contracted without changing their selection status (i.e. we would like to make it possible to select contracted sections, or deselect expanded sections without changing whether they are expanded or contracted and expand or contract sections without changing whether they are selected.) 

**Screenshots of current behavior, if any**  
![Expand without Checking](https://user-images.githubusercontent.com/7007647/134104023-01f6b124-1759-4b75-88b6-75feb357a45d.PNG)

_QA Login Details:_  
[Link to QA site](https://casa-qa.herokuapp.com/)  

Login Emails: 
- volunteer1@example.com  view site as a volunteer
- supervisor1@example.com view site as a supervisor
- casa_admin1@example.com view site as an admin

password for all users: 123456

compwron:
We still want this

vasconsaurus:
@compwron going to try this one 🙂

vasconsaurus:
Please give me more time 🙃 it's slow, but moving forward

vasconsaurus:
working on it! 🙂

PR #4081, opened 2022-10-19
Issue2627 prepwork decouple selection expansion emancipation checklist

vasconsaurus:
Progress on #2627

### What changed, and why?
Refactored the function to be more modular, so it's easier to track the collapse icon state and decouple actions. Right now it still works the same as before.

JS tests:

- will add some 🙂 

Rspec:

- spec/system/emancipations
- spec/views/emancipations
- spec/controllers/emancipations_controller_spec.rb
- spec/helpers/emancipations_helper_spec.rb

### Screenshots please :)
Screen Shot 2022-10-18 at 23 24 23

### DIFF
--- a/app/javascript/__tests__/case_emancipations.test.js
+++ b/app/javascript/__tests__/case_emancipations.test.js
@@ -1,11 +1,12 @@
 /* eslint-env jest */
-import { openChildren, closeChildren, deselectChildren, manageTogglerText } from '../src/case_emancipation'
+import { Toggler } from '../src/case_emancipation'
 
 require('jest')
 
 let category
 let categoryCollapseIcon
 let checkBox
+let toggler
 
 beforeEach(() => {
   document.body.innerHTML = `
@@ -33,34 +34,35 @@ beforeEach(() => {
   category = $('.emancipation-category')
   categoryCollapseIcon = $('.category-collapse-icon')
   checkBox = $('.emancipation-option-check-box')
+  toggler = new Toggler(category)
 })
 
 describe('Function that changes the text of the Toggler based on the state of the parent', () => {
   test('Changes the toggler text to -', () => {
     category.attr('data-is-open', 'false')
 
-    manageTogglerText(category)
+    toggler.manageTogglerText()
     expect(categoryCollapseIcon.text()).toEqual('+')
   })
 })
 
 describe('Function that opens the children of a given parent', () => {
   test('Opens the categoryOptionsContainer', () => {
-    openChildren(category)
+    toggler.openChildren()
     expect(category.data('is-open')).toEqual(true)
   })
 })
 
 describe('Function that closes the children of a given parent', () => {
   test('Closes the categoryOptionsContainer', () => {
-    closeChildren(category)
+    toggler.closeChildren()
     expect(category.data('is-open')).toEqual(false)
   })
 })
 
 describe('Function that deselects the children of a deselected parent', () => {
   test('Deselects the inputs in the categoryOptionsContainer', () => {
-    deselectChildren(category, () => '')
+    toggler.deselectChildren(() => '')
     expect(checkBox.prop('checked')).toBe(false)
   })
 })
--- a/app/javascript/__tests__/case_emancipations.test.js
+++ b/app/javascript/__tests__/case_emancipations.test.js
@@ -1,0 +1,66 @@
+/* eslint-env jest */
+import { openChildren, closeChildren, deselectChildren, manageTogglerText } from '../src/case_emancipation'
+
+require('jest')
+
+let category
+let categoryCollapseIcon
+let checkBox
+
+beforeEach(() => {
+  document.body.innerHTML = `
+  <div class="card card-container">
+    <div class="card-body">
+        <div>
+          <h6 class="emancipation-category no-select" data-is-open='false'>
+            <input type="checkbox" class="emancipation-category-check-box" value="1">
+            <label>Youth has housing.</label>
+              <span class="category-collapse-icon">+</span>
+          </h6>
+          <div
+            class="category-options"
+            style="display: none;">
+              <div class="check-item">
+                <input type="checkbox" id="O1" class="emancipation-option-check-box" value="1" checked>
+                <label>With Friend</label>
+              </div>
+          </div>
+        </div>
+    </div>
+  </div>
+  `
+
+  category = $('.emancipation-category')
+  categoryCollapseIcon = $('.category-collapse-icon')
+  checkBox = $('.emancipation-option-check-box')
+})
+
+describe('Function that changes the text of the Toggler based on the state of the parent', () => {
+  test('Changes the toggler text to -', () => {
+    category.attr('data-is-open', 'false')
+
+    manageTogglerText(category)
+    expect(categoryCollapseIcon.text()).toEqual('+')
+  })
+})
+
+describe('Function that opens the children of a given parent', () => {
+  test('Opens the categoryOptionsContainer', () => {
+    openChildren(category)
+    expect(category.data('is-open')).toEqual(true)
+  })
+})
+
+describe('Function that closes the children of a given parent', () => {
+  test('Closes the categoryOptionsContainer', () => {
+    closeChildren(category)
+    expect(category.data('is-open')).toEqual(false)
+  })
+})
+
+describe('Function that deselects the children of a deselected parent', () => {
+  test('Deselects the inputs in the categoryOptionsContainer', () => {
+    deselectChildren(category, () => '')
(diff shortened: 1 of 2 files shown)


app/javascript/__tests__/case_emancipations.test.js
```diff
@@ -60,7 +60,7 @@ describe('Function that closes the children of a given parent', () => {
   test('Closes the categoryOptionsContainer', () => {
     $(document).ready(() => {
       closeChildren(category)
-      expect(category.attr('data-is-open')).toContain('false')
+      expect(category.attr('data-is-open')).toContain('banana')
```
vasconsaurus, line 63: @compwron 
some tests that should have failed passed when I ran `npm test` locally, so I decided to check if the failure would show up when I pushed, but it doesn't seem to show up anywhere as a failure.

How should I go about this?

compwron:
Yay, I am glad you are working on this! :)

app/javascript/__tests__/case_emancipations.test.js
```diff
@@ -0,0 +1,75 @@
...
+    })
+  })
+})
+
+describe('Function that deselects the children of a deselected parent', () => {
+  test('Deselects the inputs in the categoryOptionsContainer', () => {
+    $(document).ready(() => {
+      deselectChildren(category)
+      // expect(how do I get the input and check if it's not checked ?)
```
tcione, line 72: @vasconsaurus I think you could use [`.is()`](https://api.jquery.com/is/). Something like: `$element.is(':checked')`. You could also get the native DOM element and use vanilla js (something like `$element[0].checked`)

app/javascript/src/case_emancipation.js
```diff
@@ -62,6 +62,43 @@ function saveCheckState (action, checkItemId) {
...
+
+export function openChildren ($parent) {
+  const categoryOptionsContainer = $parent.siblings('.category-options')
+
+  categoryOptionsContainer.show()
+  $parent.attr('data-is-open', 'true')
+}
+
+export function closeChildren ($parent) {
+  const categoryOptionsContainer = $parent.siblings('.category-options')
```
tcione, line 83: **nitpick**: the functions point that they receive a jquery element by using `$` at the start of the variable name (`$parents`). Then we store a different jquery element without using `$` (`categoryOptionsContainer`). I think being consistent would make code clearer (so either using `$` for all jquery elements (`$parents` and `$categoryOptionsContainer`, or using nothing)

app/javascript/src/case_emancipation.js
```diff
@@ -62,6 +62,43 @@ function saveCheckState (action, checkItemId) {
     })
 }
 
+export function manageTogglerText ($parent) {
```
tcione, line 65: **suggestion:** since these methods usually should act under the same parent, it might be worthwhile to wrap this in a class or using closures.

Something like
```js
class Toggler {
  constructor($parent) {
    this.$parent = $parent
  }

  manageTogglerText() {
    // your method using this.$parent
  }
}

const $category = $('selector')
const toggler = new Toggler($category)
toggler.manageTogglerText()
```

or

```js
function Toggler($parent) {
  const manageTogglerText = () => {
    // your method using $parent
  }

  return { manageTogglerText }
}

const $category = $('selector')
const toggler = Toggler($category)
toggler.manageTogglerText()
```

app/javascript/__tests__/case_emancipations.test.js
compwron: hmmmm that is concerning, but they are failing in CI... maybe your local is having caching issues
or maybe we are not running all the tests I should be

app/javascript/src/case_emancipation.js
```diff
@@ -62,6 +62,43 @@ function saveCheckState (action, checkItemId) {
...
+  const categoryCollapseIcon = $parent.find('.category-collapse-icon')
+
+  if ($parent.attr('data-is-open') === 'true') {
```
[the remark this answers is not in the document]
compwron: I don't understand how that would work but if it would work I am onboard!

app/javascript/__tests__/case_emancipations.test.js
```diff
@@ -0,0 +1,75 @@
...
+  categoryOptionsContainer = document.querySelector('.category-options')
+})
+
+describe('Function that changes the text of the Toggler based on the state of the parent', () => {
+  test('Changes the toggler text to -', () => {
+    $(document).ready(() => {
+      category.attr('data-is-open', 'false')
+
+      manageTogglerText(category)
+      expect(categoryCollapseIcon.text).toContain('+')
```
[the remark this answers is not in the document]
vasconsaurus: uuuuuu I'll take a look at this (an all the other suggestions 😉 ) thanks ❤️

app/javascript/__tests__/case_emancipations.test.js
vasconsaurus: so, I don't think `toContain` was the problem, as suggested [here](https://github.com/rubyforgood/casa/pull/4081/files/d0fe2fbc2df16fc0e4c35f7fe4db342a07ac5a04#r1002404117).

So, when I started to write the tests and ran `npm test`, they were failing when they should, but at some point, they stopped failing. Even if I comment my variables, and test for things like 'banana' it still passes.

Now it seems my case_emacipation.js stopped working (other js files work fine), even when I'm on the main branch, so I don't think any updates to the file on my branch are the problem. I think there is probably something wrong with my environment, but I'm at a loss about what it could be.

app/javascript/src/case_emancipation.js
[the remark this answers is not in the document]
tcione: In the end it's bit silly and probably boils down to preference. But it would become something like `parent.data('is-open')`

app/javascript/__tests__/case_emancipations.test.js
vasconsaurus: @tcione helped me and we figured out what was wrong 😄 There are two different problems:

1- I followed the pattern from one of the test files that used `$(document).ready(() => {` inside the tests, turns out those tests aren't running. So the problem with my tests was they weren't running. There are other tests that aren't running as well. I'll open an issue explaining this better.

2- On why the js stopped working: there is a regex in `line 104` that tested only for digits in the pathname, but 
this is the pathname we get: '/casa_cases/CINA-07-1003/emancipation'. If I'm not mistaken, you talked about how the pathname was going to change last office hours. So that is why this also broke on main.

vasconsaurus: I added two small issues for these bugs, and went ahead and fixed the regex one:
[issue#4107](https://github.com/rubyforgood/casa/issues/4107) and [issue#4108](https://github.com/rubyforgood/casa/issues/4108)/[PR for issue#4108](https://github.com/rubyforgood/casa/pull/4109)

vasconsaurus:
@compwron I'm finally done with this one 🙂 But while I was trying to test git rebase, I think I made a big mess of the commit history... I'm very sorry for the mess 😞

vasconsaurus:
> @compwron I'm finally done with this one 🙂 But while I was trying to test git rebase, I think I made a big mess of the commit history... I'm very sorry for the mess 😞

The problem was worse than just the commits, but it is fixed 🙃 
BUT I just found a small bug, so there is still that to fix 🫠

vasconsaurus:
> The problem was worse than just the commits, but it is fixed 🙃 BUT I just found a small bug, so there is still that to fix 🫠

@compwron It's DONE! Can you please review this for me? 
What I thought was a bug was actually the original behavior 🙃

app/javascript/src/case_emancipation.js
vasconsaurus: I had forgotten about this, but updated it to use a class

app/javascript/__tests__/case_emancipations.test.js
```diff
@@ -0,0 +1,68 @@
...
+  test('Closes the categoryOptionsContainer', () => {
+    toggler.closeChildren()
+    expect(category.data('is-open')).toEqual(false)
+  })
+})
+
+describe('Function that deselects the children of a deselected parent', () => {
+  test('Deselects the inputs in the categoryOptionsContainer', () => {
+    toggler.deselectChildren(() => '')
+    expect(checkBox.prop('checked')).toBe(false)
```
tcione, line 66: **nitpick:** `toBe` -> `toEqual` (for consistency)

tcione (approved):
LGTM

Thanks for going through all the comments and changes. That said, my approval is merely symbolic, so please wait for @compwron review :D

compwron (approved):

PR #4082, opened 2022-10-19
architecture decision around tests

compwron:
I want to clarify what tests we want and when

### DIFF
--- a/doc/architecture-decisions/0008-controller-specs.txt
+++ b/doc/architecture-decisions/0008-controller-specs.txt
@@ -37,4 +37,4 @@ If both controller and request tests for a file exist, add to the request tests,
 
 You should generally write request tests rather than controller tests.
 
-You should write system tests *in addition* to
+You should write system tests *in addition* to request specs.
--- a/doc/architecture-decisions/0008-controller-specs.txt
+++ b/doc/architecture-decisions/0008-controller-specs.txt
@@ -35,6 +35,6 @@ If only controller tests for a file exist, add to the controller tests
 
 If both controller and request tests for a file exist, add to the request tests, and consider moving over the controller test tests to the request test.
 
-You should generally write request tests rather than controller tests
+You should generally write request tests rather than controller tests.
 
 You should write system tests *in addition* to
--- a/doc/architecture-decisions/0008-controller-specs.txt
+++ b/doc/architecture-decisions/0008-controller-specs.txt
@@ -13,7 +13,7 @@ expect(response).to be_successful
 2.
 "request tests" have filenames like `spec/requests/contact_types_spec.rb` and syntax like
 ```
-sign_in_as_admin
+sign_in admin
 get new_contact_type_path
 expect(response).to be_successful
 ```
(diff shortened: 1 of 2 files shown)


doc/architecture-decisions/0008-controller-specs.txt
```diff
@@ -0,0 +1,40 @@
+# 1. Testing controllers
+
+Date: 2022-10-18
+
+1.
+"controller tests" have filenames like `spec/controllers/casa_admins_controller_spec.rb` and syntax like
```
littleforest, line 6: You mention it at the end of this file, but I think it would be great to mention directly in this section that controller specs are deprecated and they should only be used if a file exists, and preferably the whole file is rewritten and moved to the requests directory.

doc/architecture-decisions/0008-controller-specs.txt
```diff
@@ -0,0 +1,40 @@
...
+```
+allow(controller).to receive(:authenticate_user!).and_return(:supervisor)
+log_in_as_admin
+get :index
+expect(response).to be_successful
+```
+2.
+"request tests" have filenames like `spec/requests/contact_types_spec.rb` and syntax like
+```
+sign_in_as_admin
```
littleforest, line 16: I would just use:
```suggestion
sign_in admin
```

as the example which just uses the Devise helper to sign in.

doc/architecture-decisions/0008-controller-specs.txt
```diff
@@ -0,0 +1,40 @@
...
+fill_in "Current Password", with: "12345"
+fill_in "New Password", with: "123456789"
+fill_in "New Password Confirmation", with: "123456789"
+click_on "Update Password"
+expect(page).to have_content "1 error prohibited this password change from being saved:"
+```
+
+If only controller tests for a file exist, add to the controller tests
+
+If both controller and request tests for a file exist, add to the request tests, and consider moving over the controller test tests to the request test.
```
littleforest, line 36:
```suggestion
If both controller and request tests for a file exist, add to the request tests, and consider moving over the controller tests to the request test.
```

doc/architecture-decisions/0008-controller-specs.txt
```diff
@@ -0,0 +1,40 @@
...
+
+You should generally write request tests rather than controller tests
```
littleforest, line 38:
```suggestion
You should generally write request tests rather than controller tests.
```

doc/architecture-decisions/0008-controller-specs.txt
```diff
@@ -0,0 +1,40 @@
...
+
+You should write system tests *in addition* to
```
littleforest, line 40:
```suggestion
You should write system tests *in addition* to request specs.
```

doc/architecture-decisions/0008-controller-specs.txt
littleforest: I think it would also be great to open a ticket to have someone move controller specs to request specs (maybe one-at-a-time to avoid merge conflicts).

compwron: agreed

compwron:
merging for now- more changes may come

PR #4093, opened 2022-10-20
fix for #4084, contact type validation added for create

### LINKED ISSUE #4084: Bug: Casa Contact Types are shown as required on UI, but we can still submit the case with all uncheck
### Impacted User Types
 - admins
 - all casa admins

### Environment
All Environment

### Current Behavior
We can create the case without the contact type selection, even on the UI it marked as required by showing the `*` symbol after the label.

### Expected Behavior
- We should make the casa contact type as required on the UI and backend

### How to Replicate

1. - Log into staging as an admin
2. - Click on "Cases" in the left sidebar menu.
3. - Create a "New Case"
4. - Fill all the details and Uncheck All contact types, then submit the form
5. - It will create the new case
6. - Now edit the same case, you will see all the contact types are checked, but we submitted as unchecked

### Screenshot
Screenshot 2022-10-19 at 3 25 14 PM

rishijain:
Resolves #4084 

### What changed, and why?
We have added the call for validation before a case is created. We have specifically not checked this for "volunteers" as we figured volunteers cannot create a case.

- Volunteer permissions: No impact
- Supervisor permissions: No impact
- Admin permissions: No impact

Added a spec for the change, and manual testing on the browser.

### DIFF
--- a/app/controllers/casa_cases_controller.rb
+++ b/app/controllers/casa_cases_controller.rb
@@ -47,6 +47,7 @@ class CasaCasesController < ApplicationController
     )
     authorize @casa_case
 
+    @casa_case.validate_contact_type = true
     if @casa_case.save
       respond_to do |format|
         format.html { redirect_to @casa_case, notice: "CASA case was successfully created." }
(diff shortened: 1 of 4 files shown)


compwron (approved):
:)

PR #4106, opened 2022-10-23
All cases are selected by default on new case contact page

### LINKED ISSUE #3334: Multi-case volunteer New Case Contact default should be both cases
 - volunteers

**Description**
I have two cases in the same family and often the case contacts need to have both case numbers tagged. After I have put in a contact and it takes me back to the list of all case contacts, if I select New Case Contact there, I am only given a form with one of the two case numbers whether or not the contact I just entered included both case numbers. If I select "Case Contacts" on the side bar, then THAT new case contact button gives me both. It's frustrating when I am entering several contacts in a row, as is often the case.

https://casa-qa.herokuapp.com/case_contacts/new logged in as volunteer2
Screen Shot 2022-04-01 at 11 11 23 PM
Screen Shot 2022-04-01 at 11 11 15 PM

_QA Login Details:_  

Login Emails: 

### Questions? Join Slack!

ChaelCodes:
**This is what was requested, but I don't think it's the right solution. See #4147 and #4148 for better solutions. I would be happy to see this PR closed, but it is what was requested in the issue.**

Resolves #3334 

### What changed, and why?
The user says:
> I have two cases in the same family and often the case contacts need to have both case numbers tagged. After I have put in a contact and it takes me back to the list of all case contacts, if I select New Case Contact there, I am only given a form with one of the two case numbers.

The issue interprets this to mean that all cases should be selected by default. (At least that's how I understood it.) However, creating a new case contact redirects to the case page. The case page has a "new case contact" button, and that button opens a new case contact form with only one case on it (the case from the page). I believe this form is what the user is actually having issues with.

Here's a little clip explaining how it works today and the issue I'm quoting:
https://clips.twitch.tv/MoldyPatientPistachioCeilingCat-CAd_vwCrwNCqA04E

This PR has all cases selected by default, but I think we should rethink this - many volunteers may have multiple cases, and unselecting is kind of painful.

I think there are two better solutions.

1. Redirect to the case contact page if there are multiple cases selected, instead of redirecting to the first. #4147 
2. Default from the case page to selecting the correct case, but don't remove additional options from the form. #4148

- Volunteer permissions:
- Supervisor permissions:
- Admin permissions:

There are system specs for the default selection.

### GIF
hacking cat

### DIFF
--- a/app/controllers/case_contacts_controller.rb
+++ b/app/controllers/case_contacts_controller.rb
@@ -37,7 +37,7 @@ class CaseContactsController < ApplicationController
     @case_contact = CaseContact.new
 
     # By default the first case is selected
-    @selected_cases = @casa_cases[0, 1]
+    @selected_cases = @casa_cases
 
     @selected_case_contact_types = @selected_cases.flat_map(&:contact_types)
 
--- a/spec/system/case_contacts/new_spec.rb
+++ b/spec/system/case_contacts/new_spec.rb
@@ -260,12 +260,14 @@ RSpec.describe "case_contacts/new", type: :system do
       build(:contact_type, name: "Hidden", active: false, contact_type_group: grp_with_hidden)
       volunteer = create(:volunteer, :with_casa_cases)
       volunteer_casa_case_one = volunteer.casa_cases.first
+      volunteer_casa_case_two = volunteer.casa_cases.second
       create_contact_types(volunteer_casa_case_one.casa_org)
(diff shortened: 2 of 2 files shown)


compwron (approved):

ChaelCodes:
This one directly conflicts with the goals of https://github.com/rubyforgood/casa/pull/4148, since that one merged, do you want me to close this one?

PR #3637, opened 2022-06-10
Let Supervisors and Admins Hide Case Contacts Made by Previous Volunteers on a Case

arthur1041:
### What github issue is this PR for, if any?
Resolves #3338

### What changed, and why?
Supervisors and Admins are now able to choose whether new volunteers see old volunteers' case contacts on a casa case.

### How will this affect user permissions?
- Volunteer permissions: Will only be able to see old case contacts if an admin or supervisor allows it.
- Supervisor permissions: Can define whether volunteers can see old case_contacts or not.
- Admin permissions:

### How is this tested? (please write tests!) 💖💪
There were added system and request tests. Just run rspec.

### Screenshots please :)
Screenshot 2022-06-10 11:57:06

PR #4132, opened 2022-10-26
Only allow admins and supervisors of the same org to show or hide contacts

### LINKED ISSUE #4065: Limit case_assignment controller method permissions to the case assignment's organization
 - volunteers
 - supervisors
 - admins

**Description**
A case assignment links a casa case to a volunteer user

Make sure all the methods in [`app/policies/case_assignment_policy.rb `](https://github.com/rubyforgood/casa/blob/main/app/policies/case_assignment_policy.rb ) contain some logic to prevent casa case assignment access from users belonging to a different organization. 

Fix / Update tests if needed

See [`app/policies/application_policy.rb`](https://github.com/rubyforgood/casa/blob/main/app/policies/application_policy.rb) for helper methods
See https://github.com/varvet/pundit

_QA Login Details:_  

Login Emails: 

### Questions? Join Slack!

mateusdeap:
Resolves #4065.

### What changed, and why?
We tentatively altered the `show_or_hide_contacts?` method to only check if the user is an admin or supervisor and if they are in the same org as the casa case.

The reason is that the implementation added in #3637 seemed to only return true, since it or'ed `hide_contacts?` and `!hide_contacts?`

We might be wrong somehow, but decided to try and will revert if we break anything.

- Volunteer permissions:
  - Apparently, no changes
- Supervisor permissions:
  - Can only change contact exhibition if in the same org
- Admin permissions:
  - Can only change contact exhibition if in the same org

We added specs to `case_assignment_policy_spec.rb`

### Feelings gif (optional)
confused...


compwron (approved):
seems like an improvement 😅

app/policies/case_assignment_policy.rb
```diff
@@ -29,9 +29,11 @@ def hide_contacts?
   end
 
   def show_or_hide_contacts?
-    hide_contacts? || !hide_contacts?
+    admin_or_supervisor? && same_org?
   end
 
+  private
```
JuanVqz, line 35: :rocket: :1st_place_medal:

JuanVqz:
> seems like an improvement sweat_smile

what do you mean?
As far as I understand [we were required to change a method per PR.](https://github.com/rubyforgood/casa/issues/4065#issuecomment-1284046055)

So, this PR is implementing the `same_org?` method in [def show_or_hide_contacts?](https://github.com/rubyforgood/casa/pull/4132/files#diff-1e00963bb42432cb0649cf85431cb8b0a5ec72528fa403571a247aa299464644R31) but I don't know why @mateusdeap changed the rest of the logic there.

mateusdeap:
Because before it was doing `hide_contacts? || !hide_contacts?`, which can only return true. I think that's what @compwron meant.

We do have a failure, however, so we'll need to check that out. Maybe just returning true was indeed needed 🤣

compwron:
That failure may be a flake :( I re-ran it

compwron:
Merging because only docker failed and therefore it's a flake :/

PR #4137, opened 2022-10-27
remove all casa admin controller tests and move non-duplicates to req…

farrelld09:
We compared the controller spec to the request spec, and added any test expectations that were present in the controller but missing in the request, to the request spec.

Deleted controller spec afterward.

Please advise if this was the correct approach/meets expectations.


compwron (approved):
:)

compwron (approved):

Issue #3235, opened 2022-02-26
Prince George's Fund Application Form


compwron:
**What type(s) of user does this feature affect?**
non logged in users and stakeholder AM

**Description**
SB: targeted PG opportnities fund
gnerate applications for fund, email to add marie
a fillable form, send the form, “a hell of the time for the form to work for everyone” same doc, can’t fill it.
adobe document, reader.
new form that looks the same, questions,
pay-ee, other resuorces, rewuested…
SB; needs anme of kid and we dont do that… case cina number onlu
or vol name and AM can figure out kid
form goes to other parites also
volunteers in activity tracker could go there and submit it
lawyers submit the form too, dont have access to…
anyone can request funds… anyone in foster care not just casa youth
AM: no face to allowing lawyers in if they’re limited to that part of the sytem, a link send to them, here’s how to fill out the form…
if coming from outside casa, do ned the name because AM cant figure out, dont have access to all cina numbers
AM doesnt have access to cina if were not the case, no centralized case to look up cina numbers that thye ahave access to
“there will be someday” a place cina numbers lookup
cant get into “effec yet”
SB: 400 ish kids in care, PG handles 180 of them, dunno anything about them
and they send the forms to AM even if the kids arent in foster youth
donor wanted fund to live somewhere so its with casa “do it quickly wo beuaracracy, aoperatined by dept of social services, takesn months…”
AM can turn it around in an adg=ernnt
fund paid for braces, summer camp, …
some kid studio time for music
AM: reunification plane ticket

compwron:
working on v2 in my brain

compwron:
https://github.com/rubyforgood/casa/pull/3237

compwron:
working on itttt

littleforest:
@compwron I am closing this since looks like there is a merged PR in #3355

Issue #2897, opened 2021-11-02
re-enable pundit safety checks, fix tests


compwron:
```
  # after_action :verify_authorized, except: :index # TODO add this back and fix all tests
  # after_action :verify_policy_scoped, only: :index
```

LawrenceWhalen:
@compwron I can take this on. Just to be sure, you want both lines re-added?

compwron:
@LawrenceWhalen yes :)

compwron:
We still want this

littleforest:
@compwron @FireLemons what are your thoughts about removing Pundit requirements for `all_casa_admins` controllers? An `all_casa_admin` user should have access to any action in those controllers (currently). I'm leaning towards just removing Pundit requirements for those controllers since everything would default to true anyway. What are your thoughts?

littleforest:
@compwron should any supervisor be able to edit or delete any note, even if it is not for their volunteer?

compwron:
@littleforest that seems reasonable to remove for all-casa admins

re supervisors and notes... probably yes, but if it's not that way it's not a huge deal as long as admins can do it and the notes aren't volunteer-visible

littleforest:
@compwron can volunteers do any actions with notes?

littleforest:
still working on this

littleforest:
@compwron I think the last controller that needs to be fixed is the `other_duties` controller. I'm a little confused about that feature though. I see a link to the `other_duties#index` page on both Supervisor and Admins, but no link to the `new` page (although it exists and I can manually visit the new page). Do you know what the permissions should be like? Is someone creating a duty for themselves? From the controller index method, it looks like a duty is getting created for a volunteer, but there is no volunteer on the form field. Anyway, wondering if this feature is somewhat unfinished, but I'm not seeing any open tickets around it.

littleforest:
@compwron never mind! I found details in #3556

compwron:
cool cool

littleforest:
The only thing left with this is the OtherDuties controller. Even though this feature might be going away? (#4242) I'll try to get this piece finished soon.

PR #4176, opened 2022-11-06
Add pundit safety checks to FundRequest controller

littleforest:
Ref #2897 
Ref #3235 

### What changed, and why?

This adds Pundit safety checks for the FundRequest controller, allowing any type of user to submit a fund request. The FundRequest form is not actually in use yet, so these permissions could be changed later if needed.

This also modifies the FundRequest form to use the nested version of the route so that we can delete the extraneous fund_request routes. The form was also updated to add `for` attributes to the labels for accessibility. 

- Volunteer permissions: can create fund request
- Supervisor permissions: can create fund request
- Admin permissions: can create fund request

Currently this allows a volunteer to submit a request for a casa case that they are not assigned to, but this can also be modified in the future if needed.

New request specs were written and the old controller spec file was removed. A new system spec for the FundRequest form was also added.


compwron (approved):
very nice! :)

PR #4207, opened 2022-11-16
tell cloud66 to use puma like we currently do

compwron:
as per https://help.cloud66.com/rails/how-to-guides/rack-servers/puma-rack-server.html


compwron:
let's merge it and see if it breaks qa

PR #4298, opened 2022-12-14
Created not_found(404) and internal_server_error(500) pages.

### LINKED ISSUE #4150: Create a route that intentionally 500s the app
**Description**
create a `/error` route that raises an uncaught exception causing a 500 error.
Visiting the route page should result in this:
image

Write a test for the new route

_Login Details:_  

Login Emails: 

### Questions? Join Slack!

raychiranjib1:
Resolves #4150

### What changed, and why?

Generated new controller ErrorsController and added two actions - not_found and internal_server_error.

For wrong routes it will render to 404 page and if any exception causes it will render 500 page.

Specs are covered for the new routes.

### Screenshots please :)
Screenshot 2022-12-14 at 11 21 12 PM

Screenshot 2022-12-14 at 11 21 58 PM


compwron (approved):
:)

FireLemons:
@raychiranjib1 I think this pr may have been merged prematurely.

We don't need any of the 404 pages. We can look at a 404 page simply by visitng a route that does not exist in our hosted environments. For example: https://casa-qa.herokuapp.com/sdelksajdlksajlda. Could you please delete the code for 404 errors. 

It looks like you've created a custom view for 500 error pages `app/views/errors/internal_server_error.html.erb` but the view does not show up when a 500 error happens.  
  
Your new 400 error page shows up for 400 errors but the footer on it leaves a lot of space under it
image
would you consider revising the page so that the footer appears normally or reverting the page back to the old 404 seen here:
image

Could you explain the line `config.exceptions_app = routes`? Your feature seems to work without it.

Could you also explain the lines 
```
  match "/404", to: "errors#not_found", via: :all
  match "/500", to: "errors#internal_server_error", via: :all
```
in `config/routes.rb` ? The feature works without them as well.

In your tests you have
```
RSpec.describe "Errors", type: :request do
  describe "GET /not_found" do
    it "returns http success" do
      get "/404"
      expect(response).to have_http_status(:success)
    end
  end

  describe "GET /internal_server_error" do
    it "returns http success" do
      get "/500"
      expect(response).to have_http_status(:success)
    end
  end
end
```

the describe block is labeled `GET /not_found` but you don't visit the route `/not_found` in the test.

the response status for a not found page should be 404
the response status for an internal server error page should be 500-599

success is 200 and that means no error has occurred 

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

Could you address some of these concerns in a followup PR?

raychiranjib1:
@FireLemons 

Here is the followup PR @ https://github.com/rubyforgood/casa/pull/4344
Unfortunatly it's breaking docker's test. I found similar error in the other PRs, hence it's something common problem exist before.

In the first PR, I have followed the comment added on the issue [here](https://github.com/rubyforgood/casa/issues/4150#issuecomment-1308123051). 

Currently I reverted back the code, becuase I am little confused on the requirements. But I changed defalt 500.html page design similar to 404.html. 
Please add comment on that PR if anything required or we can discuss on weekly call.

PR #4350, opened 2022-12-22
Applied new theme to reports page

### LINKED ISSUE #4325: New Theme - reports pages
**Description**

This is part of #4304

We're updating the theme of the app to use the [plainadmin](https://plainadmin.com/docs/) theme and going through the process of converting the site to that theme. 

## Directions

* Checkout the `new_theme` branch and create a branch off of that branch for your work.
* Update all the views in the the `/views/reports` to use the new theme.
* The `/views/supervisor/index.html.erb` has already been converted to use the new theme, use it as a reference when converting over.
* If there are missing elements use the [plainadmin docs](https://plainadmin.com/docs/) and the [bootstrap 5 docs](https://getbootstrap.com/docs/5.1/getting-started/introduction/) as reference.
* When submitting your PR submit it against the `new_theme` branch.

The following files need to be converted to the new theme:
- [ ] _filter.html.erb
- [ ] index.html.erb

### Things to consider:

It is difficult to keep the `new_theme` branch up to date so please be considerate in claiming issues and keep the team apprised of your progress.

vasconsaurus:
Resolves #4325

### What changed, and why?
The new theme was applied to the index reports page and to the filter modal.

But, I have a few questions/observations:
1. Regarding the select input
* I managed to add the placeholder using `prompt: true`, this is a bit different from how it was written before. So I'm not sure if it's a good solution.
* I also had to remove `{include_hidden: false}`, I tested and it seemed ok, but I'm not sure if this was important.

2. Regarding labels
* Should the label be equal to the table headers (from the supervisor's index)? In the tables, we have an h6 with the table header. 'Starting From' label:
Screenshot 2022-12-21 at 22 20 10

3. Buttons
* What do you think? Is this ok?
* Should they all be primary blue?
* Should they have icons? If yes, which?
* I noticed hover isn't working on the primary buttons in the index, I still have to look into that.

### Screenshots please :)
Screenshot 2022-12-21 at 22 04 53
Screenshot 2022-12-21 at 22 05 07
Screenshot 2022-12-21 at 22 05 20
Screenshot 2022-12-21 at 22 05 34
Screenshot 2022-12-21 at 22 05 37

### Tiny video ;)
https://user-images.githubusercontent.com/87862340/209034716-616d44ca-3c67-4c84-a0c7-9c3f927751ba.mov

### DIFF
--- a/app/javascript/src/reports.js
+++ b/app/javascript/src/reports.js
@@ -12,11 +12,11 @@ const handleReportFormSubmit = (event) => {
   const buttonText = event.target.innerHTML
 
   event.target.disabled = 'disabled'
-  event.target.value = event.target.dataset.disableWith
+  event.target.innerHTML = event.target.dataset.disableWith
   event.target.form.submit()
 
   setTimeout(() => {
     event.target.disabled = false
-    event.target.value = buttonText
+    event.target.innerHTML = buttonText
   }, 3000)
 }
--- a/app/javascript/src/reports.js
+++ b/app/javascript/src/reports.js
@@ -9,7 +9,7 @@ document.addEventListener('DOMContentLoaded', () => {
 const handleReportFormSubmit = (event) => {
   event.preventDefault()
 
-  const buttonText = event.target.value
+  const buttonText = event.target.innerHTML
 
   event.target.disabled = 'disabled'
   event.target.value = event.target.dataset.disableWith
--- a/app/views/static/index.html.erb
+++ b/app/views/static/index.html.erb
@@ -69,7 +69,7 @@
 
 <div class="w-3/4 mx-auto py-24">
 
-  <div class="flex flex-col md:flex-row justify-center">
+  <div class="flex flex-col md:flex-row justify-center items-center">
     <div class="flex flex-col justify-center items-center w-full sm:w-1/2">
       <div class="pb-2">
         <%= image_tag("case-contact.svg", class: "app-images") %>
--- a/app/views/reports/_filter.html.erb
+++ b/app/views/reports/_filter.html.erb
@@ -5,18 +5,20 @@
         <h5 class="modal-title" id="filterColumnsLabel">
           Filter Exported Columns
         </h5>
-        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+        <button type="button" class="main-btn primary-btn btn-hover  btn-sm close" data-dismiss="modal" aria-label="Close">
           <span aria-hidden="true">&times;</span>
         </button>
       </div>
       <div class="modal-body">
         <div>
-          Select columns:
+          <h6 class="mb-20">
+            Select columns
+          </h6>
           <%= form.fields_for :filtered_csv_cols do |ff| %>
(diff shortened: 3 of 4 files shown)


seanmarcia:
Looks awesome!!! Have you thought about adding some icons to the buttons?
Maybe - `dashboard` for the mileage, `database` for the data report, `hourglass` for the learning hours, `envelope` for the emails, and `check-box` for the followups?

seanmarcia:
I really, really, really love the video :)

compwron (approved):
:)

seanmarcia:
It might be a bit confusing to add in icons since those buttons are actually little form objects. To add in icons you'd want to do something like:
```
<%= form_with url: mileage_reports_path(format: :csv), method: :get do |f| %>
  <%= button_tag(
        data: { disable_with: "Downloading Mileage Report" },
        class: "main-btn primary-btn btn-hover report-form-submit"
      ) do %>
      <i class="lni lni-dashboard mr-10"></i> Mileage Report
  <% end %>
<% end %>
```

You'll also want to update the `reports.js` to change the innerHTML rather than the value as there is an icon now part of the button. I imagine something like the following should fix that button transition that is there:
```
const handleReportFormSubmit = (event) => {
  event.preventDefault()

  const buttonText = event.target.innerHTML

  event.target.disabled = 'disabled'
  event.target.innerHTML = event.target.dataset.disableWith
  event.target.form.submit()

  setTimeout(() => {
    event.target.disabled = false
    event.target.innerHTML = buttonText
  }, 3000)
}
```

app/views/static/index.html.erb
```diff
@@ -69,7 +69,7 @@
 
 <div class="w-3/4 mx-auto py-24">
 
-  <div class="flex flex-col md:flex-row justify-center">
+  <div class="flex flex-col md:flex-row justify-center items-center">
```
seanmarcia, line 72: yay!

vasconsaurus:
@seanmarcia I updated the buttons to have the icons, and now the hover is fixed 🙂

vasconsaurus:
Tiny video with the updated buttons:
https://user-images.githubusercontent.com/87862340/209213760-13eba2c7-a737-4af7-9dc7-8efb4d2085b2.mov

app/javascript/src/reports.js
```diff
@@ -9,7 +9,7 @@ document.addEventListener('DOMContentLoaded', () => {
 const handleReportFormSubmit = (event) => {
   event.preventDefault()
 
-  const buttonText = event.target.value
+  const buttonText = event.target.innerHTML
 
   event.target.disabled = 'disabled'
   event.target.value = event.target.dataset.disableWith
```
seanmarcia, line 15: I think you'll need to change this to `event.target.innerHtml = event.target.dataset.disableWith`

seanmarcia: And on line 20 you'll need to do the same  `event.target.innerHTML`

vasconsaurus: I see! It makes sense. Thanks for changing it for me 🙂

PR #4351, opened 2022-12-22
Require Pundit authorization for all controller methods

### LINKED ISSUE #2897: re-enable pundit safety checks, fix tests

littleforest:
Resolves #2897 

### What changed, and why?
This adds request specs and authorization for the `other_duties` controller. It also requires Pundit authorization on all controller methods, except index and Devise controllers. It also does not require Pundit authorization for the `all_casa_admins` controller as currently there is only one time of `all_casa_admin` user and all of those users can access all methods within that controller.

The `resolve` method in the OtherDuty policy was removed since it is not in use and is invalid anyway since there is no `casa_org` field on the `other_duties` table.

Request specs were added for the other_duties controller.


compwron (approved):
this is great!! :)

PR #4378, opened 2022-12-28
moved language description to under language section

FireLemons:
### What changed, and why?
Moved language description out of new page to under section header.

### Screenshots please :)
image


seanmarcia:
Any chance you want to make this change in the `new_theme` branch? It is getting pretty close to completion and I think the language section has already been converted there.

PR #4387, opened 2023-01-02
Updates helper methods and sidebar partial to show correct active links

anshu1992:
### What github issue is this PR for, if any?
Resolves #4331 

### What changed, and why?
`menu_item` method of sidebar helper and sidebar partial to show correct active link.

### How will this affect user permissions?
- Volunteer permissions: None
- Supervisor permissions: None
- Admin permissions: None

### How is this tested? (please write tests!) 💖💪
Visually.

### Screenshots please :)

Screenshot 2023-01-02 at 8 09 04 AM

Screenshot 2023-01-02 at 8 08 57 AM

Screenshot 2023-01-02 at 8 08 48 AM

Screenshot 2023-01-02 at 8 08 38 AM

### Feelings gif (optional)
What gif best describes your feeling working on this issue? https://giphy.com/
How to embed:
`alt text`

### Feedback please? (optional)
We are very interested in your feedback! Please give us some :) https://forms.gle/1D5ACNgTs2u9gSdh9

PR #4419, opened 2023-01-06
Don't show group actions expanded by default

anshu1992:
Hey @seanmarcia, here is the fix you requested.

Issue description: [Fix requested in comment](https://github.com/rubyforgood/casa/pull/4387#issuecomment-1369111839)

### What changed, and why?
CSS class for not showing the dropdown expanded by default.

- Volunteer permissions: None
- Admin permissions: None

Visually

### Screenshots please :)

Screenshot 2023-01-06 at 8 40 15 PM

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/views/layouts/_sidebar.html.erb
+++ b/app/views/layouts/_sidebar.html.erb
@@ -107,12 +107,11 @@
             aria-expanded="true"
             aria-label="Toggle navigation">
               <span class="icon">
-              <i class="lni lni-list mr-10"></i>
-
+                <i class="lni lni-list mr-10"></i>
               </span>
             <span class="text">Group Actions</span>
           </a>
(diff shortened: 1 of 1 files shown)


anshu1992:
Currently, it doesn't. I will check how to get keep it open.

anshu1992:
The Bootstrap dropdown is not supposed to keep open after selection, and so it closes automatically. I think to avoid that problem, the `show` class was there to force it to keep open.

I think we can do is use an accordion instead of that dropdown to avoid these two problems.

seanmarcia:
I think the `show` was added as an example to display how that works. Could we add a method like the `active_class` that adds `show` in if one of the items in the dropdown active class is selected?

anshu1992:
On it,thanks!

Issue #4129, opened 2022-10-26
Move controller specs to request specs


compwron:
Move controller specs to request specs - see https://github.com/rubyforgood/casa/blob/main/doc/architecture-decisions/0008-controller-specs.txt for why and how

### Questions? Join Slack!

We highly recommend that you join us in slack https://rubyforgood.herokuapp.com/ #casa channel to ask questions quickly and hear about office hours (currently Tuesday 6-8pm Pacific), stakeholder news, and upcoming new issues.

farrelld09:
Claim

sandfortw:
@compwron I am confused about something.

The controller spec file was deleted in this PR:

https://github.com/rubyforgood/casa/pull/3432/

But then reintroduced here:

https://github.com/rubyforgood/casa/pull/3817

Was this an error? Can I safely delete it? Are we keeping some controller tests for some reason?

sandfortw:
Only application_controller_spec, static_controller_spec, and sessions_controller_spec do not already have corresponding request spec files. Is there anything left to do for this issue except remove the redundant controller specs? 

Is the task that's left to make sure everything that is being tested in the current controller tests is also in the request tests, or are all of the existing controller tests supposed to remain, in which case this issue should be closed?

If I am fundamentally missing something, I may need to have this explained in office hours, or by someone who is willing to chat with me tomorrow or Friday.

PR #4447, opened 2023-01-14
Move specs in volunteers_controller_spec to volunteers_spec

LeGorge:
Part of #4129
Not closing yet as there will be more on the way.

### What changed, and why?
The `volunteers_controller_spec` file was removed and all of its specs moved to `volunteers_spec`
 - Most of it was already being tested in other places, so I only added whatever was missing
   -  line 32: already tested between volunteer_policy_spec & application_controller_spec
   - line 42: already tested in volunteers_spec at `describe "PATCH /deactivate"`

 - I took the liberty of improving the volunteers edit action:
   - by adding a condition to an assignment as the same condition gates its usage in the view and there's no need to always query the DB for it.

 - no effect

 - does not apply

### Screenshots please :)
 - does not apply

 - I would like to see the request specs renamed to a format including the name `request`, for example:
   - `volunteers_spec` to `volunteers_request_spec`

That would follow rspec's standard naming convention and avoid a wrong connection to `volunteer_spec` which is the model one.
I'm not sure if that's already planned, because some files are named that way, but I chose to follow the instructions from here: https://github.com/rubyforgood/casa/blob/main/doc/architecture-decisions/0008-controller-specs.txt


app/controllers/volunteers_controller.rb
```diff
@@ -46,7 +46,7 @@ def create
 
   def edit
     authorize @volunteer
-    @supervisors = policy_scope current_organization.supervisors.active
+    @supervisors = policy_scope current_organization.supervisors.active unless @volunteer.has_supervisor?
```
compwron, line 49: I don't think this functionality change belongs in a test-only change

LeGorge: Not really a functionality change, the same condition added here is in the view as explained in the description... But removed.

PR #4485, opened 2023-01-27
Refactor court_dates_controller_specs to request specs

LeGorge:
Part of #4129
Not closing yet as there will be more on the way.

### What changed, and why?
The court_dates_controller_spec file was removed and its specs moved to court_dates_spec.
Notes:

 - Different authorization contexts present in the controller test were not being tested in `court_date_policy_spec` so I added them there also.
 - Added a couple of missing expectations to the logic present in the endpoint as well.

- no effect

- does not apply

### Screenshots please :)
- does not apply

The destroy action for court dates controller is allowed to any supervisor or admin, not just the ones from the same org as the case. All the other actions follow a different rule (checking for same org as the case).
I found that a bit odd, maybe it's by design, but it doesn't hurt to point it out.


compwron:
master CI is broken right now, this PR is ok to merge but I want master green first.

compwron (approved):

LeGorge:
Seems like a flickering spec. I guess I can't rerun it.

anshu1992:
It's a lint error. It can be fixed by running `bundle exec standardrb --fix` on your local.

LeGorge:
@anshu1992 The flickering test I mentioned was from a previous run.

That lint problem is not from the files I changed.

anshu1992:
Ah, okay, my bad!

compwron:
:)

PR #4538, opened 2023-02-12
Move casa_org_controller_spec to request_spec

LeGorge:
Part of #4129

### What changed, and why?
 - Refactored the controller spec to request spec for the `CasaOrgController`. Most of it was already being done in the request spec.
 - The authorization aspect of the specs was already being done in policy specs.
 - Some specs had validation tests on them it, so I moved that part to the model spec.

- no effect

 - does not apply

### Screenshots please :)
 - does not apply


compwron (approved):
:)

PR #4595, opened 2023-02-24
Scope CasaCasesController actions by organization

### LINKED ISSUE #4063: Limit casa case controller method permissions to the case's organization
- volunteers
 - supervisors
 - admins

**Description**
Make sure all the methods in [`app/policies/casa_case_policy.rb `](https://github.com/rubyforgood/casa/blob/main/app/policies/casa_case_policy.rb ) contain some logic to prevent casa case access from users belonging to a different organization. 

Fix / Update tests if needed

See [`app/policies/application_policy.rb`](https://github.com/rubyforgood/casa/blob/main/app/policies/application_policy.rb) for helper methods
See https://github.com/varvet/pundit

_QA Login Details:_  

Login Emails: 

### Questions? Join Slack!

crespire:
Resolves #4063 

### What changed, and why?
We changed the behaviour of ApplicationPolicy's `same_org?` method to check for a variety of record object types. Previously, the method assumed that a record always had a `casa_org` association. While this was true for some types of records, other types of records needed other ways to check the `casa_org`.

Additionally, to maintain the ability to `authorize Model`, the method returns `true` in the case of record being a model. This effectively means that the authorization is purely based on a user's role (volunteer, supervisor or admin). In some cases, this is necessary because there is no model instance to verify against. As an example, the `CasaCasesController#index` method effectively only tests for a user's role. However, because the index method uses a `policy_scope`, we feel this is acceptable.

Finally, the previous approach also always authorized actions when record was `nil`. This seemed like a bad idea, so the behaviour has been updated to deny `nil` record authorizations by default.

These changes then necessitated changes to `CasaCasePolicy`.

There are two things we'd like to call out as part of this PR:
1. There is a TODO comment on line 120 of the `CasaCasePolicy` file related to Supervisors and the `new` Casa Case action.
2. In `CasaCasePolicy` on line 103, the `save_emancipation?` policy is available for every user role as part of the same organization.
  a. Should any user type be authorized?
  b. Should volunteers have to be assigned to the case? Currently, any volunteer can make changes.

- Volunteer permissions: Should be scoped to the organization.
- Supervisor permissions: Should be scoped to the organization.
- Admin permissions: Should be scoped to the organization.

`CaseCasePolicy` has additional test cases added. Various tests had their expectations updated due to the change in the `set_casa_case` method.


app/controllers/casa_cases_controller.rb
```diff
@@ -140,12 +140,9 @@ def copy_court_orders
   def set_casa_case
     @casa_case = current_organization.casa_cases.friendly.find(params[:id])
   rescue ActiveRecord::RecordNotFound
-    if params[:action] != "show"
```
dpaola2, line 143: Why is this removed?

app/policies/casa_case_policy.rb
```diff
@@ -31,38 +31,37 @@ def sibling_cases
...
-  alias_method :update_judge?, :admin_or_supervisor_or_volunteer?
-  alias_method :update_court_report_due_date?, :admin_or_supervisor_or_volunteer?
-  alias_method :update_court_orders?, :admin_or_supervisor_or_volunteer?
+  alias_method :update_case_number?, :is_admin_same_org?
+  alias_method :update_case_status?, :is_admin_same_org?
+  alias_method :update_court_date?, :admin_or_supervisor_or_volunteer_same_org?
+  alias_method :update_hearing_type?, :admin_or_supervisor_or_volunteer_same_org?
+  alias_method :update_judge?, :admin_or_supervisor_or_volunteer_same_org?
+  alias_method :update_court_report_due_date?, :admin_or_supervisor_or_volunteer_same_org?
+  alias_method :update_court_orders?, :admin_or_supervisor_or_volunteer_same_org?
```
dpaola2, line 64: I wonder if using `alias_method` here is the right approach. If it is, we should probably do it for most of the calls that simply return the value of another method call. If not, we shouldn't use it at all.

app/policies/casa_case_policy.rb
```diff
@@ -89,35 +88,44 @@ def permitted_attributes
...
   end
 
   def index?
-    admin_or_supervisor_or_volunteer?
+    admin_or_supervisor_or_volunteer_same_org?
   end
 
   alias_method :show?, :same_org_supervisor_admin_or_assigned?
-  alias_method :save_emancipation?, :index?
+  alias_method :save_emancipation?, :index? # Should this be the same as edit?
```
dpaola2, line 103: I think only @compwron will know the answer to this (and if not, we should just approve it)

app/policies/casa_case_policy.rb
```diff
@@ -89,35 +88,44 @@ def permitted_attributes
...
   alias_method :new?, :same_org_supervisor_admin?
   alias_method :create?, :same_org_supervisor_admin?
   alias_method :destroy?, :same_org_supervisor_admin?
 
+  # TODO:
+  # The dashboard policy only shows the "New Case" button to admins, but here
+  # we allow supervisors to also access new. View spec specifically tests the
+  # casa_cases/new.html.erb view against supervisors.
+  #
+  # Should supervisors be able to create cases?
```
dpaola2, line 115: Good question for @compwron or @FireLemons

app/policies/casa_case_policy.rb
```diff
@@ -89,35 +88,44 @@ def permitted_attributes
...
+
   private
 
+  # This method is no longer used, candidate for removal.
```
dpaola2, line 119: Instead of commenting, we should remove it and see if the tests / QA pass.

notes.md
```diff
@@ -0,0 +1,12 @@
+TDD these changes.
```
dpaola2, line 1: This file should be removed

spec/policies/casa_case_policy_spec.rb
```diff
@@ -244,29 +402,57 @@
...
+      end
     end
 
+    context "when not part of the same organization" do
+      it "does not allow admin to create" do
+        is_expected.not_to permit(other_org_casa_admin, casa_case)
+      end
+    end
+
+    # TODO: What can supervisors do?
```
dpaola2, line 432: Open question we might need answered at some point by @compwron or @FireLemons

dpaola2 (changes requested):
Small tweaks, and some open questions. Very close!

app/controllers/casa_cases_controller.rb
crespire: I discussed with Shen about this during office hours. It seemed like the idea was to return a 404 code, but it would also return a blank body. It was decided that the intention was to not allow this kind of cross-org access, so we decided to redirect to the index in all cases.

I think this exception for `show` was for the JSON request, which we now handle via `respond_to { format.json }` instead.

app/policies/casa_case_policy.rb
crespire: Thanks for the note! This method is actually no longer used as a result of the changes we made, and the suite passes so we can definitely remove the method.

app/controllers/casa_cases_controller.rb
dpaola2: Gotcha!

app/policies/casa_case_policy.rb
GALTdea: Hi Dave, Just a quick clarification, what do you mean by "the calls that simply return the value of another method call"? Not to return values of methods that in turn return values from other methods?

GALTdea:
Also, is it a good idea to have Pundit policies/rule documentation? Something detailing what uses should have access to what records/models/resources?

dpaola2 (approved):

Issue #4594, opened 2023-02-24
New export report for volunteer email changes


compwron:
**What type(s) of user does this feature affect?**

 - supervisors
 - admins

**Description**
This ticket depends on [TODO make ticket] which allows volunteers to change (and confirm!) their own emails. Export a report of volunteers who have changed their emails
As always, only export volunteers for the supervisor's current CASA
columns: `old email(s), new email`

**Screenshots of current behavior, if any**  
https://casa-qa.herokuapp.com/reports
<img width="1080" alt="Screenshot 2023-02-24 at 11 33 02 AM" src="https://user-images.githubusercontent.com/578159/221274151-075f0369-b32d-4ee5-9638-f686198dca39.png">

_Login Details:_  
[Link to QA site](https://casa-qa.herokuapp.com/)  
[Link to new theme site](https://casa-new-theme.herokuapp.com/languages/new)  

Login Emails: 
- volunteer1@example.com  view site as a volunteer
- supervisor1@example.com view site as a supervisor
- casa_admin1@example.com view site as an admin
- all_casa_admin1@example.com view site as an all casa admin
  - go to `/all_casa_admins/sign_in`  

password for all users: 12345678  

### Questions? Join Slack!

We highly recommend that you join us in slack https://rubyforgood.herokuapp.com/ #casa channel to ask questions quickly and hear about office hours (currently Tuesday 5-7pm Pacific), stakeholder news, and upcoming new issues.

ShamiTomita:
Hi, can I work on this?

ShamiTomita:
Also, is this ticket related to #4534 and #4476?

ShamiTomita:
One more question, for the confirmation aspect of the email change, do y’all just need an additional “email confirmation” input in the form or do y’all need a confirmation email to be sent?

dpaola2:
I would suggest that having a confirmation field is probably sufficient. @compwron can chime in if she disagrees, though. I think asking the user to enter the email twice will sufficiently prevent typos.

ShamiTomita:
Awesome, followup question, would you want these email updates/confirmations extracted out of the user_params and made into their own functions in the user controller? Like how passwords are treated?

ShamiTomita:
Perfect, thank you!

ShamiTomita:
Hi still working on the related issue! I’ll try to get a working PR for that by early next week!

ShamiTomita:
almost ready to start!

[...]

PR #4734, opened 2023-04-04
Email changes for users

### LINKED ISSUE #4602: Email Updates For Volunteers
 - volunteers!!! Definitely
 - supervisors? Maybe?? (please let me know if this feature should affect supervisors as well)

**Description**
Part of #4594. In order to include previous email changes for Volunteers, I've identified the following tasks: 
1. Record all the previous emails used by the User(Volunteer) 
2. Allow Volunteers to change and then confirm their emails. 

I will be submitting a PR to address these as part of issue #4594
This is my first time creating an issue for myself so any guidance or reviews would be greatly appreciated.

_Login Details:_  
[Link to new theme site](https://casa-new-theme.herokuapp.com/languages/new)  

Login Emails: 

### Questions? Join Slack!

ShamiTomita:

Resolves #4602 as part of #4594 
These changes:
1. Allow Users (volunteers/supervisors/casa admins) to change and confirm their own emails 
2. Notify user's old email that their email changed AFTER user confirms email changed (mirroring password changed reminders)
3. Record all previously used emails in a new array attribute for Users

### What changed, and why?
- [X] Added Confirmable and Reconfirmable to User Model to leverage Devise's methods and emails
- [x] Created users#record_previous_email and added it to a before_update hook to record changes before they are persisted, as well as hooks to control when email notifications are sent.
- [X] Added an old_emails array attribute for users to track previously used emails
- [X] Created a users#filter_old_emails! method to update the old_emails attribute after every email change
- [x] Extracted emails out of user_params into its own email_params (mirroring passwords)
- [x] Added a new set of actions to update User emails users#update_emails and (private) users#email_params & users#update_user_email

- Volunteer permissions: Allowed to change and confirm own email
- Supervisor permissions: Allowed to change and confirm own email
- Admin permissions: Allowed to confirm own email

system/users/edit_spec.rb -> test that originally made sure email is disabled for volunteers/supervisors changed to a test to ensure all users can successfully change their email. Tests checking for errors also included for all three user roles. 

models/user_spec.rb -> added a test for the new #record_previous_email method as well a test to ensure users start with an empty "old_emails" array attribute

requests/users_spec.rb -> added in the update_email method to be tested for all three types of users

### Screenshots please :)
Screenshot 2023-04-05 at 4 56 45 PM

### Feelings gif (optional)
How to embed:
`alt text`


ShamiTomita:
The codeclimate failures are mostly for the email confirmation javascript file which I basically lifted from the password confirmation file. I'm not sure how much of an issue that actually is though.

app/models/user.rb
```diff
@@ -138,13 +143,47 @@ def active_for_authentication?
...
+    true
+  end
+
+  def send_email_changed_notification?
+    # Don't send the notification if the skip flag is set
+    return false if skip_email_changed_notification
+
+    # Otherwise, let Devise send the notification
+    super
+  end
```
cnorm35, line 172: I'm sure if I follow what's happening here.  `skip_email_changed_notificaton` is hard-coded to always return true, so the conditional will always read as: "return false if true"

From what I can tell, I don't think this would ever make it to the `super` call

db/migrate/20230327154626_add_confirmable_to_users.rb
```diff
@@ -0,0 +1,22 @@
...
+  def up
+    add_column :users, :confirmation_token, :string
+    add_column :users, :confirmed_at, :datetime
+    add_column :users, :confirmation_sent_at, :datetime
+    add_column :users, :unconfirmed_email, :string # Only if using reconfirmable
+    add_index :users, :confirmation_token, unique: true, algorithm: :concurrently
+    # User.reset_column_information # Need for some types of updates, but not for update_all.
+    # To avoid a short time window between running the migration and updating all existing
+    # users as confirmed, do the following
+    User.update_all confirmed_at: DateTime.now
```
cnorm35, line 13: I think this should be moved into it's own AfterParty task.  This would overwrite every User's confirmed_at value if this migration is ever ran more than once.

The Post-deployment tasks section of the README has some examples on how to create a post deployment task

app/controllers/casa_admins_controller.rb
```diff
@@ -17,7 +17,11 @@ def edit
   def update
     authorize @casa_admin
 
+    @casa_admin.skip_casa_admin_email_changes
```
cnorm35, line 20: see above comment about method name changes

app/controllers/supervisors_controller.rb
```diff
@@ -55,7 +55,11 @@ def edit
 
   def update
     authorize @supervisor
+    @supervisor.skip_casa_admin_email_changes
```
cnorm35, line 58: see above comment about combining these lines into a single method

cnorm35 (changes requested):
Great job with this @ShamiTomita ! This was a big one.  Made a few comments for some things to update and added a couple of questions.

Let me know if you have any questions on these suggestions or changes, happy to walk through everything with you.

ShamiTomita:
@cnorm35 Thank you for the review and the suggestions! I’m gonna go over them tonight and I would love to go over some of them later. Thank you!

app/controllers/all_casa_admins/casa_admins_controller.rb
```diff
@@ -22,7 +22,10 @@ def edit
 
   def update
     @casa_admin = CasaAdmin.find(params[:id])
+    @casa_admin.skip_casa_admin_email_changes
     if @casa_admin.update(all_casa_admin_params)
+      updated_emails = @casa_admin.old_emails.reject { |old| old == @casa_admin.email }
+      @casa_admin.update(old_emails: updated_emails)
```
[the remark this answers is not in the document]
ShamiTomita: Just added an instance method in the user class to replace these lines. Writing a test for it was way easier too :-)

app/models/user.rb
```diff
@@ -138,13 +143,47 @@ def active_for_authentication?
...
+
+  def skip_confirmable_email_confirmation_upon_creation
+    skip_confirmation!
+    confirm
+  end
+
+  def skip_casa_admin_email_changes
+    skip_reconfirmation!
+  end
```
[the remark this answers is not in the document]
ShamiTomita: Honestly upon reflection, there's no real benefit and I think I was on an instance writing roll. It definitely adds clutter where there shouldn't be. I've included the Devise method directly to the update methods.

app/models/user.rb
```diff
@@ -5,10 +5,15 @@ class User < ApplicationRecord
   include Roles
   include ByOrganizationScope
   include DateHelper
+  include ActiveModel::Dirty
```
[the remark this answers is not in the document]
ShamiTomita: Oh that's good to know! I had included it because, from the way I read the docs, I thought it suggested it wasn't already available. Would you recommended double checking if a module is already included by doing something like: 

```ruby
ApplicationRecord.included_modules.include?(ActiveModel::Dirty)
```

app/models/user.rb
ShamiTomita: These methods were definitely not my best. I was struggling to get the proper user story for the confirmation flow: User submits email change -> confirmation email sent -> User confirms -> Old user email receives notification of email changed 

But I kept getting this type of flow:
User submits email change -> confirmation email sent & email changed notification sent to old email  -> User confirms (no email changed notification sent to old email after confirmation) 

I was confused because, after reading the Devise docs, I thought that final, post-confirmation email changed notification would automatically send but didn't. I ended up using ChatGPT to see if it could help and this method it generated seemed to work. I definitely don't have a solid understanding of ```super``` (this is a great excuse to jump into it more) and after doing some checks I trimmed what seemed to be the useless bits into only these two relevant methods:

```
def send_email_changed_notification #hardcoded in order to prevent the email changed notification BEFORE confirm
    return false 
  end

  def after_confirmation #sends the email changed notification AFTER user confirms
    send_email_changed_notification
  end
```

Having these two set up like this gets the user story I want but still seems kind of weird. I'm pretty sure they're related to ```Devise.config.send_email_changed_notification = true``` but would love to go over this with you.

app/controllers/all_casa_admins/casa_admins_controller.rb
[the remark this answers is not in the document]
ShamiTomita:
```
 def filter_old_emails(previous_email)
      updated_emails = self.old_emails.reject { |old| old == previous_email }
      self.update(old_emails: updated_emails)
  end 
```

db/migrate/20230327154626_add_confirmable_to_users.rb
ShamiTomita: I see, that makes sense. I rolled back the migration, edited the file, and moved this into its own after party task.

ShamiTomita (commented):
Ready for review!

app/models/user.rb
```diff
@@ -138,13 +141,40 @@ def active_for_authentication?
...
     actively_assigned_and_active_cases.is_transitioned.any?
   end
+
+  def record_previous_email
+    if email_changed? && !old_emails.include?(email_was)
+      old_emails.push(email_was)
+    end
+  end
+
+  def filter_old_emails(previous_email)
```
cnorm35, line 151: my only nitpick is that i think this method name would be clearer if you added a bang at the end of it.  `filter_old_emails!` 

seeing how it's used in the files above, i wasn't sure if it was supposed to modify the objects or return them after filtering.

cnorm35 (changes requested):
1 small change request for clarity (see comment) but everything else is looking good!

cnorm35 (approved):
Looks Good To Me @ShamiTomita

dpaola2 (approved):

PR #4785, opened 2023-04-26
Update API.md

xihai01:
Added text - missing a REST tag.

### What changed, and why?
Just a minor update to API.md.
I noticed the first resource was missing a REST tag.
So I added POST because it's creating a casa case.


littleforest (approved):
Thanks for the fix!

PR #4800, opened 2023-05-05
Fix for placement test date format bug #4794

DariBerrie:
### What github issue is this PR for, if any?
Resolves #4794

### What changed, and why?
Just making the changes suggested by @littleforest! I made the updates to both the court date and the placement date checks.

PR #4802, opened 2023-05-05
Reinstate spec that was commented out

littleforest:
Reinstate spec that was commented out in #4800


littleforest:
FYI @DariBerrie

compwron (approved):
:)

PR #4818, opened 2023-05-13
Update postgres version in GitHub Actions to match production

littleforest:
Screen Shot 2023-05-13 at 2 51 39 PM


compwron (approved):
:)

PR #4852, opened 2023-05-31
Recreate db diagram 

FireLemons:
### What changed, and why?
Recreating the db diagram so we own it this time. The existing database diagram was created using a proprietary tool that dropped support for the format we wanted.


compwron (approved):
:)

johnpaulashenfelter:
@FireLemons did you consider automating this with https://github.com/voormedia/rails-erd? It speaks mermaid.js now which makes it even easier to hand edit what gets generated.

FireLemons:
@johnpaulashenfelter There's automation that makes it look like that already. I want it to look more like the current one with the colors. After it's done the first time we only need to edit it a little bit after.

PR #4854, opened 2023-05-31
Update README to reflect current Ruby version

bacchist:
### What changed, and why?

Changed the Ruby version in the setup instructions from 3.1.2 to 3.2.2 to match `.ruby-version`.


littleforest (approved):
Thanks!

PR #4932, opened 2023-07-01
update sweetalert2

compwron:
https://github.com/sweetalert2/sweetalert2/blob/main/CHANGELOG.md


FireLemons:
Does dependabot not do this for us?

PR #4951, opened 2023-07-03
Make test_checker.rake respect request tests for controllers

### LINKED ISSUE #4917: Make test_checker.rake respect request tests for controllers
Make https://github.com/rubyforgood/casa/blob/main/lib/tasks/test_checker.rake respect request tests for controllers
As in the test for https://github.com/rubyforgood/casa/blob/main/app/controllers/casa_org_controller.rb is https://github.com/rubyforgood/casa/blob/main/spec/requests/casa_org_spec.rb but https://github.com/rubyforgood/casa/blob/main/.allow_skipping_tests#L11 ignores the controller because it does not pick up the test.

sandfortw:
Resolves #4917 

### What changed, and why?
The spec_files method, in the test_checker.rake file:
image

Will be changed to this:
image

This allows the test_checker to look for controller tests in the spec/requests directory. Previously, it was assumed that all controller tests would be in a spec/controllers directory. Because of this, the test_checker was not recognizing tested controllers. 

Now, if a spec file is in a 'requests' directory, the path string will be manipulated to match a corresponding controller file, so that the missing spec files method (below) can continue to work as designed:

image

Additionally, I changed the name of a file that was not following the naming conventions for requests, and therefore not being picked up by the test_checker. 

This should not affect user permissions.

This was tested manually by entering `rake test_checker` into the terminal. 

### Feelings gif (optional)
How to embed:
alt text


sandfortw:
Just to be clear, this is my first PR for this repo--I hope this solution makes sense and solves the problem, but extra scrutiny is probably warranted.

compwron (approved):
nice! :)

PR #4956, opened 2023-07-05
try to fix docker

sandfortw:
This commit will try to fix docker. We ran 'bundle install --full-index' and the gemfile.lock was changed.

### What changed, and why?
This updates the Gemfile.lock file, with an updated version of rack-attack

CI


FireLemons:
:1st_place_medal:

PR #5090, opened 2023-08-04
5070 - Update banner creation UX

### LINKED ISSUE #5070: Update banner creation UX
 - admins
 - all casa admins

## Description
When an org has an active banner and wants to add a new one, the user has to create the new banner as inactive and deactivate the old banner before activating the new one.

Proposed changes:
- When a user has a current active banner and checks `Active` in creating a new banner, a warning is displayed that all currently active banners will be deactivated.
- Deactivate currently active banners before creating a new active banner.

image

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

carrollsa:
Resolves #5070

First time using Stimulus. Open to any suggestions, and I'm fairly new at most of this!

### What changed, and why?
Adds 2 UX improvements:
- Displays warning message that alternate active banner will be deactivated upon checking `Active?` on a new or edited banner form if alternate active banner exists.
- Deactivates alternate active banner upon submission if new or edited banner has `Active?` checked.

- Volunteer permissions: No effect
- Supervisor permissions: No effect
- Admin permissions: No effect

- System spec tests for deactivation of alternate active banner
- View tests for conditional render of red warning
- Helper tests for method to determine initial hidden status for warning message

### Screenshots please :)
image

### Feelings gif (optional)
Not a gif, but sort of this guy because I created this issue, and this felt like a lot of conditional stuff to add for a fairly minor improvement.
image


app/controllers/banners_controller.rb
```diff
@@ -11,18 +11,22 @@ def new
...
     @banner = current_organization.banners.find(params[:id])
+    @has_alternate_active_banner = current_organization.banners.where(active: true).where.not(id: @banner.id).exists?
   end
 
   def create
     authorize :application, :admin_or_supervisor?
 
     @banner = current_organization.banners.build(banner_params)
+    deactivate_alternate_active_banner if banner_params[:active]&.to_i == 1
+
```
carrollsa, line 29: Wanted to avoid repeating this in two methods and tried with a `:before_action, only: [:create, :edit]`, but didn't know how to ensure we had an `@banner` to work with.

app/views/banners/_form.html.erb
```diff
@@ -21,9 +21,10 @@
       <%= form.label :name %>
       <%= form.text_field :name, required: true %>
     </div>
-    <div class="form-check checkbox-style mb-20">
-      <%= form.check_box :active, class: 'form-check-input' %>
+    <div data-controller="checkbox reveal" data-action="<%= @has_alternate_active_banner && 'checkbox:toggle->reveal#toggle'%>" data-reveal-hidden-class="d-none" class="form-check checkbox-style mb-20">
+      <%= form.check_box :active, data: { action: "click->checkbox#toggle", checkbox_target: "checkbox" }, class: 'form-check-input' %>
       <%= form.label :active, "Active?", class: 'form-check-label' %>
+      <span data-reveal-target="item" class="text-danger<%= form.object.active && @has_alternate_active_banner ? '' : ' d-none' %>">Warning: This will replace your current active banner</span>
```
carrollsa, line 27: I don't love a ternary that can return `''` but thought it read a little better than `(!form.object.active || !@has_alternate_active_banner) && ' d-none'`. Any opinion?

app/models/banner.rb
```diff
@@ -6,17 +6,8 @@ class Banner < ApplicationRecord
   scope :active, -> { where(active: true) }
 
   validates_presence_of :name
-  validate :only_one_banner_is_active_per_organization
 
   private
-
-  def only_one_banner_is_active_per_organization
```
kcdragon, line 13: Even though we will no longer get this validation message in the UI, I think we should still keep it in. For example, a developer could try to update a banner to active in `rails console` and we'd want that to fail if there is already one active.

app/controllers/banners_controller.rb
kcdragon: I think its fine to double up the code. Create and update often have similar code.

app/models/banner.rb
kcdragon: There's also a race condition where when the page loads, there is no active banner, but by the time the form is submitted, another user has set a banner to be active.

app/controllers/banners_controller.rb
```diff
@@ -53,4 +59,9 @@ def destroy
   def banner_params
     params.require(:banner).permit(:active, :content, :name).merge(user: current_user)
   end
+
+  def deactivate_alternate_active_banner
+    alternate_active_banner = current_organization.banners.where(active: true).where.not(id: @banner.id).first
+    alternate_active_banner&.update(active: false)
```
kcdragon, line 65: I don't think you need to change anything but I want to point something out for you future benefit.

If the call to `update` fails due to some validation error (unlikely, but there are scenarios where the record could get into a state where that happens) then we will be in a state where `alternate_active_banner` is active and we're also trying to make `@banner` active. Usually, you'd want to handle the possibility of this returning `false`. However I think its fine to ignore it in this case as long as we add back in the `only_one_banner_is_active_per_organization` validation since that will catch the issue.

app/javascript/controllers/checkbox_controller.js
```diff
@@ -0,0 +1,15 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  static targets = ['checkbox']
+  static values = { checked: Boolean }
+
+  connect () {
+    this.checked = this.hasCheckboxTarget && this.checkboxTarget.checked
```
kcdragon, line 8: What's the benefit of maintaining this state? I think you could look at `this.checkboxTarget.checked` inside of `toggle` and not have to worry about maintaining this state.

app/views/banners/_form.html.erb
```diff
@@ -21,9 +21,10 @@
       <%= form.label :name %>
       <%= form.text_field :name, required: true %>
     </div>
-    <div class="form-check checkbox-style mb-20">
-      <%= form.check_box :active, class: 'form-check-input' %>
+    <div data-controller="checkbox reveal" data-action="<%= @has_alternate_active_banner && 'checkbox:toggle->reveal#toggle' %>" data-reveal-hidden-class="d-none" class="form-check checkbox-style mb-20">
+      <%= form.check_box :active, data: { action: "click->checkbox#toggle", checkbox_target: "checkbox" }, class: 'form-check-input' %>
       <%= form.label :active, "Active?", class: 'form-check-label' %>
+      <span data-reveal-target="item" class="text-danger<%= form.object.active && @has_alternate_active_banner ? '' : ' d-none' %>">Warning: This will replace your current active banner</span>
```
kcdragon, line 27: Could you elaborate more on what the reveal controller does? My thought is that an alternative is to add a `hidden` css class to determine if the error message is shown or hidden. Does the reveal controller do more than that?

kcdragon (commented):
Nice work! I have a couple comments/questions

app/controllers/banners_controller.rb
carrollsa: Ah, thanks, that makes sense. Was on the fence about removing that validation but couldn't think of when it would be necessary. Here's a good reason!

app/javascript/controllers/checkbox_controller.js
carrollsa: It appears to be unnecessary! Thanks!

app/models/banner.rb
carrollsa: Ah, thanks, that makes sense. Was on the fence about removing that validation but couldn't think of when it would be necessary. Here's a good reason!

app/views/banners/_form.html.erb
carrollsa: As far as I know, it does not anything outside of hide/show/toggle. My initial impression was to do this manually, but I gathered `stimulus-components` was a pretty standard add.

app/javascript/controllers/checkbox_controller.js
schoork: Might be worth trying to do this with just the stimulus-reveal-controller. Feels like you could put the action directly on the checkbox.

carrollsa: @schoork Good catch! I'm pretty sure I can remove `checkbox_controller` since the logic for when to show/hide ended up living outside of it.

spec/system/banners/new_spec.rb
carrollsa, line 1: I renamed this to follow the convention I was seeing.

spec/views/banners/new.html.erb_spec.rb
carrollsa, line 1: I'd like some input on these tests. A few questions:
- Are these necessary? I figured they couldn't hurt, but I don't want to bloat the test suite.
- Should I pretty much copy these into an `edit.html.erb_spec.rb` file? I have one written, but I wasn't sure if that was necessary, given both `new` and `edit` render the same form.
- I didn't seem to have access to the same actions like `click_on`, `fill_in`, etc. within the `views` specs, so I'm passing a banner in with the `active` field set already to what we want. This doesn't exactly text whether the view is responsive, but it's at least checking the conditions to show/hide the warning. Is there a way I can improve this?

carrollsa:
Anybody know what might be going on with the docker check?

app/controllers/banners_controller.rb
```diff
@@ -34,6 +38,8 @@ def update
     authorize :application, :admin_or_supervisor?
 
     @banner = current_organization.banners.find(params[:id])
+    deactivate_alternate_active_banner if banner_params[:active]&.to_i == 1
```
littleforest, line 41: I think you could move the `if banner_params[:active]&.to_i == 1` to the `deactivate_alternate_active_banner` method for a little better readability.

Also, `nil.to_i = 0`, so you can get away with: `banner_params[:active].to_i == 1`

app/views/banners/_form.html.erb
littleforest: What do you think about moving it to a helper method? Something like:
```
class="text-danger <%= alternate_banner_class(form.object.active) %>"
```

```
def alternate_banner_class(active)
  unless active && @has_alternate_active_banner
    "d-none"
  end
end
```

app/controllers/banners_controller.rb
littleforest: Yes, ideally you would want to do this in a transaction so that both save/updates will be rolled back if one of them fails.

You could do:

```
  def create
    authorize :application, :admin_or_supervisor?

    @banner = current_organization.banners.build(banner_params)

    Banner.transaction do
      deactivate_alternate_active_banner # and make sure to use `update!` in this method
      @banner.save!
    end

    redirect_to banners_path
  rescue
    render :new
  end
```

and similar for the update method.

app/controllers/banners_controller.rb
carrollsa: Ah, I like this. Thanks!

app/views/banners/_form.html.erb
carrollsa: Bingo! I like that much more. Thanks!

app/controllers/banners_controller.rb
```diff
@@ -11,34 +11,44 @@ def new
...
-      render :edit
+
+    Banner.transaction do
+      deactivate_alternate_active_banner
+      @banner.update!(banner_params)
     end
+
+    redirect_to banners_path
+  rescue
+    render :new
```
littleforest, line 51:
```suggestion
    render :edit
```

app/views/banners/_form.html.erb
```diff
@@ -21,9 +21,10 @@
       <%= form.label :name %>
       <%= form.text_field :name, required: true %>
     </div>
-    <div class="form-check checkbox-style mb-20">
-      <%= form.check_box :active, class: 'form-check-input' %>
+    <div data-controller="reveal" data-reveal-hidden-class="d-none" class="form-check checkbox-style mb-20">
+      <%= form.check_box :active, data: { action: @org_has_alternate_active_banner && 'click->reveal#toggle' }, class: 'form-check-input' %>
       <%= form.label :active, "Active?", class: 'form-check-label' %>
+      <span data-reveal-target="item" class="text-danger <%= conditionally_add_hidden_class(form.object.active) %>">Warning: This will replace your current active banner</span>
```
littleforest, line 27: Would it be better to actually use a Stimulus class here? https://stimulus.hotwired.dev/reference/css-classes

app/controllers/banners_controller.rb
carrollsa: Whoops, good catch!

app/controllers/banners_controller.rb
```diff
@@ -11,34 +11,44 @@ def new
     authorize :application, :admin_or_supervisor?
 
     @banner = Banner.new
+    @org_has_alternate_active_banner = current_organization.banners.where(active: true).where.not(id: @banner.id).exists?
```
littleforest, line 14: You could also consider making this a method:

```
@org_has_alternate_active_banner = current_organization.has_active_banner?(@banner.id)
```

```
# app/models/casa_org

class CasaOrg
  ...

  def has_active_banner?(current_banner_id)
    banners.where(active: true).where.not(id: current_banner_id).exists?
  end
end
```

app/views/banners/_form.html.erb
carrollsa: Ah, might be! Will be out of town this weekend, but will look into this early next week.

app/controllers/banners_controller.rb
littleforest: Also if you make it a method, you could probably get away without passing the instance variable around, and just call the method on line 25 of `app/views/banners/_form.html.erb`

app/views/banners/_form.html.erb
carrollsa: I believe we already do make use of a stimulus class the parent `div` element with `<div ... data-reveal-hidden-class="d-none" ...>`, and this pattern is what was given in the `stimulus-reveal-controller` documentation.

I'm still solidifying my mental model for how everything works with stimulus, but I'm not 100% sure of how to make use of a stimulus class for the above, as we're just setting the initial classes of the span which are then toggled by the stimulus controller.

littleforest: Okay no worries, I'm not familiar with the stimulus-reveal-controller and how it would interact with a Stimulus class. Fine as-is.

carrollsa:
@compwron @littleforest @FireLemons 

I think this one sort of fell below the pack because it was a draft for so long. Would any of you have time to take a look at this?

littleforest:
@carrollsa we have actually been talking with the stakeholders about allowing multiple active banners 😬 . They are interested in having an admin-wide banner that shows up to everyone, and allowing smaller supervisor-level banners that would just show to a supervisor's volunteers. You've put in a lot of work on this PR, so I hate to put the kibosh on it, but I'm worried we are going to want to modify this functionality really soon.

carrollsa:
@littleforest that's a bummer, but I understand. Thanks for letting me know!

FireLemons:
:1st_place_medal:

PR #5104, opened 2023-08-09
Delete unused languages_users table from db

FireLemons:
### What changed, and why?
Deleted `languages_users` table from db because I suspect `user_languages` is used instead.

Adding a language to a user only makes a row in the latter table
image
 image

Do not merge until we can confirm the prod version of `languages_users` has no rows


FireLemons:
The prod data says the `languages_users` table does have rows but the `user_languages` contains all the rows the `languages_users` table has with additional rows from later in time. 

See: https://github.com/rubyforgood/casa/commit/43bc4e931b522d3fc6cc649d12102ee001b10879

I think it's safe to drop the unused table

PR #5145, opened 2023-08-23
fixed not being able to select multiple cases after failing the case contact form

FireLemons:
#### First bug
If you go to a new case contact form with get parameters, and if you select multiple cases
image

and fail to fill out the case contact form, you can't select multiple cases anymore
image

#### Second bug
If you navigate to the new case contact form and you have GET parameters and you select cases that don't include the casa case in the GET parameters,
image
it responds as though no cases were selected.
image

Only Manually

### Screenshots please :)
image


littleforest (commented):
@FireLemons would you be able to add a (initially) failing spec for this change?

schoork:
@FireLemons This is closable, right? The fix was the two PRs we did earlier this week?

PR #5151, opened 2023-08-27
Move CaseContactReportsController controller specs to request

LeGorge:
Part of #4129

### What changed, and why?
 - Refactor controller specs of CaseContactReportsController to request specs
   - I have verified that all spec cases on the existing controller spec are already covered on the request spec. So I just deleted the unnecessary controller spec file.

- no effect

 - does not apply

### Screenshots please :)
 - does not apply


FireLemons:
:medal_sports:

Issue #5037, opened 2023-07-29
Meta Issue: Test Suite Cleanup


meg-gutshall:
I ran the test suite on the entire codebase to see what warnings and deprecations would arise.

Also, specs tend to be forgotten once marked pending.

### Environment

The following screenshots are the result of running `bin/rails spec` locally on `main` after cloning down the app.

### Sample Error Output:

<img width="1662" alt="test-suite-warning-1" src="https://github.com/rubyforgood/casa/assets/37842352/9cb9ae0f-7167-425f-b84d-b1ec52b19253">
<img width="1662" alt="test-suite-warning-2" src="https://github.com/rubyforgood/casa/assets/37842352/e5c853ed-8719-42b1-9b8a-98a320915e3b">
<img width="1805" alt="test-suite-warning-3" src="https://github.com/rubyforgood/casa/assets/37842352/4ba2f1c6-45de-4321-9dc7-dcc1bcf01b07">
<img width="1805" alt="test-suite-warning-4" src="https://github.com/rubyforgood/casa/assets/37842352/3f6771e6-896a-4050-8ccf-560ac7a0fbfa">
<img width="1822" alt="test-suite-warning-5" src="https://github.com/rubyforgood/casa/assets/37842352/6b5641c4-3f0c-4678-b9d3-7e54a79ae15d">
<img width="1822" alt="test-suite-warning-6" src="https://github.com/rubyforgood/casa/assets/37842352/283fce0e-a769-48e5-83ca-cfe1b7c270c7">
<img width="1822" alt="test-suite-warning-7" src="https://github.com/rubyforgood/casa/assets/37842352/a9b5202b-9aad-4316-8a25-c8456602b2cc">

### Questions? Join Slack!

We highly recommend that you join us in slack https://rubyforgood.herokuapp.com/ #casa channel to ask questions quickly and hear about office hours (currently Tuesday 5-7pm Pacific), stakeholder news, and upcoming new issues.

meg-gutshall:
@littleforest Do you want to groom this a little?

bcastillo32:
linked PR merged https://github.com/rubyforgood/casa/pull/5053

littleforest:
This actually has a lot of things to work on still. Haven't gotten to them all.

sean-dickinson:
@littleforest I'd be happy to open a PR for some of the low hanging fruit if you like.

littleforest:
@sean-dickinson that would be great, thanks!

[...]

PR #5165, opened 2023-08-29
Fix test suite warnings

sean-dickinson:
References #5037

### What changed, and why?

The test suite is currently outputting a lot of noise relating to some deprecation warnings and changed autoloading behavior. I have addressed the 2 main culprits:

- In the `_active_volunteer_info.html.erb` partial, there was a call to a deprecated method from ActiveSupport that allowed passing an argument to the `to_s` method on various objects. In the Rails 7 release this method was moved to `to_fs` and the overridden `to_s` method was deprecated.
- In the `test_checker.rake` task there were 2 constants defined at the top level of the task. Due to the way the autoloader works, these were causing a ton of warnings due to this file being autoloaded and redefining the constant (rake files aren't namespaced in a traditional way so technically that constant is available more globally than might be apparent). To fix this I changed the constant definitions to method definition inside of the task block. I would have kept them as regular variables, however the linter considered them to be "unnecessary" as they were only called inside of other inline methods in this task block. Realistically I think a better solution would be to move the body of this rake task into a class and then just instantiate and call methods on the class inside the task, but that seemed out of the scope of this issue.

Separately I switched another rake task that is called in the test suite to only print to the screen when the environment is not the test environment as has been done elsewhere in the app.

- N/A

- Neither of these changes are covered by the current test suite. 
- I can probably add a test to ensure the output of the `to_fs` is as expected if you would like, but I might need some help to make sure I fully understand the context of that particular partial. 
- The rake task itself is currently not actually tested, but if you would like testing around it, I think moving the logic to an actual class and testing that would be the best option. If you would like me to go ahead and do that larger refactor so as to decouple the logic from the rake environment I can do that as well as part of this PR or a separate one. 

### Screenshots please :)
Before:
image
image

After: 
image

### DIFF
--- a/app/views/supervisor_mailer/_active_volunteer_info.html.erb
+++ b/app/views/supervisor_mailer/_active_volunteer_info.html.erb
@@ -36,7 +36,7 @@
   <% if recently_unassigned %>
     <br>
     This case was unassigned from <%= volunteer_display_name %> on
-    <%= case_assignment.updated_at.to_date.to_s(:long_ordinal) %>
+    <%= case_assignment.updated_at.to_date.to_fs(:long_ordinal) %>
     <% if successful_contacts + unsuccessful_contacts > 0 %>
       The above activity only describes the part of the week when the case was still assigned to
       <% volunteer_display_name %>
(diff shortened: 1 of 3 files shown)


FireLemons:
:1st_place_medal:

PR #5177, opened 2023-09-02
Update CasaCasesController & CaseAssignmentsController specs to request specs

LeGorge:
Part of #4129

### What changed, and why?
CasaCasesController
 - the controller specs were all covered on the request ones.
 - removed the controller specs
 - added a couple of missing specs to the existing spec endpoints

CaseAssignmentsController
 - most of the controller specs were already covered with a couple of exceptions, which I added.
 - removed the controller specs
 - refactored existing request specs
 - added several missing spectations on the existing specd endpoints
 - added missing specs for `reimbursement` endpoint that was not specd
 - added a couple of skipped specs to point out a bug in the controller code for action `show_hide_contacts`. I added a TODO comment on them explaining the problem

 - no effect

 - does not apply

### Screenshots please :)
 - does not apply


FireLemons:
:medal_military:

PR #5190, opened 2023-09-04
Render Local Time

### LINKED ISSUE #5134: Local Time Display Component
 - volunteers
 - supervisors
 - admins
 - all casa admins

## Description
Displayed time across the app has timezones determined by the server but the user may not have the same time zone as the server resulting in incorrect times. In some cases the time displayed is actually UTC time.

Create a [view component](https://viewcomponent.org/guide/getting-started.html) that renders the local time ~html with a unix timestamp and converts it to the local time on the client via javascript. Similar to [discord timestamps](https://www.pattlesstore.com/articles/how-to-use-timestamps-on-discord)~

The component should take as arguments
 - the time in unix timestamp format
 - the format

The [`title`](https://www.w3schools.com/TAGS/att_title.asp) of the component should show a more specific representation of the time.

~so on the page it will render something like <span><t: UNIX_TIME_VALUE, f: FORMAT ></span> and then javascript will convert the html into something human readable~ it's probably better to get the time zone from the cookies from the browser so no js required

Write Tests

See https://github.com/rubyforgood/casa/wiki/Working-with-Frontend-Assets

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

SajjadAhmad14:
Resolves #5134

### What changed, and why?
We created a new view component named LocalTimeComponent. This component renders user local time.

Component is tested with both positive and negative tests.

### DIFF
--- a/app/components/local_time_component.rb
+++ b/app/components/local_time_component.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class LocalTimeComponent < ViewComponent::Base
+  include ActionController::Cookies
+  attr_reader :format, :unix_timestamp
+
+  def initialize(format:, unix_timestamp:)
+    @format = format
+    @unix_timestamp = unix_timestamp
+  end
+
+  def local_time
+    # Time format should be passed as 12 or 24 only
+    unless [12, 24].include?(format)
+      raise ArgumentError, "Invalid time format argument"
+    end
+
+    time = Time.at(unix_timestamp).in_time_zone(@time_zone)
+    time_format = format == 12 ? "%I:%M %p" : "%H:%M"
+    formatted_date = time.strftime("%B %d, %Y")
+    time_of_day = time.strftime(time_format)
+    time_zone = time.zone
+    return { date: formatted_date, time: time_of_day, zone: time_zone }
+  end
+
+  private
+
+  def before_render
+    browser_tz = ActiveSupport::TimeZone.find_tzinfo(cookies[:browser_time_zone])
+    ActiveSupport::TimeZone.all.find { |zone| zone.tzinfo == browser_tz } || Time.zone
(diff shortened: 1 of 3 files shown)


app/components/local_time_component.rb
```diff
@@ -0,0 +1,34 @@
...
+    formatted_date + " at " + time_of_day + " " + time_zone
+  end
+
+  private
+
+  def before_render
+    browser_tz = ActiveSupport::TimeZone.find_tzinfo(cookies[:browser_time_zone])
+    ActiveSupport::TimeZone.all.find { |zone| zone.tzinfo == browser_tz } || Time.zone
+  rescue TZInfo::UnknownTimezone, TZInfo::InvalidTimezoneIdentifier
+    @time_zone = Time.zone
```
FireLemons, line 32: we need this from the client. Can you add this as a parameter for the component?

FireLemons:
Can you add this to a random view in your local app without committing and take a screenshot?

FireLemons (changes requested):

app/components/local_time_component.rb
SajjadAhmad14: Hi @FireLemons, this is the local time zone of the user taken from browser. The arguments required for the component and mentioned in the issue are the time in unix timestamp format, and the format. So do I need to pass the time zone as the third argument?

SajjadAhmad14:
> Can you add this to a random view in your local app without committing and take a screenshot?
time

  <%= render (LocalTimeComponent.new(format: 12, unix_timestamp: 1694349775))%>
I rendered this component in users/edit_profile.rb

app/components/local_time_component.rb
FireLemons: Yes, it's from the client

FireLemons:
Oh my bad. I didn't know `Time.zone` was from the client. :1st_place_medal:

PR #5198, opened 2023-09-07
Add nix flake

jasonodoom:
This PR is not dependent on any particular issue.

### What changed, and why?

It seems there is a lot of "work" to get casa up and running.  Especially for new users. I decided to make a nix flake that will allow people to begin development quickly. You don't need to worry about installing or configuring rvm, yarn, postgresql, nodejs or even ruby. Just run this file and all the dependencies will be available in an environment separate from your host system. 

- Volunteer permissions: None
- Admin permissions: None

I ran this on macOS (Intel) and Linux

### Screenshots please :)
image


littleforest (approved):
I have never heard of this, but thanks for adding another dev setup option!

PR #5209, opened 2023-09-15
Issue 3028 convert csv to excel

### LINKED ISSUE #3028: Convert download case contact notes from CSV to Excel and turn on auto-wrap
 - volunteers?
 - supervisors?
 - admins?
 - all casa admins?

**Description**

Case supes asking, when download conacct notes to print,
wondering if “note fields” could wrap text automatically, don’t want to have to wrap ourselves

if its not a excel (and its probably not) we can convert it

_QA Login Details:_  

Login Emails: 

### Questions? Join Slack!

xeniabarreto:
Resolves [#3028](https://github.com/rubyforgood/casa/issues/3028)

### What changed, and why?

I implemented data extraction via Excel format with automatic line break correction in the "Case Contact Notes" column. 

As instructed by Shen Yang and Linda, I have also added a modal that allows users to choose the desired file format, either CSV or Excel. 

- Supervisor permissions:
- Admin permissions:

I provide below a video demonstration of the changes.

### Screenshots please :)

[Gravação de tela de 14-09-2023 23:01:41.webm](https://github.com/rubyforgood/casa/assets/88126195/10c6566f-e6ac-4c54-9db1-2e66d2c01463)

### Feelings gif (optional)
How to embed:
`alt text`


app/services/case_contacts_export_excel_service.rb
```diff
@@ -0,0 +1,87 @@
...
+  def string_width(string, row, font_size)
+    font_scale = font_size / row.worksheet.workbook.font_scale_divisor
+    (string.to_s.size + 3) * font_scale
+  end
+
+  def self.DATA_COLUMNS(case_contact = nil)
+    # NOTE: these header labels are for stakeholders and do not match the
+    # Rails DB names in all cases, e.g. added_to_system_at header is case_contact.created_at
+    {
+      internal_contact_number: case_contact&.id,
```
chahmedejaz, line 71: It's better to return an empty hash unless `casa_contact` present rather than using the safe navigation each time accessing this. i.e.
```ruby
def self.DATA_COLUMNS(case_contact = nil)
  return {} unless casa_contact

  {
    internal_contact_number: case_contact.id,
    .
    .
    case_contact_notes: case_contact.notes
  }
end 
```

app/services/case_contacts_export_excel_service.rb
```diff
@@ -0,0 +1,87 @@
...
+    end.max
+    row.height = (physical_lines * FONT_SIZE) + LINE_PADDING
+  end
+
+  def string_width(string, row, font_size)
+    font_scale = font_size / row.worksheet.workbook.font_scale_divisor
+    (string.to_s.size + 3) * font_scale
+  end
+
+  def self.DATA_COLUMNS(case_contact = nil)
```
chahmedejaz, line 67: I can see that this function is almost identical to the one we are using in `case_contacts_export_csv_service.rb` with format changes to `added_to_system_at` date:
https://github.com/rubyforgood/casa/blob/8f08a6ab596a120a686025e48d7ec56f49b2bb16/app/services/case_contacts_export_csv_service.rb#L23-L42

It's better to have consistency (_excel version of the hash looks better_) across both CSV and Excel reports and extract it out in a common helper module to be included in both services.

chahmedejaz (commented):
Good work overall :+1: 
**Edit:** There are some failing specs. Please review if they are related to your changes. Thanks.

app/services/case_contacts_export_excel_service.rb
xeniabarreto: I'm in the process of learning, I'm very grateful for your review and suggestion.

app/services/case_contacts_export_excel_service.rb
xeniabarreto: @chahmedejaz Thank you for your suggestion! 
I've implemented the changes as you recommended and pushed the modifications to the code.

app/services/case_contacts_export_excel_service.rb
xeniabarreto: @chahmedejaz I'd like to mention that I couldn't use your suggestion to return an empty hash if 'casa_contact' isn't present. The reason is that we're already using the same function with its keys for creating a modal where users select columns to filter before exporting. Thanks for your understanding :)

chahmedejaz:
Thank you so much for working on it, @xeniabarreto. Looks like we are almost there. Great work :+1:

.allow_skipping_tests
```diff
@@ -39,6 +39,7 @@ policies/note_policy.rb
 presenters/base_presenter.rb
 presenters/case_contact_presenter.rb
 services/case_contacts_export_csv_service.rb
+services/case_contacts_export_excel_service.rb
```
compwron, line 42: Hmm I am not a fan of skipping tests...

app/models/case_contact_report.rb
```diff
@@ -11,6 +11,11 @@ def to_csv
     CaseContactsExportCsvService.new(case_contacts_for_csv, columns).perform
   end
 
+  def to_excel
+    case_contacts_for_excel = @case_contacts
```
compwron, line 15: do we need this extra variable?

compwron (approved):
It looks good! Let's get the CI build working

app/services/case_contacts_export_data_columns.rb
schoork, line 1: I liked how you extracted the same code here so both CSV and Excel services could use it.

But what I don't like (and predates your code) is how this is used for two totally different things:

1. To get the list of columns for the view
2. To get a hash for each row of the CSV/Excel file

This is why all the safe operators (&) because in the view it's sending in nil for the case contact. I think a better pattern is to have a list of the columns and use them in the decorator.

app/views/reports/index.html.erb
```diff
@@ -221,3 +221,30 @@
...
+      <div class="modal-body">
+        <p>Select the export format:</p>
+        <button type="button" class="btn btn-primary" onclick="updateFormat('csv')" data-bs-dismiss="modal">CSV </button>
+        <button type="button" class="btn btn-primary" onclick="updateFormat('xlsx')" data-bs-dismiss="modal">Excel</button>
+      </div>
+    </div>
+  </div>
+</div>
+
+<script>
```
schoork, line 243: Not a fan of putting JS in a view. This should normally be separated into a JS file, but you can achieve the same result by moving the modal inside the form and letting the CSV and Excel buttons be your submit buttons. You just need to put  name and value tags on the buttons.

.allow_skipping_tests
schoork: I agree, but the methods are hard to test in isolation, and you can test them in the system spec.

app/services/case_contacts_export_excel_service.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+    end.to_stream.read
+  end
+
+  private
+
+  def configure_case_contact_notes_width(sheet)
+    case_contact_notes_index = filtered_columns.index(:case_contact_notes)
+
+    if case_contact_notes_index
```
schoork, line 45: This was a good catch from the previous code. It was throwing a nil --> Integer error when the notes weren't a selected column.

PR #5364, opened 2023-11-14
Case Contact Case ID Fix

schoork:
### What changed, and why?
Resolves a bug for multiple users logged in on the same computer where the autosave functionality of the case contact form could make it look like cases were checked when they really were not.
[Loom](https://www.loom.com/share/bb8c1b91851a420593f5133cbc87f68e?sid=ec011541-ea72-432d-bb60-894d06ccae37)

It will reduce the number of identical keys in the case contact autosave hash so autosave data will have a reduced chance of populating parts of a case contact form where the autosave data did not originate from. i.e. when another user logs in to use the case contact form


FireLemons:
:medal_sports:

PR #5367, opened 2023-11-15
Remove nix installation from general setup instructions

FireLemons:
### What changed, and why?
The nix setup instructions were under the general setup instructions section which confused new contributors into thinking nix was a required step in setting up the project.

 - The nix setup instructions have been moved to its own platform installation README
 - Also moved all of the WSL setup instructions into the platform specific install file.


jasonodoom:
sorry!

PR #5368, opened 2023-11-15
Fix browser warnings for volunteer pages

FireLemons:
### What changed, and why?
The browser gives warnings that a lot of our labels' for attributes don't match anything which is wrong.
Screenshot from 2023-11-15 06-29-59

Gave the learning hours title field the `required` attribute because it's required on the backend.

Just manually


FireLemons:
Well now some of the warning are different but they're not red at least
image

PR #5382, opened 2023-11-18
Refactor identical groups of form inputs to partials

FireLemons:
### What changed, and why?
Moved part of some forms where duration was specified by hours and minutes into a reusable partial.

CI, Manually tested removal of `.inline-fields` styling to make sure it did nothing. Checked mobile views too

### Screenshots please :)
Some of the forms are slightly different

Before Case Contact Form: 
The Hour(s) and Minute(s) text on the form is just a `<strong>`
image

After: 
The Hour(s) and Minute(s) text on the form are now labels coming before the inputs, matching the rest of the site. A label matching an input improves accessibility.
image

Before Learning Hours Form:
image

After: 
There is no improvement here. The difference is just a side effect of reusing the component.
image


schoork:
Really love this PR @FireLemons!

A suggestion. I like the use of the component, but I think a ViewComponent makes it even more explicit. It's a direction I think is worth heading. I put the initial changes in a branch (refactor_identical_groups_of_form_inputs_to_partials_with_components), but we'd want to write a few tests.

I like these because:
1) (Like your solution) you change the field in one place and it changes it everywhere
2) The partial is testable
3) You can use View Components' `render?` method to put an optional rendering on it (see the sidebar components for an example)
4) It's just so different from the normal "put all the fields in the same form partial" that it's very explicit to future programmers of the different pattern

PR #5384, opened 2023-11-18
Refactor identical groups of form inputs to partials with components

schoork:
For a suggestion to #5382


FireLemons:
Getting an instance of `ActionView::Helpers::FormBuilder` outside of an erb is the hardest thing ever. I had to ask the ai with some effort
image
and it comes up with this
```ruby
  let(:form_builder) {ActionView::Helpers::FormBuilder.new(:object, double('object'), ActionView::Base.new(ActionView::LookupContext.new('app/views'), {}, ActionController::Base.new), {})}
```

schoork:
Come to think of it, I'm not sure I've tested that way. Kudos!

PR #5415, opened 2023-12-02
fix: fixes the list of "no_recent_sign_in" on supervisor weekly_digest

### LINKED ISSUE #4672: Email bug: Inactive volunteers list is inaccurate - tracks logged in at but login lasts several days, so do most recent of case contact created OR logged in at
Volunteers that have been active are still showing up

LeGorge:
Resolves #4672 

### What changed, and why?
 - Changes the list of Volunteers the `weekly digest` supervisor email shows in the section `no_recent_sign_in`:
   - It was pointed out here https://github.com/rubyforgood/casa/issues/4672#issuecomment-1636244266 that the list should use the more recent of `last_sign_in_at` or last case_contact created. My understanding of it is that even if a volunteer has not logged in in the last 30 days, but he has a "recent" case_contact associated, the email section should not list that volunteer.
   - I used the case_contact's `occurred_at` field. Please verify and let me know that this is the field we want instead of using one of the other regular timestamps.
   - I changed the text in the email to let the supervisor know of that new behaviour.
   - I chose to maintain `User.no_recent_sign_in` as a scope even though this was the only place where it was used, so I just created a new call for supervisors to get that exact list. I feel like this situation should have its own name, but I'll leave that for you guys to figure out, in the meantime I used a big explanatory one just in case... let me know...

- No change

 - Added specs making sure all 3 scenarios are included and verified

### Screenshots please :)
 - does not apply

### DIFF
--- a/app/mailers/supervisor_mailer.rb
+++ b/app/mailers/supervisor_mailer.rb
@@ -10,6 +10,7 @@ class SupervisorMailer < UserMailer
     @supervisor = supervisor
     @casa_organization = supervisor.casa_org
     @inactive_messages = InactiveMessagesService.new(supervisor).inactive_messages
+    @inactive_volunteers = supervisor.inactive_volunteers
     if supervisor.receive_reimbursement_email
       mileage_report_attachment = MileageReport.new(@casa_organization.id).to_csv
       attachments["mileage-report-#{Time.current.strftime("%Y-%m-%d")}.csv"] = mileage_report_attachment
--- a/app/models/supervisor.rb
+++ b/app/models/supervisor.rb
@@ -37,6 +37,14 @@ class Supervisor < User
     ).where(invitation_accepted_at: nil).where.not(invitation_created_at: nil)
   end
 
+  def not_signed_in_nor_have_case_contacts_volunteers
+    recent_case_contact_volunteer_ids = volunteers.joins(:case_contacts).where(
+      case_contacts: {occurred_at: 30.days.ago..}
+    ).pluck(:id)
+
+    volunteers.no_recent_sign_in.where.not(id: recent_case_contact_volunteer_ids)
+  end
+
   def recently_unassigned_volunteers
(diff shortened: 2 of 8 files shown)


schoork:
Great work @LeGorge! I added a few tests, changed the wording a little, and made two small logical changes. Will highlight in the diff.

app/models/supervisor.rb
```diff
@@ -37,6 +37,14 @@ def pending_volunteers
     ).where(invitation_accepted_at: nil).where.not(invitation_created_at: nil)
   end
 
+  def inactive_volunteers
+    recent_case_contact_volunteer_ids = volunteers.joins(:case_contacts).where(
+      case_contacts: {created_at: 30.days.ago..}
```
schoork, line 42: Changed `occurred_at` to `updated_at`. Since these can be done in the past, a volunteer could create a contact 5 days in the past, and then the occurred_at might be misleading in the future.

app/models/user.rb
```diff
@@ -56,7 +56,7 @@ class User < ApplicationRecord
     where(casa_org_id: org.id)
   }
 
-  scope :no_recent_sign_in, -> { active.where("last_sign_in_at <= ?", 30.days.ago) }
+  scope :no_recent_sign_in, -> { active.where("last_sign_in_at <= ?", 30.days.ago).or(User.active.where(last_sign_in_at: nil)) }
```
schoork, line 59: Added the or clause for users who had never signed in. Found the use case when viewing in the mailer preview.

app/mailers/supervisor_mailer.rb
```diff
@@ -10,6 +10,7 @@ def weekly_digest(supervisor)
     @supervisor = supervisor
     @casa_organization = supervisor.casa_org
     @inactive_messages = InactiveMessagesService.new(supervisor).inactive_messages
+    @inactive_volunteers = supervisor.inactive_volunteers
```
schoork, line 13: Moved the logic from the view to the mailer.

schoork (approved):

PR #5453, opened 2023-12-20
Bootstrap V5 Badges class

fchagasjr:
### What changed, and why?
Updated "badge" class declaration on elements style to match Bootstrap V5.
Previous class implementation was not supported and it was not properly being shown on page rendering. 

### Screenshots please :)
Before:
Screen Shot 2023-12-19 at 9 11 00 PM

After:
Screen Shot 2023-12-19 at 9 07 48 PM


schoork:
Thank you @fchagasjr. Before I merge, have you used [View Components](https://viewcomponent.org/) before? This might be a good candidate for one.

fchagasjr:
@schoork I've never used, but I agree it would be a good candidate. Please, don't merge it yet, I'll do the changes first.

app/components/badge_component.rb
```diff
@@ -0,0 +1,18 @@
...
+  def initialize(text:, type:, rounded: false, margin: true)
+    @text = text
+    @type = type.to_sym
+    @rounded = rounded ? "rounded-pill" : nil
+    @margin = margin ? "my-1" : nil
+  end
+
+  def style
+    badge_style = ["bg-#{type}", rounded, margin]
+    badge_style.push("text-dark") if DARK_TEXT_TYPES.include?(type)
```
schoork, line 15: Beautiful!

app/components/badge_component.rb
```diff
@@ -0,0 +1,18 @@
+class BadgeComponent < ViewComponent::Base
+  DARK_TEXT_TYPES = [:warning, :light]
+
+  attr_reader :text, :type, :rounded, :margin
```
schoork, line 4: You don't technically need this line if you just use `@` notation in the style method.

app/helpers/sidebar_helper.rb
```diff
@@ -8,7 +8,7 @@ def cases_index_title
   def inbox_label
     unread_count = current_user.notifications.unread.count
     return "Inbox" if unread_count == 0
-    "Inbox <span class='badge badge-danger'>#{unread_count}</span>".html_safe
+    "Inbox <span class='badge bg-danger'>#{unread_count}</span>".html_safe
```
schoork, line 11: Not using here?

spec/components/badge_component_spec.rb
schoork, line 1: Tests! Yes!

app/helpers/sidebar_helper.rb
fchagasjr: I checked all the other instances on the browser to confirm they still kept the same design characteristics, but this function is not being called anywhere on the app to confirm this one.

app/components/badge_component.rb
fchagasjr: You're right. I guess only the `text` attribute needs to be accessible, the other ones can be private.

app/helpers/sidebar_helper.rb
schoork: Okay. I didn't notice this wasn't an html template either. All good.

schoork (approved):
Great work @fchagasjr. Really impressed how well you did with the View Components on your first shot!

fchagasjr:
@schoork thanks for the feedback! I'm glad to know that

PR #5504, opened 2024-02-01
Implements contact_topics on CasaOrg and CaseContact models (fixes #5500)

### LINKED ISSUE #5500: Create models and controllers needed for ContactTopic.
 - volunteers?
 - supervisors?
 - admins?
 - all casa admins?

## Description

This issue blocks #5388.

Issue: Create models and controllers needed for `ContactTopic`.

Behavior is still not fully figured out but the basic principles are such:
- Organizations create contact topics
  - These topics have a title and details (maybe other stuff in the future)
  - initially organizations will just have several default ones ability to modify them will be added later
- When creating a CaseContact the user should be presented with all of an organizations contact topics and answer some, none or all of them.
  - Those contact topics will then be associated with a case contact (contact topics can be selected but not answered)

## Acceptance Criteria

- There should be model tests added for the new model and whatever other models are afftects
- There should be controller / request tests for the forms controller

## Some Notes from Slack

The controller in question here is `controllers/case_contacts/form_controller.rb`. Depending on how you decide the database question (see below) you may or may not need to update it, but you will definitely need to update the `CaseContactParamters` which are listed in the controller. Here's a diagram of how the controller works.
image

potential idea: Maybe just start with the join table, since that's what this will need by the end. Something like this...
image

my current plan: 

The way I was thinking was each court report topic essentially has a title, details, and an answer. I was thinking

- each organization has many topics without answers
- When you start creating the case contact, you create copies of those cord topics and assign them to the case contact
- When you save the case contact they finally get created

contact topic ideas

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

elasticspoon:
This adds the model level behavior needed for contact topics. I thought this was a big enough PR to not want include request related stuff here.

### General Implementation

Topics are implements as a jsonb field on Organizations and CaseContacts in the form:
```
case_contact.contact_topics = 
[
    {"title" => "Main question", "details" =>["followup question", "followup question 2"], "active" => "true or false" },
    {...}, 
    ...
]
```

The allowed keys are set as `ContactTopicsValidator::PERMITTED_ATTRIBUTES`.

I created a Validator class for validations and ContactTopics for the main behavior itself. These could be combined but I was unsure if the behavior might grow so I separated them.

I was considering bringing in a gem to validate the json: https://github.com/voxpupuli/json-schema. I am not sure if that is too much for this case. Currently validations check the names of keys only.

#### Defaults

Defaults for topics are set during Organization creation, this is done in the controller. 

The values for defaults are set in `data/default_contact_topics.yml`. I don't know if this is the best place for this data or if it should by a .yml file.

### Tests

I created model tests. I wanted to create them as shared examples but the behavior differences were too large between CaseContacts and Orgs. 

I can't get the docker tests to pass even on main locally.

### Why JSON?

I initially wanted to use a polymorphic relationship the belongs_to org or case_contact. However, issues arose during the wizard that creates case_contacts:

image

For the duration of the wizard the CaseContact is built but it is not saved to database till the end. That means that if I create CaseContactTopics I don't have a way to persist them, since CaseContact does not have an id to reference. 

Thus, I decided that I had to store the values of topics on the object directly, same as the `notes` field.

Resolves #5500 partially

### DIFF
--- a/app/models/casa_org.rb
+++ b/app/models/casa_org.rb
@@ -11,6 +11,7 @@ class CasaOrg < ApplicationRecord
   validates :name, presence: true, uniqueness: true
   validates_with CasaOrgValidator
   validate :validate_twilio_credentials, if: -> { twilio_enabled || twilio_account_sid.present? || twilio_api_key_sid.present? || twilio_api_key_secret.present? }, on: :update
+  validates_with ContactTopicsValidator
 
   has_many :users, dependent: :destroy
   has_many :casa_cases, dependent: :destroy
@@ -85,11 +86,17 @@ class CasaOrg < ApplicationRecord
     self.slug = name.parameterize
   end
 
-  def generate_contact_types_and_hearing_types
+  def generate_defaults
     ActiveRecord::Base.transaction do
       ContactTypeGroup.generate_for_org!(self)
       HearingType.generate_for_org!(self)
+      ContactTopics.generate_for_org!(self)
     end
+    # FIXME: is this needed? do we care about the case where the defaults
+    # are invalid?
+  rescue ActiveRecord::RecordInvalid => e
+    errors.add(:base,
+      "Failed to generate default contact type groups, hearing types, or contact topics: #{e.message}")
   end
 
   def contact_types_by_group
@@ -149,6 +156,7 @@ end
 #  id                          :bigint           not null, primary key
 #  additional_expenses_enabled :boolean          default(FALSE)
 #  address                     :string
+#  contact_topics              :jsonb
 #  display_name                :string
 #  footer_links                :string           default([]), is an Array
 #  learning_topic_active       :boolean          default(FALSE)
--- a/app/validators/contact_topics_validator.rb
+++ b/app/validators/contact_topics_validator.rb
@@ -1,0 +1,16 @@
+class ContactTopicsValidator < ActiveModel::Validator
+  PERMITTED_ATTRIBUTES = %w[title details active].freeze
+
+  def validate(source)
+    contact_topics = Array.wrap(source.contact_topics)
+
+    contact_topics.each do |topic|
+      if !topic.is_a?(Hash)
+        source.errors.add(:contact_topics, "must be a hash or array of hashes")
+      elsif (topic.keys - PERMITTED_ATTRIBUTES).any?
+        source.errors.add(:contact_topics,
+          "expected keys: #{PERMITTED_ATTRIBUTES.join(", ")} got keys: #{topic.keys}")
+      end
+    end
+  end
+end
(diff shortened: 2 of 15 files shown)


data/default_contact_topics.yml
elasticspoon, line 1: Not totally sure about the best place or format to hold defaults. A YAML file seemed reasonable cause it is quite a bit of info.

spec/models/casa_org_spec.rb
elasticspoon, line 1: I realize it look like there are a lot of changes to this file. I simply fixed the nesting of a describe block
prev it was like 
```
describe 'default stuff'
  desc 'milage'
  end
end
```

pretty sure that was meant to be
```
desc 'defaults' do end
desc 'milage' do end
```

app/models/casa_org.rb
```diff
@@ -85,11 +86,17 @@ def set_slug
...
-  def generate_contact_types_and_hearing_types
+  def generate_defaults
     ActiveRecord::Base.transaction do
       ContactTypeGroup.generate_for_org!(self)
       HearingType.generate_for_org!(self)
+      ContactTopics.generate_for_org!(self)
     end
+    # FIXME: is this needed? do we care about the case where the defaults
+    # are invalid?
+  rescue ActiveRecord::RecordInvalid => e
```
elasticspoon, line 97: I'm wasn't too sure how to address the situation where the defaults are invalid? Should we pretend it can't happen?

data/default_contact_topics.yml
elasticspoon, line 1: I was not sure about the best place for defaults. Its quite a lot of text so I thought an YAML file was reasonable. Not sure about the best place for it.

spec/models/casa_org_spec.rb
```diff
@@ -128,52 +181,56 @@
       end
     end
 
-    describe "mileage rate for a given date" do
-      let(:casa_org) { build(:casa_org) }
+    it "generates default contact topics" do
+      expect(org.contact_topics).to eq(contact_topics)
+    end
+  end
```
elasticspoon, line 187: This looks like a lot of changes but I simply moved an `end` to fix nesting. Prev it went like 
```
desc 'defaults' do
   desc 'mileage' do
   end
end
```
now its
```
desc 'defaults' do end
desc 'mileage' do end
```

app/validators/contact_topics_validator.rb
elasticspoon, line 1: I am not sure this is the best approach to validations. I was considering using a gem to validate json schema but that might be too heavy handed for this small use case. https://github.com/voxpupuli/json-schema

elasticspoon:
It would also be cool if contact topics could be delegate its methods somehow, because in its current state controllers and views have to use it as `contact_topic["some_key"]` vs contact_topic.some_key. 

I could not figure out an easy way to implement this.

elasticspoon:
Rethinking my approach to this after going to office hours. I think a non-json approach should be possible if I write the needed case contacts to memory

PR #5517, opened 2024-02-09
Add Contact Topics (#5388)

### LINKED ISSUE #5388: Add "4. Court report topics (optional)" card/section to case contact form
 - volunteers
 - supervisors
 - admins

## Description
This is part of The Case Contact Form Improvement Epic. If you have any questions, please reach out to me, @bcastillo32, and/or @mallorypricedesign. The Case Contact form is one of the most important and heavily used features and your contribution will have a huge impact on CASA volunteers! :) 

## Reference Materials

[Figma File/Prototype](https://www.figma.com/file/QHOE6wHBESpsbxQmlnAZ2L/Case-Contact-Reporting-Process?type=design&node-id=196-19512&mode=design&t=5c48MqogwZ4AZctd-0)

**Steps to view a Figma prototype:**

1. Access the provided link in your web browser.
2. Press the play button located at the top right of the browser window.
3. (Optional) Create a free account and log in. This grants access to view comments left on the file.
4. (Optional) If the prototype extends beyond your screen width, click the "options" dropdown at the top right of the screen and select "fit width."
5. Explore the prototype by clicking different areas to discover interactive elements.

**Styling Chart**

image

## Accessing the Case Contact Form
- Login as a Volunteer
- Click on a case
- Click "+ New Case Contact" button

## Acceptance Criteria

1. The card header should read "4. Court report topics (optional)" - see reference materials above
2. Card components should match that of the figma file and styling chart - see reference materials
3. Card text should match figma file and use correct styling - see reference materials above

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

We highly recommend that you join us in slack https://rubyforgood.herokuapp.com/ #casa channel to ask questions quickly. And [discord](https://discord.gg/qJcw2RZH8Q) for office hours (currently Tuesday 5-7pm Pacific), stakeholder news, and upcoming new issues.

### LINKED ISSUE #5500: Create models and controllers needed for ContactTopic.

elasticspoon:
## Purpose

Fixes #5388

## Models

Adds 2 Models, modifies 2 existing models:
image

### Why?

I tried an approach where every topic would be a single record, however, that leads to a lot of duplicated information in the database. Saving a details with every record would result in a lot of extra data. 

Technically, we could store details and question as single objects and have each ContactTopic reference then, however, that result in issues with cross org changes to those records. Given in production there are 10 Orgs and 44k Cases that optimization did not seem worth it.

### Testing

- casa_org: checks that correct defaults are generated
- case_contact: checks the creation method works
- contact_topic: checks that correct defaults are used and invalid values do not pass

## Controllers

Added contact_topic controller. Call don't happen directly, happen as nested attributes from orgs
Added policy for changing contact topics: only same org admin can CRUD.
Generation of new topics happen after org creation in controller

### Tests

- contact_topics: test new, edit, create, update and soft delete
- form: tests update functions correctly during details, notes and final steps
- case_contacts: checks that new case_contacts copy org topics
- casa_org: check that topics can be edited, created

policy:
- only same org admins should be able to change topics

## Views

Admin Panel:
image

Delete Modal:
image

Form Details:
image

Form Notes:
image

### Tests

Systems tests focused on checking the javascript: making sure the dropdowns, read more/less work

## Components

The admin panel called for a kebab menu and a modal. I chose to make them generic and extract them to ViewComponents.

DropdownMenu Component: added a generic drop down menu component to wrap bootstrap behavior. It expects a body of items and have options to change how the button for the dropdown itself appears. The default if no icon is given and label is disabled is a kebab style icon. (I would have prefered to not add the svg but the icon pack did not include that icon)

Modal Component:
- group: combines a header, body and footer into a working modal. requires only the header or body + footer. needs a open button or link separately to open it.
- header: generic header with optional icon and text. if given content that will override
- body: takes text as string or array of strings and wraps in p tags. if given content that will override
- footer: has a close button, take content which it will render next to the close button
- open button/link: button/link to open the modal. requires the id of the modal to work. take optional icon. content overrides.


spec/models/casa_org_spec.rb
```diff
@@ -128,52 +134,61 @@
...
-          expect(casa_org.mileage_rate_for_given_date(Date.today)).to be_nil
-        end
+  describe "mileage rate for a given date" do
+    let(:casa_org) { build(:casa_org) }
+
+    describe "with a casa org with no rates" do
+      it "is nil" do
+        expect(casa_org.mileage_rate_for_given_date(Date.today)).to be_nil
       end
+    end
```
elasticspoon, line 154: I corrected the nesting here, which resulted in a massive amount of diff in this file. I promise nothing really changed.

spec/models/case_contact_topic_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        expect(created_topic.title).to eq(topics.first["title"])
+        expect(created_topic.details).to eq(topics.first["details"])
+      end
+
+      context "when defaults are invalid" do
+        it "fails if not all required attrs are present " do
+          fake_topics.first["title"] = nil
+          expect { CaseContactTopic.generate_for_org!(org) }.to raise_error(ActiveRecord::RecordInvalid)
+        end
+        it "creates if needed fields all persent" do
```
veezus, line 51: Typo: present

bcastillo32:
@elasticspoon do you need this approved and merged to continue with the original ticket? Let me know and I can help get that done.

app/models/case_contact_topic.rb
```diff
@@ -0,0 +1,42 @@
+class CaseContactTopic < ApplicationRecord
+  CASA_DEFAULT_COURT_TOPICS = Rails.root.join("data", "default_contact_topics.yml")
+  REQUIRED_ATTRS = %w[title details].freeze
+
+  belongs_to :source, polymorphic: true
```
schoork, line 5: I think you should split the topics owned by a CasaOrg and those owned by a CaseContact into separate tables and models.

Polymorphism is a cool, high-level design strategy, but it complicates the design anytime it is used. We have some strong developers who work on the project, but we also have some junior ones who will be effectively forced out of working in this area of the application. Imagine a partial that gets shared by both the Org ones and the CC ones. Then imagine this partial references some method from its source. Then imagine a dev puts the method on just one of the two source models. Then it breaks and they don't know why.

I just don't see the benefits of a polymorphic table here.

data/default_contact_topics.yml
schoork, line 1: This is a TON of data. So for each CC created from now on, we're going to store all 6 of these details sections in the database even if the volunteer doesn't answer the question? That's gonna expand the database fast.

db/migrate/20240128015203_create_case_contact_topics.rb
```diff
@@ -0,0 +1,14 @@
+class CreateCaseContactTopics < ActiveRecord::Migration[7.0]
+  def change
+    create_table :case_contact_topics do |t|
+      t.string :title
+      t.string :details
```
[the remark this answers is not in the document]
schoork: String is probably going to be too small. Text is going to be more future resilient. https://stackoverflow.com/questions/3354330/difference-between-string-and-text-in-rails

spec/factories/case_contact_topics.rb
```diff
@@ -0,0 +1,16 @@
+FactoryBot.define do
+  factory :case_contact_topic do
+    title { Faker::Lorem.sentence }
+    details { Faker::Lorem.paragraph }
```
schoork, line 4: You need to make sure you are generating more than 256 characters here.
https://faker.readthedocs.io/en/master/providers/faker.providers.lorem.html

db/migrate/20240128015203_create_case_contact_topics.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.string :answer
```
schoork, line 6: Also should be text.

schoork:
@elasticspoon I would highly suggest adding in the views for the Casa Org portion of this. That will give you more information about the UX, which should drive the backend as much as possible.

app/models/case_contact_topic.rb
elasticspoon: I already commented this in slack but I should probably put it here too so context is not lost. 

image

This is the other model structure that me and firelemons were discussing. Thoughts? I think the ids might be off but arrows show general belongs to ness. 

Amount of rows could also be further reduced with a triple join somewhere.

elasticspoon:
@schoork I have redone the models and such and added the org views. They are not complete yet but I think they give a general sense of the UX. I could not for the life of me figure out how to properly test policies. So those tests are currently incorrect. This PR has gotten extremely large, so I would love some more input before I move in any direction.

app/views/case_contacts/form/details.html.erb
```diff
@@ -79,6 +79,24 @@
...
+            <%= check_box_tag "case_contact[draft_case_ids][]",
+              0,
+              topic.selected,
+              id: "case_contact_draft_case_id_#{0}",
+              class: ["form-check-input", "casa-case-id"] %>
+            <label class="form-check-label" for="case_contact_draft_case_id_0">
+              <%= topic.contact_topic.question %>
+            </label>
+          </div>
+        <% end %>
```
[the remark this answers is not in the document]
schoork: Check out CaseContactParameters

app/views/case_contacts/form/details.html.erb
```diff
@@ -79,6 +79,24 @@
...
+      <h4 class="mb-20"><label for="case_contact_casa_case">1. Select the relevant <%= "case".pluralize(@casa_cases.count) %> for this contact <span class="red-letter">*</span></label></h4>
+      <div class="">
+        <% @case_contact.contact_topic_answers.each do |topic| %>
+          <div class="form-check checkbox-style mb-10 ml-5">
+            <%= check_box_tag "case_contact[draft_case_ids][]",
+              0,
+              topic.selected,
+              id: "case_contact_draft_case_id_#{0}",
+              class: ["form-check-input", "casa-case-id"] %>
+            <label class="form-check-label" for="case_contact_draft_case_id_0">
```
schoork, line 92: The ids in `id` and `for` should be tied to the topic's id. Currently, this happens.
https://www.loom.com/share/5cd50cc412d847e9ac8a8736c308fdbf?sid=388a37ce-b953-43d7-9491-ca8bc63e6a47

schoork: Line 90 is causing a serious bug.
https://www.loom.com/share/92d7526db848427c87e56998383a5347?sid=fab1458c-7648-446e-b7f0-7dbbe78d7202

spec/policies/contact_topic_policy_spec.rb
schoork, line 1: The main things to test here are:
1. Can a volunteer/supervisor create/edit these?
2. Can an admin from another org create/edit these for an org they don't belong to?

schoork (changes requested):
Overall it's looking great, and I can tell you've put a lot of time in this.

db/migrate/20240216013254_add_contact_topics.rb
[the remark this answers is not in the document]
elasticspoon: My thought process was if there are multiple casa orgs with the same default topics: say 5 questions and 5 details. We would only need to store 5 questions and 5 details records total that way instead of a new set for each org.

spec/policies/contact_topic_policy_spec.rb
elasticspoon: Is 2 not already tested?

```rb
 it "does not permit supervisor" do
      is_expected.to_not permit(supervisor, contact_topic)
    end

    it "does not permit volunteer" do
      is_expected.to_not permit(volunteer, contact_topic)
    end
```

app/controllers/contact_topics_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  # DELETE /contact_topics/1 or /contact_topics/1.json
+  def destroy
+    authorize @contact_topic
+
+    if @contact_topic.update(active: false)
+      redirect_to edit_casa_org_path(current_organization), notice: "Contact topic was successfully removed."
+    else
+      render :show, status: :unprocessable_entity
+    end
+  end
```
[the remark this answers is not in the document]
elasticspoon: So about this. Are we sure it makes sense for them to make a topic active again? 

The reason I did it this way is because topics can't be deleted otherwise court reports get messed up. Which means that is an admin creates 10 topics then decides they only want 4 they have to deal with 6 inactive topics forever. 

My logic was: wouldn't it be better UX for an admin to just "delete" and add topics instead of only being able to add.

app/views/casa_org/_contact_topics.html.erb
```diff
@@ -0,0 +1,61 @@
...
+          </tr>
+          </thead>
+          <tbody>
+          <% @contact_topics.each do |contact_topic| %>
+            <tr id="contact_topic-<%= contact_topic.id %>">
+              <td scope="row" class="min-width">
+                <%= contact_topic.question %>
+              </td>
+              <td scope="row" class="min-width">
+                <%= contact_topic.detail %>
```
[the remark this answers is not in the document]
elasticspoon: That's fair. The issue is I think an admin would want to see the full details on that screen without going to the edit page. 

I think the solution is probably having the topics be expandable cards like they will eventually be in the wizard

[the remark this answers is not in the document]
elasticspoon: what is we did it like this instead?
image
image

thats just a mockup. it still needs edit buttons, etc

db/migrate/20240216013254_add_contact_topics.rb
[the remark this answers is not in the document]
schoork: Here's a way to think about it. In Prod we have...

- 40,661 case contacts
- 10 Casa Orgs

If we were to duplicate 5 questions and details across case contacts (original idea for implementation) we would be storing over 200,000 details and questions. If we duplicate over Casa Orgs, we have 50.

We lose some functionality each time we simplify. In my mind it's worth it for not storing 200,000 of the same records, but 50 is totally fine for the benefit of simplicity.

spec/policies/contact_topic_policy_spec.rb
schoork: It looks like (1) is tested in that code snippet. Honestly, I didn't look at the tests. You mentioned you weren't sure how to test, so I just added what the tests should be. Maybe I misunderstood.

elasticspoon: Yea I wasn't sure but then me and shen went over it and figured it out. I forgot to update my comment so we all good in that regard

db/schema.rb
```diff
@@ -640,14 +664,14 @@
...
+  add_foreign_key "mileage_rates", "casa_orgs", validate: false
   add_foreign_key "mileage_rates", "users"
-  add_foreign_key "notes", "users", column: "creator_id"
+  add_foreign_key "notes", "users", column: "creator_id", validate: false
   add_foreign_key "other_duties", "users", column: "creator_id"
   add_foreign_key "patch_notes", "patch_note_groups"
   add_foreign_key "patch_notes", "patch_note_types"
   add_foreign_key "placement_types", "casa_orgs"
-  add_foreign_key "placements", "casa_cases"
+  add_foreign_key "placements", "casa_cases", validate: false
```
schoork, line 674: Let's manually take these out.

schoork: There's a ticket that I'll write up to address why this is happening.

elasticspoon: To clarify. If notes is going away, does that mean that it should be rolled in as a contact topic?

app/views/casa_org/_contact_topics.html.erb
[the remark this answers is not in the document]
elasticspoon: also 
image
it looks a lot more reasonable when the newlines aren't getting lost

db/schema.rb
schoork: Actually, run the commands to rebuild it and they will go away.

elasticspoon: it did. btw @schoork you have anything against me rebasing the branch and squashing some commits? Github is gonna lose of the changes since last review but I don't really like that I have merges between some of my commits.

schoork: No that's fine

app/controllers/contact_topics_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      render :edit, status: :unprocessable_entity
+    end
+  end
+
+  # DELETE /contact_topics/1 or /contact_topics/1.json
+  def destroy
+    authorize @contact_topic
+
+    if @contact_topic.update(active: false)
+      redirect_to edit_casa_org_path(current_organization), notice: "Contact topic was successfully removed."
```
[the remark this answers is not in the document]
elasticspoon: After talking to Malory and Brandon, I added a soft delete boolean and added being able to change active to the edit view. This controller action now soft deletes, I changed the action. But I have kept the language of the alert the same because I think its an implementation details that it soft deletes vs actually removing a record.

app/views/casa_org/_contact_topics.html.erb
```diff
@@ -0,0 +1,69 @@
...
+                  <%= link_to soft_delete_contact_topic_path(contact_topic), method: :delete,
+                    class: "btn-sm main-btn danger-btn btn-hover ms-auto" do %>
+                    <i class="lni lni-trash-can mr-10"></i>
+                    Delete Court Report Topic
+                  <% end %>
+                <% end %>
+                <% component.with_modal_content do %>
+                  <p>This topic and its related questions will be deleted and will no longer be presented while filling out case contacts.</p>
+                  <br>
+                  <p>This will not affect case contacts that have already been created.</p>
```
schoork, line 60: How about building this more into the ModalComponent itself. Create a dir for Modals:
```
# not including the html templates here, but they should all have one
app/component/modal/
  - modal_component.rb
  - content_component.rb
  - footer_component.rb
  - header_component.rb
```

app/views/case_contacts/form/_contact_topic_notes.html.erb
```diff
@@ -0,0 +1,49 @@
...
+           data-controller="icon-toggle"
+           data-icon-toggle-icons-value='["lni-chevron-up", "lni-chevron-down"]'>
+           <h4 class="details__topics-label">
+             <%= f.label :question, "#{f.index + 1}. #{topic.question}" %> <span class="content-1">(optional)</span>
+           </h4>
+           <button class="bg-transparent border-0"
+                   type="button"
+                   id="<%= answer_id %>_button"
+                   aria-expanded="false">
+             <i class="lni lni-chevron-up" data-icon-toggle-target="icon"></i>
```
schoork, line 20: The chevron is the same whether or not the card is open or closed. 
Screenshot 2024-03-02 at 8 07 35 AM

app/views/case_contacts/form/details.html.erb
schoork, line 1: If you're going to display all topics even if they are not selected on the first page, then if a user types in the text field, I think the autosave should also mark that topic as selected. Here's the current flow:

1. User doesn't select any. Clicks continue.
2. On the Notes page, opens a card and types. It gets saved.
3. User clicks back. The topic they entered the notes for is still displaying as unchecked.
4. When they click continue, the card is minimized again.

app/views/case_contacts/form/_contact_topic_notes.html.erb
```diff
@@ -0,0 +1,49 @@
...
+      </div>
+
+      <div class="topic-questions collapse <%= f.object.selected ? "show" : nil %>"
+           id="<%= answer_id %>">
+           <span class="content-2">Court report questions:
+             <span class="content-3 collapse show read-more <%= answer_id %>_details">
+               ... <a class="content-2"
+                      href=".<%= answer_id %>_details"
+                      data-bs-toggle="collapse"
+                      role="button">[read more]</a>
```
schoork, line 31: I like this functionality! Did you run it by Mallory or Brandon?

app/models/contact_topic_answer.rb
schoork, line 1: Should there be a validation for notes in the answer when it is selected?

app/models/contact_topic_answer.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  default_scope { joins(:contact_topic).order("contact_topics.question") }
+end
+
+# == Schema Information
+#
+# Table name: contact_topic_answers
+#
+#  id               :bigint           not null, primary key
+#  selected         :boolean          default(FALSE), not null
```
schoork, line 15: Is there functionality of `selected` beyond it's effect on the Notes page? Not saying it shouldn't be a database column, I'm just curious what you envision.

app/views/case_contacts/form/details.html.erb
schoork: You could add this as a callback on the ContactTopicAnswer model.

schoork (changes requested):

app/views/case_contacts/form/_contact_topic_notes.html.erb
elasticspoon: Yup it was in the original figma file I just missed it somehow

app/models/contact_topic_answer.rb
elasticspoon: Overall, I totally agree with your points. This is how the Casa folks want it. (That might potentially change when they actually get their hands on it)

Brandon says this how the selection should work:
- Selecting something on details opens it on notes
- What you do on notes page should not have any impact on select values on details page
- Regardless of what is selected all topics go into the final report

They (Brandon and Mallory) talked to the Casa folks that's the behavior they agreed on. The overall goal is just to not overwhelm volunteers with excessive text when they hit the notes page. (Source: https://rubyforgood.slack.com/archives/C0682C006L9/p1708372749226539 somewhere in that thread and we also talked over google meet)

app/views/casa_org/_contact_topics.html.erb
```diff
@@ -0,0 +1,69 @@
...
+          </div>
+        </div>
+      </div>
+      <div class="table-wrapper table-responsive">
+        <table class="table striped-table" id="contact-topics">
+          <thead>
+            <tr>
+              <th>Question</th>
+              <th>Details</th>
+              <th>Active?</th>
```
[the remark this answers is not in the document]
elasticspoon: Inactive: present on org page as Inactive and can be reactivated, not present on future topics until reactivated.
Soft Delete: removed from admin page and future topics, cannot be reverted. (cept by a dev)

Nothing can actually be deleted to preserve old topics.

app/views/casa_org/_contact_topics.html.erb
```diff
@@ -0,0 +1,69 @@
...
+                  <%= contact_topic.question %>
+                </td>
+                <td scope="row" class="min-width pre-line"><%= contact_topic.details %></td>
+                <td scope="row" class="min-width">
+                  <%= contact_topic.active ? "Yes" : "No" %>
+                </td>
+                <td>
+                  <%= render(KebabMenuComponent.new(menu_title: "Actions Menu")) do %>
+                    <li><%= link_to "Edit", edit_contact_topic_path(contact_topic), class: "dropdown-item" %></li>
+                    <li><%= render(ModalComponent.new( id: "contact_topic-#{contact_topic.id}", modal: false, button_value: "Delete")) %></li>
```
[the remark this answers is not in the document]
elasticspoon: >  I don't think a ModalComponent should handle a modal and a link.

Can you elaborate? In my mind a modal has two parts (also how Bootstrap has it and I'm copying their examples):
1. The content that covers screen and you have to interact with
2. The button that opens that content. 

Given that one has to link to the other with a common set of ids I thought it made sense to keep them together. 

Beyond that the modal (thing that pops up not specifically not button) has slots to fill in content.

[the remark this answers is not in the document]
schoork: Notice how in their documentation there is a button and a modal. You're ModalComponent is being used to create a button/link AND a modal itself. 
https://getbootstrap.com/docs/5.0/components/modal/#live-demo
Screenshot 2024-03-02 at 12 33 43 PM

I agree with generic, 100%. As much as possible make view components reusable. Check out how the sidebar uses the Group and Link components for an example.

app/views/casa_org/_contact_topics.html.erb
schoork: Sidebar is a smaller example for this.

app/models/contact_topic_answer.rb
schoork: Gotcha. Definitely go with what the customer wants and Brandon and Mallory say.

app/views/casa_org/_contact_topics.html.erb
elasticspoon: How about?
app/component/modal/
  - group_component.rb
  - content_component.rb
  - footer_component.rb
  - header_component.rb
  - button_component.rb
  - link_component.rb

I really want to include a way to generate a link/button associated with a modal. Cause that button/link need some bootstrap data attrs that I would like to abstract away. Also in some situations a <a> might make more sense vs a <button> so providing both would be nice. (This is actually one of those situation, the stuff in the dropdown renders different cause 1 is a button and the other is a link)

schoork: Sounds good

app/models/case_contact.rb
```diff
@@ -257,6 +259,14 @@ def volunteer
     end
   end
 
+  def create_contact_topic_answers!(casa_org)
+    casa_org.contact_topics.active.each do |topic|
+      contact_topic_answers << ContactTopicAnswer.new(contact_topic: topic)
+    end
+
+    save!
```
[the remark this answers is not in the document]
elasticspoon: made a new method that combines creation of case contact and answers:
```rb
  def self.create_with_answers(casa_org, **kwargs)
    create(kwargs).tap do |case_contact|
      casa_org.contact_topics.active.each do |topic|
        unless case_contact.contact_topic_answers << ContactTopicAnswer.new(contact_topic: topic)
          case_contact.errors.add(:contact_topic_answers, "could not create topic #{topic&.question.inspect}")
        end
      end
    end
  end
```
and added a check for errors in the controller:
```rb
    @case_contact = CaseContact.create_with_answers(current_organization,
      creator: current_user, draft_case_ids: draft_case_ids)
    if @case_contact.errors.any?
      flash[:alert] = @case_contact.errors.full_messages.join("\n")
      redirect_to request.referer
    else
      redirect_to case_contact_form_path(CaseContact::FORM_STEPS.first, case_contact_id: @case_contact.id)
    end

```

app/controllers/casa_cases_controller.rb
```diff
@@ -17,6 +17,7 @@ def show
 
     respond_to do |format|
       format.html {}
+      # TODO: add contact topic for generation
```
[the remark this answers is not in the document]
elasticspoon: i want to leave this so I remember where report creation comes from. If I don't have enough info to finish the report creation code by the time the rest of the PR is ready then ill remove it.

elasticspoon:
After some clarification on how topics are represented in reports I went back on the idea of sorting topics in alphabetical order. Casa wants them in a particular order for the reports which is not alphabetical. 

Depending on how it goes there might be need to implement reordering of topics. That said, apparently topics rarely change so that should not be much of an issue.

app/views/case_contacts/form/details.html.erb
```diff
@@ -79,6 +79,17 @@
       </div>
     </div>
 
+    <div class="card-style-1 pl-25 mb-10">
+      <h4 class="mb-20 details__topics-label">Court report <%= "topic".pluralize(@case_contact.contact_topic_answers.count) %> <span class="content-1">(optional)</span></label></h4>
+      <div class="">
+        <%= form.fields_for(:contact_topic_answers) do |f| %>
```
schoork, line 85: These change order when you move through and back in the form. A little confusing. Can we do alphabetical?

app/components/modal/body_component.rb
```diff
@@ -0,0 +1,21 @@
...
+    @class = klass
+  end
+
+  def body_content
+    return content if content.present?
+
+    Array.wrap(@text).map do |text|
+      content_tag :p, text
+    end.join.html_safe
+  end
```
schoork, line 16: This seems pretty restrictive to what can go in to the body. Why not just insert what the content is and do something like this when rendering...?
```
<% component.with_body do %>
  <p>
    This topic and its related questions...
  </p>
  <p>
    This will not affect case contacts that...
  </p>
<% end %>
```

app/components/modal/open_button_component.rb
schoork, line 1: I don't think you're using this anymore. Is there a reason to keep this? 
Screenshot 2024-03-09 at 11 18 51 AM

schoork:
This is a really flaky test. I think I wrote it. And I'm fine with you removing it. Maybe change it to an `xit` so we know to re-write it in the future.
Screenshot 2024-03-09 at 11 24 36 AM

schoork (approved):

schoork:
@elasticspoon This looks really good! A few extra notes. Nothing blocking that I can see. Ready when you are to get this across the line.

app/components/modal/body_component.rb
elasticspoon: You can do that, currently if you provide a block it overrides and text is optional. 

It's just if you provide text as a string or array of strings. It gets wrapped in ptags. 

https://github.com/rubyforgood/casa/blob/897d5f79737f894c32dc0e28a0570daf8f833f6d/spec/components/modal/body_component_spec.rb#L20

app/components/modal/open_button_component.rb
elasticspoon: The idea was to provide two different options to open the model. As a button tag and a link tag, they format slightly differently. 

In most situations you would want to use a button because it's semantically correct. But in some situations you're forced to use a link like in my case. So I wanted to have both.

elasticspoon:
@schoork Cleaned up the remaining comments. I think we are good to go with this part. I will work on the report generation in a new PR.

PR #5518, opened 2024-02-09
Attempt to fix failing tests in new_spec.rb under learning_hours

pshong79:
N/A

### What changed, and why?
Tests were failing locally in the `new_spec.rb` under `app/system/learning_hours`. Made the change from `Time.zone.now` to `Time.now`, which seems to have fixed the issue locally; however, pushing to check to see if CI also passes with this change.

N/A

N/A

### Screenshots please :)
N/A

### Feelings gif (optional)
How to embed:
`alt text`

### DIFF
--- a/app/views/learning_hours/_form.html.erb
+++ b/app/views/learning_hours/_form.html.erb
@@ -49,13 +49,17 @@
   </div>
   <div class="field-card mt-4">
     <h5 class="mb-3"><%= form.label :occurred_at, "Occurred On" %>:</h5>
-    <% occurred_at = learning_hour.occurred_at || Time.zone.now %>
+    <div><%= "learning_hour.occurred_at = #{learning_hour.occurred_at}" %></div>
+    <div><%= "Time.now = #{Time.zone.now}" %></div>
+    <% occurred_at = learning_hour.occurred_at || Time.now %>
     <div class="input-style-1">
       <%= form.text_field :occurred_at,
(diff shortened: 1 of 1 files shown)


app/views/learning_hours/_form.html.erb
```diff
@@ -49,13 +49,17 @@
   </div>
   <div class="field-card mt-4">
     <h5 class="mb-3"><%= form.label :occurred_at, "Occurred On" %>:</h5>
-    <% occurred_at = learning_hour.occurred_at || Time.zone.now %>
+    <div><%= "learning_hour.occurred_at = #{learning_hour.occurred_at}" %></div>
+    <div><%= "Time.now = #{Time.zone.now}" %></div>
+    <% occurred_at = learning_hour.occurred_at || Time.now %>
```
littleforest, line 54: We should never be using `Time.now` as that refers to system time and not the application time. [Ref](https://thoughtbot.com/blog/its-about-time-zones#three-time-zones). Can you revert these changes and instead work on [freezing the time](https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html) in the spec instead? Which spec is flaky?

pshong79: @littleforest Sorry! I completely forgot about this and I can't reproduce this test failure anymore... I'm going to close this PR. 

But thanks for those `Time` examples. They are really helpful!

PR #5519, opened 2024-02-09
Fixes bugsnag console message

elasticspoon:
Without an API key bugsnag displays the following warning: 
```sh
Not delivering sessions due to an invalid api_key
```

It is something that new people often get worried about and is very annoying when using irb for debugging. This PR sends that warning to `dev/null` if an API key is not provided.

Solution take from [here](https://github.com/chaskiq/chaskiq/issues/251#issuecomment-1813804184)


compwron (approved):
seems reasonable :)

PR #5478, opened 2024-01-08
Rails upgrade 7.0.8 to 7.1.0

sarvaiyanidhi:
### What github issue is this PR for, if any?
Resolves #5462 

### What changed, and why?
Upgrade Rails to newer version

### How will this affect user permissions?
- Volunteer permissions: Not Effected
- Supervisor permissions: Not Effected
- Admin permissions: Not Effected

### How is this tested? (please write tests!) 💖💪
Have made changes in existing test to pass as per new configuration

### Screenshots please :)
NA

PR #5535, opened 2024-02-17
Remove unused actions in controller filters

### LINKED ISSUE #5502: By default this was set to true but I have changed it to false as it was failing specs for verify_authorized action
By default this was set to true but I have changed it to false as it was failing specs for verify_authorized action

_Originally posted by @sarvaiyanidhi in https://github.com/rubyforgood/casa/pull/5478#discussion_r1447657413_

JoshDevHub:
Resolves #5502 

### What changed, and why?
- In `CaseContactsController`, removed the `update` action (which isn't defined) from the `only` list for one of the callbacks
- In `CourtDatesController`, removed the `generate` action (which isn't defined) from the `only` list for one of the callbacks

In Rails 7.1, they've added a new configuration to catch typo'd and undefined controller actions listed in the `only` or `except` lists to a controller filter. [Relevant PR](https://github.com/rails/rails/pull/43487).

This configuration option is written by default in `config/environments/test.rb` and `config/environments/development.rb`. In the upgrade to Rails 7.1 in #5478, this configuration was disabled in order to pass the tests. By re-enabling it, we were able to catch a couple of places where unused actions were listed for a controller filter.

We are unable to set that configuration `true` because of two lingering issues:
1. In `ApplicationController`, the `#verify_authorized` callback fires except for the `:index` action. This is an issue because `ApplicationController` is an abstract controller -- it offers no actions of its own. One way to fix this would be to fully apply `#verify_authorized` in the `ApplicationController`, and then, in places where this should be skipped, use `skip_before_action :verify_authorized` to do so.
2. In `Api::V1::BaseController`, the `#authorize_user!` callback fires except for `:create`. This is the same issue as above, as this is an abstract controller that doesn't have its own actions. So it could be fixed in the same way.

Glad to pursue fully fixing those two remaining issues and enabling the `raise_on_missing_callback_actions` if that's desired. Or it could just be left disabled.

Edit: another option would be to use option 2 specified in the following issue: https://github.com/ackama/rails-template/issues/507

- Volunteer permissions: N/A
- Supervisor permissions: N/A
- Admin permissions: N/A

N/A

### Screenshots please :)
N/A

### Feelings gif (optional)
train on rails

### DIFF
--- a/app/controllers/case_contacts_controller.rb
+++ b/app/controllers/case_contacts_controller.rb
@@ -2,7 +2,7 @@
 
 class CaseContactsController < ApplicationController
   before_action :set_case_contact, only: %i[edit destroy]
-  before_action :set_contact_types, only: %i[new edit update create]
+  before_action :set_contact_types, only: %i[new edit create]
   before_action :require_organization!
   after_action :verify_authorized, except: %i[leave]
 
--- a/app/controllers/court_dates_controller.rb
+++ b/app/controllers/court_dates_controller.rb
@@ -2,7 +2,7 @@ class CourtDatesController < ApplicationController
   include CourtDateParams
 
   before_action :set_casa_case
-  before_action :set_court_date, only: %i[edit show generate update destroy]
+  before_action :set_court_date, only: %i[edit show update destroy]
   before_action :require_organization!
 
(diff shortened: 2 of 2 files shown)


compwron (approved):
seems good, tests pass

PR #5570, opened 2024-03-12
improve invalid factorybot linter, some specs

compwron:
factorybot linting

### DIFF
--- a/spec/factories/case_contacts.rb
+++ b/spec/factories/case_contacts.rb
@@ -58,7 +58,6 @@ FactoryBot.define do
     end
 
     trait :expenses_status do
-      casa_case { nil }
       draft_case_ids { [1] }
     end
 
--- a/spec/factories/supervisors.rb
+++ b/spec/factories/supervisors.rb
@@ -5,7 +5,8 @@ FactoryBot.define do
 
     trait :with_casa_cases do
       after(:create) do |user, _|
-        create_list(:case_assignment, 2, volunteer: user)
+        volunteer = create(:volunteer)
+        create_list(:case_assignment, 2, volunteer: volunteer)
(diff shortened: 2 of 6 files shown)


spec/factories/case_contacts.rb
```diff
@@ -58,7 +58,6 @@
     end
 
     trait :expenses_status do
-      casa_case { nil }
```
compwron, line 61: this was invalid and seemed bad so I removed it

spec/factories/supervisors.rb
```diff
@@ -5,7 +5,8 @@
 
     trait :with_casa_cases do
       after(:create) do |user, _|
-        create_list(:case_assignment, 2, volunteer: user)
+        volunteer = create(:volunteer)
```
compwron, line 8: the user in this case is I believe the supervisor. The cases should not be assigned to the supervisor.

spec/factories/users.rb
```diff
@@ -12,7 +12,7 @@
     token { "verysecuretoken" }
 
     trait :inactive do
-      volunteer
+      type { "Volunteer" }
```
compwron, line 15: this previous way was simply not working I think

spec/models/case_contact_spec.rb
```diff
@@ -109,17 +109,23 @@
     end
 
     it "requires duration minutes" do
-      expect(build_stubbed(:case_contact, :details_status, duration_minutes: nil)).not_to be_valid
+      obj = build_stubbed(:case_contact, :details_status, duration_minutes: nil)
```
compwron, line 112: better testing of errors - I'd like more good testing of errors

PR #5577, opened 2024-03-20
Jd refactor import case date script

jdsoteldo:
This refactor doesnt have any issues attached, I was just browsing the files for something to refactor and decided on cleaning up the script a little bit. 

### What changed, and why?
There were no logic or behavior changes, only some cleanup on the code. It was a refactor so no changes to any expectations.

it wont.

From my understanding this is a script that gets run manually with real prod data of which i have no access to. Since there were no changes to funcionalities I believe there is no need for new tests

Sure :D


compwron:
This is a cool refactor! If you fix lint, we can merge

compwron:
eh, let's merge and fix lint forward

PR #5590, opened 2024-03-27
refactor: expose court report methods

elasticspoon:
### Context
I will soon need to make a few changes to report generation and the current `CaseCourtReportContext` makes it very difficult to change behavior.

### The Problem
Currently `CaseCourtReportContext` has a single public method that outputs a very complex object:

```rb
  {
    created_date: I18n.l(Time.current.in_time_zone(@time_zone).to_date, format: :full, default: nil),
    casa_case: prepare_case_details,
    case_contacts: prepare_case_contacts,
    case_court_orders: prepare_case_orders,
    case_mandates: prepare_case_orders, # backwards compatible with old Montgomery template - keep this! TODO test full generation
    latest_hearing_date: latest_hearing_date.nil? ? "___<LATEST HEARING DATE>____" : I18n.l(latest_hearing_date.date, format: :full, default: nil),
    org_address: org_address(is_default_template),
    volunteer: volunteer_info,
    hearing_type_name: @court_date&.hearing_type&.name || "None"
  }
```

The hash itself gets built using smaller methods but is tested by creating the full hash then inspecting each key. Ex:
```rb
expect(court_report_context.context[:casa_case][:court_date]).to eq("March 1, 2021")
```

This makes it hard to extend the behavior of the class and write additional tests.

### What changed, and why?
This PR reworks some of the methods and tests for `CaseCourtReportContext`.

- Expose methods in court report context to make unit testing easier.
- Write additional unit tests.
- Remove old tests.
- Remove redundant tests, some behavior is already tested by `CaseContactsContactDates`.

None

Unit tests

### DIFF
--- a/spec/models/case_court_report_context_spec.rb
+++ b/spec/models/case_court_report_context_spec.rb
@@ -13,274 +13,152 @@ RSpec.describe CaseCourtReportContext, type: :model do
   end
 
   describe "#context" do
+    it "has the right shape" do
+      date = 1.day.ago
+      court_date = build(:court_date, :with_hearing_type, date: date)
+      context = create(:case_court_report_context, court_date: court_date)
+
+      allow(context).to receive(:case_details).and_return({})
+      allow(context).to receive(:case_contacts).and_return([])
+      allow(context).to receive(:case_orders).and_return([])
+      allow(context).to receive(:org_address).and_return(nil)
+      allow(context).to receive(:volunteer_info).and_return({})
+      allow(context).to receive(:latest_hearing_date).and_return("")
+
+      expected_shape = {
+        created_date: "January 1, 2021",
+        casa_case: {},
+        case_contacts: [],
+        case_court_orders: [],
+        case_mandates: [],
+        latest_hearing_date: "",
+        org_address: nil,
+        volunteer: {},
+        hearing_type_name: court_date.hearing_type.name
+      }
+
+      expect(context.context).to eq(expected_shape)
+    end
+  end
+
+  describe "case_orders" do
+    it "returns the correct shape" do
+      court_orders = [
+        build(:case_court_order, text: "Court order 1", implementation_status: :unimplemented),
+        build(:case_court_order, text: "Court order 2", implementation_status: :implemented)
+      ]
+      expected = [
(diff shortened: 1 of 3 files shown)


spec/models/case_court_report_context_spec.rb
```diff
@@ -13,274 +13,152 @@
...
-          end
-        end
-      end
+    # not sure how to test this without just restating the code
+    # context 'when there are interviewees' do
+    #   it 'it calls CaseContactsContactDates with filtered values' do
+    #     create_list(:case_contact_contact_type, 3, case_contact: create(:case_contact, casa_case: casa_case))
+    #     context.case_contacts
+    #   end
+    # end
```
elasticspoon, line 125: Not sure how best to test the case where there are interviewees, should I even test it?

```rb
  def case_contacts
    cccts = CaseContactContactType.includes(:case_contact, :contact_type).where("case_contacts.casa_case_id": @casa_case.id)
    interviewees = filter_out_old_case_contacts(cccts)
    return [] unless interviewees.size.positive?

    CaseContactsContactDates.new(interviewees).contact_dates_details
  end
```

All this could really test is that the `ActiveRecord` query is correct, and its not a very complex query. Is there value in it?

compwron: I am somewhat unclear on what an interviewee is
In general I am in favor of more testing rather than less tho

compwron (commented):
In general I am in favor of this but I am nervous about it possibly breaking court report generation somehow

compwron (approved):

elasticspoon:
> In general I am in favor of this but I am nervous about it possibly breaking court report generation somehow

It certainly could be just because the testing around report generation is pretty convoluted. That said these changes still pass all the tests directly related to report generation. 

In my other PR I test the report like this: https://github.com/rubyforgood/casa/blob/deea6bbc8c5eb841432a39efb81959c5e2364232/spec/models/case_court_report_spec.rb#L11

Which imo makes it much easier to be sure we get all the edge cases. Ideally I would like to refactor the whole report test to be of that form. But it's a pretty big shift so idk.

PR #5603, opened 2024-04-11
test ci stuff

elasticspoon:
Can we speed up rspec tests by running them in parallel?


compwron:
quite possibly!

elasticspoon:
It seems that human essentials does the CI parallelization with KnapsackPro, which seems to be free for open source projects as long as you add a badge. I'll ask how they got a license, maybe we could do the same in this project.

compwron:
> It seems that human essentials does the CI parallelization with KnapsackPro, which seems to be free for open source projects as long as you add a badge. I'll ask how they got a license, maybe we could do the same in this project.

sounds good!

PR #5623, opened 2024-04-15
fix: remove sleeps

elasticspoon:
Blocked by https://github.com/rubyforgood/casa/pull/5625

Some system specs were relying on sleep to wait for client side JS to finish.

This changes the behavior to be more reliable by using built in Capybara methods.

Also removes animations in systems tests.

### DIFF
--- a/spec/support/capybara.rb
+++ b/spec/support/capybara.rb
@@ -11,6 +11,9 @@ Capybara.register_driver :selenium_chrome_in_container do |app|
     capabilities: [:chrome]
 end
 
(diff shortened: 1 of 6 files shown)


spec/support/capybara.rb
```diff
@@ -11,6 +11,9 @@
     capabilities: [:chrome]
 end
 
+# disable CSS transitions and js animations
+Capybara.disable_animation = true
```
compwron, line 15: nice

compwron (approved):
YAYYYY!

PR #5625, opened 2024-04-15
fix: select correct report in controller

elasticspoon:
image

Failures on main for me. Not sure how CI is passing.  

Apparently in controllers relating to court reports, when were never scoping correctly. If there were multiple reports with the same case number it would show the incorrect report.

This is possible if two orgs have cases with the same number.


compwron (approved):
nice! Thanks :)

PR #5628, opened 2024-04-16
parallel_test gem experimentation

elasticspoon:
Uses `parallel_tests` gem to run tests in parallel in CI. Parallelizes the tests across 4 CPU cores on each Github Action and across 3 runners. Additional runners did not seem useful, the setup time was starting to be as long as the test time. 

As a result the RSpec run time was cut down from around 7:30 to 3:10.

### Changes in Test Output

Since the test output is now spread across multiple runners we want to have a way to combine those outputs back together into a single useful result. Long story short it will look something like this:

Success:
image

Failure:
image

### Docker

I was unable to do a similar thing for docker. This issue with docker is the setup for docker compose takes almost 2 mins, which means even with many runners the tests will be spending nearly as much time setting up as testing.

### Notes

Rspec is marked as a required check. This will need to be changed to Continuous Integration.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -333,6 +333,8 @@ GEM
       bigdecimal (>= 3.0)
     orm_adapter (0.5.0)
     parallel (1.24.0)
+    parallel_tests (4.7.1)
+      parallel
     paranoia (2.6.3)
       activerecord (>= 5.1, < 7.2)
     parser (3.3.1.0)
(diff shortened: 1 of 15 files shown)


Gemfile.lock
```diff
@@ -555,15 +559,10 @@ GEM
     zeitwerk (2.6.13)
 
 PLATFORMS
-  arm64-darwin-20
```
compwron, line 558: why remove these?

compwron (approved):
it its working, let's merge!

PR #5663, opened 2024-05-05
test: system test for new prod fix

elasticspoon:
Tests that a user can create a new case contact in the situation where an org has neither travel expense or driving reimbursement.


compwron (approved):
very nice :)

PR #5692, opened 2024-05-05
remove prod debugging bugsnag in form controller

compwron:


app/controllers/case_contacts/form_controller.rb
```diff
@@ -25,13 +25,7 @@ def update
...
     else
-      begin
-        params[:case_contact] ||= []
-        params[:case_contact][:status] = step.to_s # TODO: where is this used?? what is it for??
-      rescue => e
-        # TODO https://app.bugsnag.com/ruby-for-good/casa/errors/6637007c6857010008cfc9dd
-        Bugsnag.notify(e)
-      end
+      params[:case_contact][:status] = step.to_s
     end
```
elasticspoon, line 29: `params[:case_contact][:status] = step.to_s unless @case_contact.active?`

is what it was in the past unless you think this is more clear.

elasticspoon (approved):
Small nit. Up to you if you wanna fix.

app/controllers/case_contacts/form_controller.rb
compwron: I like it in the long form tbh

PR #5725, opened 2024-05-18
fix: restore old rspec in CI

elasticspoon:
RSpec parallel is still too flaky. Bring back old pipeline until the new one is ready.

This is one option. If there is not enough CI budget for all the runners I will open a PR to disable the parallel ones completely until they are more ready.


compwron (approved):
you rock!

PR #5749, opened 2024-05-26
feat: precompile assets in CI

elasticspoon:
Maybe this will reduce memory consumption? (Is memory consumption even the issue?)

`rails assets:precompile` does the equivalent of `yarn:build` + `yarn:build:css` + compiles other assets to serve statically


compwron (approved):
nice

PR #5814, opened 2024-06-04
Unveil codespaces

elasticspoon:
#5801 Added integration with Github Codespaces, it was a bit rushed to allow use at the RubyforGood 2024 event. 

This PR makes the adds mention to the Readme that the project supports codespaces.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -81,11 +81,12 @@ Rails.application.configure do
   # Uncomment if you wish to allow Action Cable access from any origin.
   # config.action_cable.disable_request_forgery_protection = true
 
-  # Uncomment if you wish to use codespaces
-  # config.action_controller.forgery_protection_origin_check = false
-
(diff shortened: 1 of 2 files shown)


config/environments/development.rb
```diff
@@ -81,12 +81,13 @@
...
-  # Uncomment if you wish to use codespaces
-  # config.action_controller.forgery_protection_origin_check = false
-
   config.hosts << ENV["DEV_HOSTS"]
   config.hosts << ".app.github.dev"
 
+  if ENV["CODESPACES"] === "true"
+    config.action_controller.forgery_protection_origin_check = false
+  end
+
```
elasticspoon, line 90: I don't love this but I could not find a better solution.

compwron: I like the env var protection

compwron (approved):
nice

PR #5816, opened 2024-06-06
feat: add contact notes to case contact card

### LINKED ISSUE #5784: Add Court Report Topics Answers to Case Contact Cards
On the `/casa_case/:case_id` page we have cards for each case contact associated with a case.

Image

As you see there are notes for that that case contact, however, there should be more of them. 

Image

Each Case Contact can have additional topics set by the organization that the volunteers fill in. These are not currently being displayed on the card. Add them. 

They should be shown in a format like 
```
Contact Topic 1: Notes on topic
Contact Topic 2: Notes on topic
Additional Notes: even more notes....
```

I believe there should be some functionality already in the application to truncate the notes and then show/ hide the full note with javascript.

elasticspoon:
Resolves #5784

### What changed, and _why_?
Created a component to enable text truncation behavior as well as showing the full text. 

Shows notes and answers to contact topics on a per case contact basis.

component and view specs

### Screenshots please :)
image

### Notes

Truncation is done with CSS rather than server side. I think this is a better approach because it makes the truncation dynamic based on screen size. In this particular use case, I specifically wanted the component to be at most one line in size. 

Given that, maybe it could be named a bit better? I am open to suggestions.


app/views/case_contacts/_case_contact.html.erb
```diff
@@ -21,19 +21,25 @@
...
-              <% else %>
-                <%= simple_format(contact.decorate.full_notes) %>
-              <% end %>
+            <div class="card-text mb-2">
+              <ul>
+                <% contact.contact_topic_answers.reject { _1.value.blank? }.each do |answer| %>
+                  <li>
+                    <%= render(TruncatedTextComponent.new) do %>
+                      <span class="text-bold d-inline"><%= answer.contact_topic.question %>:</span>
+                      <%= answer.value %>
```
schoork, line 30: Why not move these into the component's template?

app/components/truncated_text_component.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def text_content
+    return content if content.present?
+
+    simple_format(@text)
+  end
+
+  def render?
+    @render_check && (@text.present? || content.present?)
+  end
```
schoork, line 17: (Lines 9-17) Unless you're seeing a future use of these, I'd consider removing them, since you're not using them now.

schoork (approved):
A few comments, but overall looks great!

PR #5828, opened 2024-06-10
Improve banner expiration validations and form

### LINKED ISSUE #5815: Improve banner expiration validations and form
 - admins

## Description

#5800 added expiration times to banners (`/banners`)

After testing the feature we want to make some improvements to it.

- [ ] add server side (and client if possible with basic html) to validate that expiry cannot be in the past. 
- [ ] validate the banner actually has content, otherwise a banner will be active but nothing will be shown
- [ ] when a banner expires set it to inactive (this might be complicated with the added validations but explore it)
- [ ] Rename values as shown below

From:
Image

To:
Image

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

jp524:
Resolves #5815.

### What changed, and _why_?
Modified banners per issue description:
- [x] Rename values in 'Expires At' column
- [x] Add server-side and client-side validation to validate that expiry cannot be in the past
- [x] Validate the banner actually has content (done server-side only. No built in client-side validation exists for `<trix>` elements)
- [x] When a banner expires set it to inactive (I implemented this by modifying the existing `Banner#expired?` method. I'm open to feedback if there are better ways to implement this)

Added some new model and system tests.

### Screenshots please :)
**New banner table**
New banner table

**Expires at validation**
Expires at validation

**Content validation**
Content validation


spec/system/banners/new_spec.rb
```diff
@@ -49,14 +49,29 @@
...
+    visit banners_path
+    click_on "New Banner"
+    fill_in "Name", with: "Annoucement"
+    fill_in "banner_expires_at", with: 1.hour.ago.strftime("%m%d%Y\t%I%M%P")
+    fill_in_rich_text_area "banner_content", with: "Please fill out this survey."
+    click_on "Submit"
+
+    message = page.find("#banner_expires_at").native.attribute("validationMessage")
+    expect(message).to start_with("Value must be")
+    # Can't get an exact match for the date/time since cookie "browser_time_zone" isn't set in the test Chrome session
```
elasticspoon, line 74: This test fails for me locally. In my case the browser is providing a cookie with the EST timezone but the test and server are assuming UTC. So `1.hour.ago` gets converted from EST to UTC but upon input it gets treated as it if was EST since that is what the cookie says to it end up not actually being an hour in the past.

Basically if its 1AM now => 1 hour ago = 12AM => inputted in UTC = 5AM (that is what goes into the field)
But the browser has a cookie set for EST so it converts that to UTC = 10AM UTC which is after 6AM (1 AM EST / 6AM UTC) so there is no error.

Long story short: `page.driver.browser.manage.add_cookie(name: "browser_time_zone", value: "Something")` to set a cookie.

jp524: Thank you! I changed the test to specify a timezone and it's working for me locally. Let me know if it still fails on your end.

jp524:
I can see the new test failing on CI - I'll have a look.

spec/system/banners/new_spec.rb
jp524: The test now fails in CI though. I think that the browser in CI is not picking up the timezone cookie because this is the assertion that fails (for test running at 5:08pm CST):

```
expected: "Value must be 06/11/2024, 05:08 PM or later."
            got: "Value must be 06/11/2024, 10:08 PM or later."
```

The time we are getting is in UTC, although CST is requested [here](https://github.com/rubyforgood/casa/blob/c0164bddf43d32ad26816636abc7319a6dee8477/spec/system/banners/new_spec.rb#L76).

spec/system/banners/new_spec.rb
```diff
@@ -49,14 +49,32 @@
...
+      page.driver.browser.manage.add_cookie(name: "browser_time_zone", value: "America/Chicago")
+      click_on "New Banner"
+      fill_in "Name", with: "Announcement"
+      fill_in "banner_expires_at", with: (current_time - 1.hour).strftime("%m%d%Y\t%I%M%P")
+      fill_in_rich_text_area "banner_content", with: "Please fill out this survey."
+      click_on "Submit"
+
+      message = page.find("#banner_expires_at").native.attribute("validationMessage")
+      expect(message).to eq("Value must be #{current_time.in_time_zone("America/Chicago").strftime("%m/%d/%Y, %I:%M %p or later.")}")
+    end
```
elasticspoon, line 77: I know that I suggested using `page.driver.browser.manage.add_cookie(name: "browser_time_zone", value: "America/Chicago")` but it doesn't seem to work as expected. It feels more tricky that its worth to test these error messages in system tests. I think we should simplify this test to just check that an expired banner cannot be created but not actually check the error.

I do think that checking the cookie logic is important and I think we should move those tests to requests specs. Off the top of my head the tests that matter for edge cases are:
- client in ahead of server, sets expired at time that is behind current time for client but ahead for server => this should be invalid
- client is behind of server, set expired at time that is ahead of current time for client but behind for server => this should be valid

Also when checking the error values it is much easier to read if they are in a form like 
```
expect(something).to have_text("Invalid time: must be after 1/1/1 12:00 PM")
``` 
instead of 
```
"Value must be #{current_time.in_time_zone("America/Chicago").strftime("%m/%d/%Y, %I:%M %p or later.")}"
```
That should be simple to do by just traveling to a specific date instead of one relative to today.

elasticspoon (changes requested):
Hey @jp524. Thanks for changes so far! Having poked around trying to make those systems tests work I think its better to just move those tests into request specs to simplify the testing.

jp524:
Thanks for your help @elasticspoon ! I implemented the changes and found a couple of bugs along the way so it was a good exercise.
Some of the CI workflows are failing but they don't give an actual error or test that is failing. I did run the test suite locally though and it all passes.

spec/requests/banners_spec.rb
```diff
@@ -71,4 +72,71 @@
...
+      it "fails" do
+        travel_to Time.new(2024, 6, 1, 11, 0, 0, "-04:00") do # 15:00 UTC
+          sign_in admin
+          post banners_path, params: {banner: banner_params}
+          expect(response).to render_template "banners/new"
+          expect(response.body).to include "Expires at must take place in the future (after 2024-06-01 15:00:00 UTC)"
+        end
+      end
+    end
+  end
```
elasticspoon, line 141: nit: i find these travel_to blocks kind of hard to read. I think it would be easier to understand in a form like:

```rb
context "when client timezone is ahead of UTC" do
  before { travel_to Time.new(2024, 6, 1, 11, 0, 0, "+03:00") } # 08:00 UTC

  context "when submitted time is behind client but ahead of UTC" do
    let(:expires_at) { Time.new(2024, 6, 1, 9, 0, 0, "UTC") } # 12:00 +03:00

    # it blocks....

  context "when submitted time is behind client and behind UTC" do
    let(:expires_at) { Time.new(2024, 6, 1, 7, 0, 0, "UTC") } # 10:00 +03:00

    # it blocks....
  end
end
```

elasticspoon (approved):
Nice job dealing with all the time zone stuff. Its a pain in the ass and I'm glad I didn't have to do it. 

Overall, I am happy with the PR. I had some minor nits but nothing blocking.

app/models/banner.rb
```diff
@@ -6,10 +6,14 @@ class Banner < ApplicationRecord
...
   validates_presence_of :name
+  validates_presence_of :content
   validate :only_one_banner_is_active_per_organization
+  validate :expires_at_must_be_in_future
 
   def expired?
-    expires_at && Time.current > expires_at
+    expired = expires_at && Time.current > expires_at
+    update(active: false) if active && expired
+    expired
```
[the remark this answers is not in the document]
jp524: Yeah I'm not a fan of the logic either but I couldn't think of another way to do it. The key for me was to ensure that we always had the latest `active` status set.

jp524:
Nits have been addressed! Should be good now @elasticspoon

PR #5848, opened 2024-06-21
feat: add timeout and concurrency limit to rspec GH action

elasticspoon:
We occasionally have timeouts with our RSpec Github Action. Lets save the planet a little and cancel it if it goes too long (20 mins)


compwron (approved):
legit

PR #5850, opened 2024-06-21
fix: authorize mark_as_read in Notification controller

elasticspoon:
When we upgraded Noticed we added skipping of callbacks to the NotificationsController. This resulted in issues with unauthed users hitting the notifications page.

To fix this we removed the skipping of callbacks, however, one of the callbacks being skipped was a callback to authorize the Notification. Authorization logic had never been added so now mark as read would not work.

This PR added authorization logic as well as request and policy specs for the controller.

Fixes theses errors:
https://app.bugsnag.com/ruby-for-good/casa/errors/6674487b0fc5c70008cbe9d0?filters%5Berror.status%5D=open&filters%5Bevent.since%5D=30d

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -1,13 +1,25 @@
 class NotificationsController < ApplicationController
+  after_action :verify_authorized
+  before_action :set_notification, only: %i[mark_as_read]
+
   def index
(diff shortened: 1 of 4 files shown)


elasticspoon:
Merging this because it's causing issues for users.

PR #5851, opened 2024-06-21
feat: add flipper for feature flags

elasticspoon:
I've added the `flipper` gem (https://www.flippercloud.io/docs) for managing feature toggles. I also put in `flipper-ui` for a nice web based interface for flipper that you can access at `/flipper`.

To do so, you need to have environment variables set for `FLIPPER_USERNAME` and `FLIPPER_PASSWORD`. You can set these in your `.env` file.

The defaults in QA will be username: `admin` and password: `password`. Production credentials will differ.


compwron:
Nice. Do keep in mind that we have an existing handrolled feature flag system that you should unify at some point

compwron:
https://github.com/rubyforgood/casa/blob/main/app/services/feature_flag_service.rb

elasticspoon:
I think I would also roll the fund request logic in, since there is not user facing way to toggle them.  https://github.com/rubyforgood/casa/blob/259dd5b73c35c81b7a072ec2c1496cc77a03f8a0/lib/tasks/deployment/20220409174149_enable_feature_flag_prince_george_fund_request.rake#L9

PR #5868, opened 2024-06-28
fix: n+1 issues in case_contacts#index

elasticspoon:
`CaseContactsController#index` was our worst page according to scout.

image
image

This was mainly because we had 2 N+1 queries:
image

Those have been fixed.

### DIFF
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -240,7 +240,7 @@ class CaseContact < ApplicationRecord
   end
 
   def requested_followup
-    followups.requested.first
+    followups.find(&:requested?)
   end
 
   def should_send_reimbursement_email?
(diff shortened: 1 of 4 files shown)


app/models/case_contact.rb
```diff
@@ -240,7 +240,7 @@ def contact_groups_with_types
   end
 
   def requested_followup
-    followups.requested.first
+    followups.find(&:requested?)
```
elasticspoon, line 243: We are including `followups` in the query already, however, the call to `.requested` forces another query `Followups.where(case_id: id, status: 0)` because it needs to check the status. Since all the data is already loaded it makes more sense to check that status in ruby rather than making another request.

schoork (approved):

compwron (approved):

elasticspoon:
image
:muscle:

PR #5871, opened 2024-06-30
5847 surface common filters

### LINKED ISSUE #5847: Surface common available filters for case contacts
 - supervisors
 - admins
 - volunteers

## Description

We recently added a feature for drafts of case contacts (`/case_contacts`). 

Image

This was added to make sure that volunteers would not lose progress towards their completion of the form. However, volunteers are having issues with accidentally creating drafts and those polluting their case contact screen.

Image

However, we have filters that would allow users to hide those drafts.

Image
Image

Volunteers were not aware that they could do that, thus, we should surface this functionality better. Lets show the most commonly used filters while the menu is still folded. Something like (THIS IS JUST AN EXAMPLE. ITS UGLY. PLEASE MAKE IT NICER :heart:):

Image

## Criteria for completion

- [ ] show hide drafts in folded state
- [ ] show sort order in folded state
- [ ] show clear and filter buttons in folded state
- [ ] write some system tests

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

We highly recommend that you join us in [slack](https://join.slack.com/t/rubyforgood/shared_invite/zt-2k5ezv241-Ia2Iac3amxDS8CuhOr69ZA) #casa channel to ask questions quickly. And [discord](https://discord.gg/qJcw2RZH8Q) for office hours (currently Tuesday 5-7pm Pacific), stakeholder news, and upcoming new issues.

mononoken:
Resolves #5847 

### What changed, and _why_?
- Moved the `Sorted by` and `Hide drafts` filter options outside of the collapsable area of the form element, so that they display regardless of filter menu collapse status.
- Fixed the positioning of the form select pseudo icons by refactoring the HTML syntax around the select elements
- Changed the filter-card-body collapse class logic to, fix a bug where the "Expand / show" button did not respond to the first click after filtering. This change accounts for how Bootstrap implements its version of collapse.

I did not add any new tests, since this was only a matter of changing options that are collapsed. However, I did update the button text and the tests that expected the old text.

I can add some system tests to check that the "sticky" options are always displayed if that is desired.

### Screenshots please :)
SCR-20240629-swmf
SCR-20240629-swrz

### Feelings gif (optional)
Instead of a gif, I asked ChatGPT to write a haiku about Bootstrap.
ブートストラップ
制約を感じて
自由を求め

### DIFF
--- a/app/javascript/controllers/collapse_controller.js
+++ b/app/javascript/controllers/collapse_controller.js
@@ -6,7 +6,7 @@ export default class extends Controller {
   static targets = ['collapsible']
 
   initialize () {
-    const isCollapsed = sessionStorage.getItem('filtersCollapsed') === 'true'
+    const isCollapsed = window.sessionStorage.getItem('filtersCollapsed') === 'true'
 
     if (isCollapsed) {
       this.collapsibleTarget.classList.add('collapse')
@@ -24,6 +24,6 @@ export default class extends Controller {
 
     this.collapsible.toggle()
 
-    sessionStorage.setItem('filtersCollapsed', isCollapsed ? 'false' : 'true')
+    window.sessionStorage.setItem('filtersCollapsed', isCollapsed ? 'false' : 'true')
   }
 }
--- a/app/javascript/controllers/collapse_controller.js
+++ b/app/javascript/controllers/collapse_controller.js
@@ -1,0 +1,29 @@
+import { Controller } from '@hotwired/stimulus'
+import { Collapse } from 'bootstrap'
+
+// Connects to data-controller="collapse"
+export default class extends Controller {
+  static targets = ['collapsible']
+
+  initialize () {
+    const isCollapsed = sessionStorage.getItem('filtersCollapsed') === 'true'
+
+    if (isCollapsed) {
+      this.collapsibleTarget.classList.add('collapse')
+    } else {
+      this.collapsibleTarget.classList.add('collapse', 'show')
+    }
+  }
+
+  connect () {
+    this.collapsible = Collapse.getOrCreateInstance(this.collapsibleTarget, { toggle: false })
+  }
+
+  toggle () {
+    const isCollapsed = !this.collapsibleTarget.classList.contains('show')
+
+    this.collapsible.toggle()
+
+    sessionStorage.setItem('filtersCollapsed', isCollapsed ? 'false' : 'true')
+  }
+}
--- a/app/javascript/controllers/collapse_controller.js
+++ b/app/javascript/controllers/collapse_controller.js
@@ -6,12 +6,12 @@ export default class extends Controller {
   static targets = ['collapsible']
 
   initialize () {
-    const isCollapsed = window.sessionStorage.getItem('filtersCollapsed') === 'true'
+    const isExpanded = window.sessionStorage.getItem('filtersCollapsed') === 'false'
 
-    if (isCollapsed) {
-      this.collapsibleTarget.classList.add('collapse')
-    } else {
+    if (isExpanded) {
       this.collapsibleTarget.classList.add('collapse', 'show')
+    } else {
+      this.collapsibleTarget.classList.add('collapse')
     }
   }
 
--- a/app/views/case_contacts/index.html.erb
+++ b/app/views/case_contacts/index.html.erb
@@ -43,9 +43,9 @@
           </div>
         </div>
         <div class="select-style-1 mb-0">
-          <div class="select-position pr-5">
-            <%= f.label :sorted_by %>
-            <%= f.select(:sorted_by, @filterrific.select_options[:sorted_by], {}, {class: "select-style-1 mb-0 bg-white"}) %>
+          <%= f.label :sorted_by %>
+          <div class="select-position">
+            <%= f.select(:sorted_by, @filterrific.select_options[:sorted_by], {}, {class: ""}) %>
           </div>
         </div>
         <div>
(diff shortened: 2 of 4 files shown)


app/views/case_contacts/index.html.erb
```diff
@@ -15,35 +15,47 @@
...
+          <%= button_tag( :class=> "btn-sm main-btn dark-btn") do %>
+            <i class="lni lni-funnel mr-10"></i> Filter
+          <% end %>
+          <%= link_to("Reset filters", reset_filterrific_url, class: "btn-sm main-btn dark-btn-outline btn-hove") %>
+        </div>
       </div>
+    </div>
+
+    <% collapse_class = params[:filterrific] ? "collapse show" : "collapse" %>
+    <div class="card-content mb-10 ml-20 <%= collapse_class %>" id="filter-card-body">
```
elasticspoon, line 58: This creates a small issue where as soon as you use a filter the whole menu pops open and we want to avoid that for the commonly used filters.

filter_popping_open

I'll leave it up to you to decide how to fix it :slightly_smiling_face:

mononoken: @elasticspoon To clarify, do you want the filter collapse state to be agnostic from all filters or just the commonly used filters? I am thinking the former is nicer.

Currently on main, if you nav to the page with filter params, the menu will be expanded. Should the menu always be collapsed on loading the page?

elasticspoon: > @elasticspoon To clarify, do you want the filter collapse state to be agnostic from all filters or just the commonly used filters? I am thinking the former is nicer.

I think agnostic from all. 

> Currently on main, if you nav to the page with filter params, the menu will be expanded. Should the menu always be collapsed on loading the page?

So I think ideally: 
- I navigate to page => filter is closed
- I select a common filter, hit filter button => page updates filter menu remains closed
- I open the menu, select a some other filter, hit filter button => page updates, filter menu remains open

You _might_ have to add a turbo frame for this? Cause I am not sure how you could preserve filter openness state if a full page refresh is happening. 

If that is not possible I would rather it default to always closed.

mononoken: @elasticspoon I could see a turbo frame being beneficial for this case, but it looks like casa is not using turbo-rails. I could look at a solution that adds it, but maybe that is outside the scope of this PR?

I thought of a different approach: 28c8b1859b732650356527c9253fdbf33e8080c6. I made a Stimulus controller that encapsulates some of the BS collapse behavior, and it stores the state of the collapse menu in sessionStorage. 

> * I navigate to page => filter is closed
> * I select a common filter, hit filter button => page updates filter menu remains closed
> * I open the menu, select a some other filter, hit filter button => page updates, filter menu remains open

This fulfills the 2nd and 3rd cases you mentioned above, but for the 1st, instead the filter is open/closed depending on what the user left it as in the session.

I would be happy to look at adding turbo frames instead. In that approach, I don't think we would have to worry about storing the collapse state, since we would no longer be re-rendering the whole page.

elasticspoon: > I could see a turbo frame being beneficial for this case, but it looks like casa is not using turbo-rails.

Well then.... I am a bit embarrassed to learn that just now :blush: 

> I made a Stimulus controller that encapsulates some of the BS collapse behavior, and it stores the state of the collapse menu in sessionStorage.

That sounds pretty reasonable. I'll take a closer look soon. I would like to get this update in sooner rather than later. So while I do think turbo would provide the more ideal solution but I would much rather have the fix sooner. 

It would be dope if after this PR you could add in turbo-rails, but it is certainly out of the scope of **this** PR.

mononoken: > > I could see a turbo frame being beneficial for this case, but it looks like casa is not using turbo-rails.
> 
> Well then.... I am a bit embarrassed to learn that just now 😊
> 
> > I made a Stimulus controller that encapsulates some of the BS collapse behavior, and it stores the state of the collapse menu in sessionStorage.
> 
> That sounds pretty reasonable. I'll take a closer look soon. I would like to get this update in sooner rather than later. So while I do think turbo would provide the more ideal solution but I would much rather have the fix sooner.
> 
> It would be dope if after this PR you could add in turbo-rails, but it is certainly out of the scope of **this** PR.
😊 def understandable! There is a lot going on with this project, and I'm sure some parts are fairly old.

app/javascript/controllers/collapse_controller.js
```diff
@@ -0,0 +1,29 @@
+import { Controller } from '@hotwired/stimulus'
+import { Collapse } from 'bootstrap'
+
+// Connects to data-controller="collapse"
+export default class extends Controller {
+  static targets = ['collapsible']
+
+  initialize () {
+    const isExpanded = window.sessionStorage.getItem('filtersCollapsed') === 'false'
```
schoork, line 9: This should incorporate a key that is settable in the view template, so it can be used on other pages in the app.

app/views/case_contacts/index.html.erb
schoork: I'm not sure I like session storage as the solution here. It overcomplicates this. The only reason is so that we can preserve the filters being opened after a user clicks the filter button, right? 

I agree that a turbo-frame is a better long-term fix and might be out of scope for this issue.

app/views/case_contacts/index.html.erb
```diff
@@ -15,35 +15,46 @@
...
-      <div class="row">
-        <div class="col-12">
-          <h3 class="mt-10"><label>Draft status</label></h3>
-        </div>
-        <div class="col-md-4">
+  <div class="card-style" data-controller="collapse">
+    <div class="card-content">
+      <div class="d-flex justify-content-between align-items-center">
+        <div class="h4 btn-lg">Filter by</div>
+        <button href="#"
```
schoork, line 25: If you add this attribute to the button, you don't need stimulus at all here.
```
data-bs-toggle="collapse"
```

app/javascript/controllers/collapse_controller.js
schoork, line 1: This controller feels unnecessary. If you don't use session storage to persist the state, then you can use html attributes to get the same functionality in a much simpler (and more Bootstrappy way).

schoork (changes requested):
Most of my requested changes hinge on not using session storage for persisting the state. However, there is a key that should be added to the controller if we stick with it.

app/views/case_contacts/index.html.erb
mononoken: @schoork thank you for the comments! I agree I am not happy with the session storage solution, but I wasn't sure of a cleaner way to persist client state between form submissions without it.

> The only reason is so that we can preserve the filters being opened after a user clicks the filter button, right?

That is correct.

I am unclear on the direction to take with all your comments together. To clarify, it seems to me that the two solutions you propose are either:

1. Revert to my previous solution. Not persist the collapse menu state and remove the Stimulus controller and just use Bootstrap for collapse behavior. Only downside really is that every time the user applies filters, the menu will collapse.
2. Refactor the Stimulus controller to set a key for the session storage. If Turbo is added, this controller will probably no longer be needed and deleted.

Is that correct? And if so, which of these is preferred by the project team?

It seems to me that 1 is cleaner for developer happiness, but 2 may have a slightly higher user happiness. And if turbo is added, the problem with either solution is temporary. I am happy to go with whatever solution y'all prefer.

mononoken: Maybe I am over-complicating this! Another solution I am thinking of is:
- Revert to https://github.com/rubyforgood/casa/commit/0f61c0ed4a74c058c58b31719cc6cef278c2cea3
- No Stimulus controller, just bootstrap classes
- When checking for filterrific params to determine collapse state, ignore the "surfaced filter" params. I could just reject them from the hash.

That way, we don't expand the menu when just checking "Hide draft". The only downside is that if you were for some reason expanding the menu and only submitted hide draft, it would collapse, but I think that is a niche situation.

elasticspoon:
closing in favor of https://github.com/rubyforgood/casa/pull/5879

PR #5918, opened 2024-07-16
draft case contact cleanup

### LINKED ISSUE #5442: Periodically cleanup drafts with no case id
None

## Description
#5414 introduced the potential for case contact drafts to not be attached to a Casa Case (or cases). In this situation, if a user leaves step 1 without selecting at least one casa case, they will not be able to return to that draft. 

We need to introduce a way to clean these up periodically. They should be any case contacts that meet the following criteria:

1. `draft_case_ids` is empty
2. case contact was created more than 1 day go

As part of this work, 

- You may need to introduce a tool to run rake tasks or jobs periodically. One such tool is the [whenever gem](https://github.com/javan/whenever) (but you could use a different one). 
- If introduces a tool (like above) then incorporate with #5328 to make sure this runs on the 15th of every month

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

We highly recommend that you join us in [slack](http://bit.ly/3Quxc1Q) #casa channel to ask questions quickly. And [discord](https://discord.gg/qJcw2RZH8Q) for office hours (currently Tuesday 5-7pm Pacific), stakeholder news, and upcoming new issues.

Smullz622:
Resolves #5442 

### What changed, and _why_?
Added a scope to CaseContact to return drafts without an attached case as well as drafts older than one week. The rake task actually deletes the case contact. The task will need to be added to Heroku task scheduler.

There's a model test on CaseContact that tests different scenarios and ensures only case contacts that should be cleaned up are returned.

### Screenshots please :)
Before running rake task:
Screenshot 2024-07-16 at 11 20 22 AM
After running rake task:
Screenshot 2024-07-16 at 11 21 04 AM

### Feelings gif (optional)

moira rose

### DIFF
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -148,7 +148,7 @@ class CaseContact < ApplicationRecord
   scope :drafts_for_removal, -> {
     where.not(status: "active")
       .where("case_contacts.created_at < ?", 1.week.ago)
-      .or(where(draft_case_ids: nil).where("case_contacts.created_at < ?", 1.day.ago))
+      .or(where(draft_case_ids: []).where("case_contacts.created_at < ?", 1.day.ago))
   }
 
   filterrific(
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -145,12 +145,6 @@ class CaseContact < ApplicationRecord
 
   scope :no_drafts, ->(checked) { (checked == 1) ? where(status: "active") : all }
 
-  scope :drafts_for_removal, -> {
-    where.not(status: "active")
-      .where("case_contacts.created_at < ?", 1.week.ago)
-      .or(where(draft_case_ids: []).where("case_contacts.created_at < ?", 1.day.ago))
-  }
-
   filterrific(
     default_filter_params: {sorted_by: "occurred_at_desc"},
     available_filters: [
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -145,6 +145,12 @@ class CaseContact < ApplicationRecord
 
   scope :no_drafts, ->(checked) { (checked == 1) ? where(status: "active") : all }
 
+  scope :drafts_for_removal, -> {
+    where.not(status: "active")
+      .where("case_contacts.created_at < ?", 1.week.ago)
+      .or(where(draft_case_ids: nil).where("case_contacts.created_at < ?", 1.day.ago))
+  }
+
   filterrific(
     default_filter_params: {sorted_by: "occurred_at_desc"},
     available_filters: [
--- a/app/services/draft_case_contact_cleanup_service.rb
+++ b/app/services/draft_case_contact_cleanup_service.rb
@@ -1,9 +1,10 @@
 class DraftCaseContactCleanupService
+  ORPHANED_DRAFT_EXPIRATION_DELAY = 1.week.ago
+  ORPHANED_DRAFT_EXPIRATION_DELAY_WITHOUT_CASE = 1.day.ago
+
   def self.call
-    draft_expiration = 1.week.ago
-    draft_expiration_without_case = 1.day.ago
-    drafts_for_removal = CaseContact.where.not(status: "active").where("case_contacts.created_at < ?", draft_expiration)
-      .or(CaseContact.where.not(status: "active").where(draft_case_ids: []).where("case_contacts.created_at < ?", draft_expiration_without_case))
+    drafts_for_removal = CaseContact.where.not(status: "active").where("case_contacts.created_at < ?", ORPHANED_DRAFT_EXPIRATION_DELAY)
+      .or(CaseContact.where.not(status: "active").where(draft_case_ids: []).where("case_contacts.created_at < ?", ORPHANED_DRAFT_EXPIRATION_DELAY_WITHOUT_CASE))
 
     drafts_for_removal.each do |draft|
       draft.destroy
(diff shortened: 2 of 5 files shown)


lib/tasks/case_contact_draft_cleanup.rake
```diff
@@ -0,0 +1,6 @@
+desc "delete draft case contacts marked for clean up, run by heroku scheduler."
```
compwron, line 1: good desc

compwron: for later PR thought - we may want to move away from heroku scheduler towards something like 

```
      include Sidekiq::Job
      sidekiq_options queue: :general_12h

      def self.crontab
        "0 8 * * *"
      end
```

Smullz622: I'm not that familiar with heroku scheduler but would like to learn so I'm curious, what's the rationale for moving away from it in favor of Sidekiq?

app/models/case_contact.rb
```diff
@@ -145,6 +145,12 @@ def active_or_expenses?
 
   scope :no_drafts, ->(checked) { (checked == 1) ? where(status: "active") : all }
 
+  scope :drafts_for_removal, -> {
+    where.not(status: "active")
+      .where("case_contacts.created_at < ?", 1.week.ago)
+      .or(where(draft_case_ids: []).where("case_contacts.created_at < ?", 1.day.ago))
```
elasticspoon, line 151: I'm not sure what the best names would be but can we turn `1.week.ago` into named constants as opposed to magic values?

app/models/case_contact.rb
```diff
@@ -145,6 +145,12 @@ def active_or_expenses?
...
+  }
+
```
elasticspoon, line 153: We aren't the most consistent with it in this code base but I think we should keep business logic out of our scopes. In this case the designation of what is "for removal" is that business logic. 

Most of the other scopes are pretty generic.

```rb
scope :created_max_ago, ->(time_range = nil) {
  where("case_contacts.created_at > ?", time_range) if time_range.present?
}
```

I think this query belongs in a service object or maybe just some plain ruby object.

spec/models/case_contact_spec.rb
```diff
@@ -547,6 +547,19 @@
         end
       end
     end
+
+    describe ".drafts_for_removal" do
+      let!(:active_case_contact) { create(:case_contact, status: "active") }
+      let!(:new_draft_case_contact) { create(:case_contact, status: "started", created_at: Time.now, draft_case_ids: []) }
+      let!(:draft_case_contact_without_draft_case_id) { create(:case_contact, status: "started", created_at: Time.now - 2.days, draft_case_ids: []) }
+      let!(:draft_case_contact_with_draft_case_id) { create(:case_contact, status: "started", created_at: Time.now - 2.days, draft_case_ids: [1]) }
+      let!(:eight_day_draft_case_contact) { create(:case_contact, status: "details", created_at: Time.now - 8.days) }
```
elasticspoon, line 556: can this also be `status: "started"` for consistency?

spec/models/case_contact_spec.rb
elasticspoon, line 1: Good test :+1:

elasticspoon (changes requested):
Most of my feedback is pretty minor BUT I took a closer look at the production data and want to double check that stakeholders are fine with the deletion timelines. (The request for changes is mainly to ensure this does not get  accidentally merged.)

There are a surprising amount of case contacts that are in draft state that have good info in them but are old. I am not sure we actually want to delete those.

app/models/case_contact.rb
Smullz622: will do

app/models/case_contact.rb
Smullz622: Got it. I'll move it into a service class and do the destroy there as well so the rake task just calls the service method. I'll update tests accordingly

spec/models/case_contact_spec.rb
Smullz622: I actually did that intentionally to mimic a draft that was saved correctly after step one (moving it to 'details') but was then abandoned & was over a week old to cover the second half of the ticket (remove any drafts older than a week)

spec/services/draft_case_contact_cleanup_service_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+
+RSpec.describe DraftCaseContactCleanupService do
+  describe ".call" do
+    let!(:past_expiration) { Time.now - 8.days }
+    let!(:past_exp_without_case) { Time.now - 2.days }
```
elasticspoon, line 6: You can also do `8.days.ago`, etc.

app/services/draft_case_contact_cleanup_service.rb
```diff
@@ -0,0 +1,12 @@
+class DraftCaseContactCleanupService
+  def self.call
+    draft_expiration = 1.week.ago
+    draft_expiration_without_case = 1.day.ago
```
elasticspoon, line 4: I meant constants like `DRAFT_EXPIRATION_DELAY` and `ORPHANED_DRAFT_EXPIRATION_DELAY`

Not sure if orphaned is a better name

elasticspoon (changes requested):
Overall it looks good to me. But I will pause on making a final merge decision until next week (we have a stakeholder meeting friday).

compwron:
Cool stuff :)

PR #5930, opened 2024-07-20
5906 update renders for turbo

### LINKED ISSUE #5906: add turbo-rails
Steve Ballmer DEVELOPERS DEVELOPERS DEVELOPERS

**Description**

We would like to have more SPA (Single Page Application) esque behavior in parts of the application. The common way to do this in Rails is via Hotwire https://hotwired.dev/. 

We currently already have Stimulus so lets add Turbo.

### Criteria for Completion

- [ ] Add the turbo-rails gem https://github.com/hotwired/turbo-rails
- [ ] check if anything else needs to be added
- [ ] add a tiny bit of turbo somewhere so that we can be sure that it works

_QA Login Details:_  

Login Emails: 

### Questions? Join Slack!

mononoken:
This does not resolve #5906, but is the first part in completing it.

### What changed, and _why_?
Turbo expects all form failures to return 422 (unprocessable entity). This used to not be required in Rails, so many renders from form failures just returned 200 by default. This PR adds 422 to all render messages in controllers for form failures.

Beyond the Turbo requirement, this is a better practice and should not break any old behavior.

I update tests that are affected by the changed message. Many tests were asserting for `were_successful` or similar but now they properly expect 422. 

### Screenshots please :)
Note the response in the network tab:
image

### DIFF
--- a/app/controllers/all_casa_admins/casa_admins_controller.rb
+++ b/app/controllers/all_casa_admins/casa_admins_controller.rb
@@ -12,7 +12,7 @@ class AllCasaAdmins::CasaAdminsController < AllCasaAdminsController
       service.create!
       redirect_to all_casa_admins_casa_org_path(@casa_org), notice: "New admin created successfully"
     rescue ActiveRecord::RecordInvalid
-      render :new
+      render :new, status: :unprocessable_entity
     end
   end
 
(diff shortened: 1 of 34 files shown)


elasticspoon (approved):
You can also do `expect(response).to be_unprocessable`

PR #5938, opened 2024-07-22
change div classes to fix checkbox location

### LINKED ISSUE #5729: Fix send_reminder_button
The partial `volunteers/_send_reminder_button.html.erb` needs fixing. 

image

It has a label(1) and a checkbox(2). They should not be separate. Consider turning the partial into a ViewComponent.

Smullz622:
Resolves #5729 

### What changed, and _why_?
I changed the div classes to correct the alignment and make sure the checkbox is next to the checkbox label.

There are already tests for the functionality in `spec/system/volunteers/edit_spec.rb:470-517`

### Screenshots please :)
image

### Feelings gif (optional)
alt text

### DIFF
--- a/app/views/volunteers/_send_reminder_button.html.erb
+++ b/app/views/volunteers/_send_reminder_button.html.erb
@@ -1,14 +1,16 @@
-<%= button_tag(
-  type: "submit",
-  class: "main-btn primary-btn btn-hover btn-sm",
-  id: "reminder_button",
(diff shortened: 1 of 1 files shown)


elasticspoon (approved):
Thank you! Those check boxes have bothered me for so long. 

It is not a blocker but you think you could add a small gap (4 ish px) between the checkbox and the label?

compwron:
:)

PR #5943, opened 2024-07-24
Remove littleforest from codeowners

littleforest:
Removing myself from CODEOWNERS for the time being since I haven't been that active in reviewing code recently.


elasticspoon (approved):
:crying_cat_face:

Issue #5563, opened 2024-03-08
☂️ Replace stub `pending` specs with real tests


compwron:
**What type(s) of user does this feature affect?**
![Steve Ballmer DEVELOPERS DEVELOPERS DEVELOPERS](https://i.makeagif.com/media/7-29-2015/OA5hkn.gif)

**Status (updated 2026-05-02):** the original framing of this umbrella is stale. `.allow_skipping_tests` is now empty — that gate is satisfied. But it was emptied in #6517 (Oct 2025) by stubbing 52 spec files with `pending "add some tests for X"` placeholders rather than by writing real tests. The umbrella's actual goal — _real_ tests for those files — is not done.

**41 stub specs still remain.** Find them with:

```bash
grep -rl 'pending "add some tests for' spec/
```

**Definition of done for one stub:**
- Delete the `pending "add some tests for X"` line.
- Add at least one real `it` / `describe` block that exercises a behavior of the class/module under test.
- Make sure the spec actually runs — no `pending`, `skip`, or `xit` left in the file.
- One stub file per PR is fine; bundle several if they're tiny or share a theme.

**Reference:** the placeholder pattern was introduced in #6517. Anything matching the grep above is fair game.

---

_QA Login Details:_
[Link to QA site](https://casa-qa.herokuapp.com/)

Login Emails:
- volunteer1@example.com  view site as a volunteer
- supervisor1@example.com view site as a supervisor
- casa_admin1@example.com view site as an admin

password for all users: 12345678

### Questions? Join Slack!

We highly recommend that you join us in slack https://rubyforgood.herokuapp.com/ #casa channel to ask questions quickly and hear about office hours (currently Wednesday 6-8pm Pacific), stakeholder news, and upcoming new issues.

blitzblade:
I could not login to the QA site with the email: volunteer1@example.com and password: 123456

compwron:
@blitzblade sorry about that, try using password 12345678 !

alancovarrubias:
I can work on this

compwron:
Refreshed the title and body to match the current state — the original `.allow_skipping_tests` framing was misleading because that file is now empty. The real remaining work is to replace the 41 `pending "add some tests for X"` stub specs (introduced in #6517) with real tests.

@alancovarrubias @pbbyrro — you both expressed interest previously; if you're still up for it, the grep one-liner in the issue body will give you the current list of stubs to pick from.

fuentesjr:
[...]

PR #5944, opened 2024-07-25
BulkCourtDateController spec implemented

thejonroberts:
Addresses #5563

### What changed, and _why_?
- Ran `bin/rails g rspec:controller BulkCourtDates new create` to create spec file for the BulkCourtDatesController.
  - this also created a pending spec for bulk_court_dates/new view. I kept but did not add to it.
- Removed `controllers/bulk_court_dates_controller.rb` from `.allow_skipping_tests` file
- Modified CaseGroup Factory to allow multiple associated cases (:case_count transient attribute)
- Wrote specs for the controller's behavior

It is tested well! 😉   It's all tests.  

I added some shared examples for user & org authorization - may be overkill or there may be a different implementation somewhere.  I'm open to suggestions here, implementation or especially where to put them - currently in  `spec/support/shared_examples/requests/action_shared_examples.rb`

### Feelings gif (optional)
alt text

### DIFF
--- a/.allow_skipping_tests
+++ b/.allow_skipping_tests
@@ -1,7 +1,6 @@
 blueprints/api/v1/session_blueprint.rb
 channels/application_cable/channel.rb
 channels/application_cable/connection.rb
-controllers/bulk_court_dates_controller.rb
 controllers/case_groups_controller.rb
 controllers/concerns/court_date_params.rb
 controllers/followup_reports_controller.rb
--- a/app/controllers/bulk_court_dates_controller.rb
+++ b/app/controllers/bulk_court_dates_controller.rb
@@ -4,13 +4,13 @@ class BulkCourtDatesController < ApplicationController
   before_action :require_organization!
 
   def new
-    authorize :application, :admin_or_supervisor?
+    authorize :bulk_court_date, :new?
 
     @court_date = CourtDate.new
   end
 
   def create
-    authorize :application, :admin_or_supervisor?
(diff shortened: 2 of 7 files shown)


elasticspoon:
hey. typically the pattern that we use is:

controller: https://github.com/rubyforgood/casa/blob/main/app/controllers/contact_topics_controller.rb

which controls auth to restful actions via 

policy: https://github.com/rubyforgood/casa/blob/main/app/policies/contact_topic_policy.rb

that policy is tested for the auth related things: https://github.com/rubyforgood/casa/blob/main/spec/policies/contact_topic_policy_spec.rb


in this situation the controller is using the application policy in a hacky way. It would be really nice if you could rework this to follow existing patterns:
- add a specific policy for the bulk court dates
- replace the policy in the contoller
- test the policy
- remove the shared example tests

compwron (approved):
imo perfect is the enemy of good, we should merge and improve  :)

thejonroberts:
@elasticspoon ah, using pundit, that makes sense.  I can implement proper policies w/ specs and use them in the controller.

thejonroberts:
@elasticspoon updated with BulkCourtDatePolicy - there is no BulkCourtDate model, so it is a headless policy. Also, there is no CaseGroupPolicy, so added a spec for current behavior when user tries to update a CaseGroup from a different casa org.

elasticspoon (approved):
I'm good with this! Thank you!

PR #5948, opened 2024-07-28
Fix flaky CaseContactReportSpec

thejonroberts:
No issue - flaky spec seen in PR CI runs. ([here](https://github.com/rubyforgood/casa/actions/runs/10113566490/job/27970214915?pr=5946) and [here](https://github.com/rubyforgood/casa/actions/runs/10100601593/job/27932322976?pr=5944), for example)

### What changed, and _why_?
Changed spec to look for row in array of expected results rather than check each row.  Spec previously assumed created_at ordering, but the report does not explicitly enforce an order, afaik.

I ran full suite of a known bad seed (24693), and these changes passed.

### DIFF
--- a/spec/models/case_contact_report_spec.rb
+++ b/spec/models/case_contact_report_spec.rb
@@ -409,9 +409,15 @@ RSpec.describe CaseContactReport, type: :model do
     end
 
     it "includes topic answers in csv rows" do
-      expect(csv[0].fields).to eq ["Ans Contact 1 Topic 1", "Ans Contact 1 Topic 2"]
-      expect(csv[1].fields).to eq [nil,                     "Ans Contact 2 Topic 2"]
-      expect(csv[2].fields).to eq [nil,                     nil]
+      expected_rows = [
+        # ['Used topic 1',        'Used topic 2'] (header)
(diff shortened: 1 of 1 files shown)


elasticspoon:
I need to double check this. This code was a pain in the ass to write and very complicated. I want to say there might have been an order that we were going for. 

@solebared any thoughts on this?

thejonroberts:
> I need to double check this. This code was a pain in the ass to write and very complicated. I want to say there might have been an order that we were going for.
> 
> @solebared any thoughts on this?

Okay, if there is an order desired (created_at / :added_to_system_at seems likely?), I could try to fix based on the known bad seed, just let me know.

elasticspoon (approved):
I thought about this some more and yea. its the order within the rows that matters not the rows themselves so your solution is good.

solebared:
Apologies for the slow response. I don't remember OTTOMH whether the csv rows are expected to be returned in a particular order. I can try to dig into that later but agree with @elasticspoon that the order of columns (within a row) is more significant. Ofc the header has to be the first row but we're testing that separately 👍🏾 .

PR #5957, opened 2024-07-31
Case contact form 'create another' checkbox

### LINKED ISSUE #5925: add button to "create another" case contact when submitting case contact
 - supervisors
 - admins
 - volunteers

## Description

Often times our volunteers will create case contacts (`/case_contacts/:id:/form`) in bulk after the actual contact. We would like to improve the process of creating bulk contacts for them.

Currently, after a case contact is created the land back on the Case page and have to restart that process from scratch. Lets implement something like linear or jira when creating multiple issues.

Image

When a user is on the final page submitting their case contact they should have the option to "Create Another". This will take them right back to the start of the form after creating the Case Contact so they can start directly on another contact.

This new case contact should retain some of the values of the case contact that gets created such as the case with which it is associated but everything else should get cleared out.

## Criteria for Completion

- [ ] Add a "Create Another" toggle or checkbox on the final page of the `/case_contacts/:id/form`
- [ ] Redirect the user back to the first page of the case contact form if it is submitted with  "Create Another" selected
- [ ] After redirection makes sure all the fields are cleared out except the case  
- [ ] Add some sort of a popup or descriptive tooltip so that users have a better idea of what the button does
- [ ] case contacts have a `jsonb` `metadata` field. when a new case contact is created store whether it was created via "Create Another" or not so that we have metrics

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

thejonroberts:
Resolves #5925  - Add "create another" option to case contact form

### What changed, and _why_?
Adds a checkbox to the last step of case contact form, which indicates user wants to create another case contact for the same case(s). After submission, if checked, the user is redirected to the case contact form's first step with the same draft_case_ids as the previously created case contact.

Also added a CaseContact scope to view how many times this has been used. There may be a more ActiveRecord-y way to deal with the jsonb?

- spec/requests/case_contacts/form_spec.rb tests the changes to the form controller - that requests are redirected to correct path with draft case ids param.
- system specs for CaseContact new & edit
- I refactored the specs because there was a lot of duplicate setup and misleading descriptions, sorry for the huge diff there.

### Screenshots please :)
Screenshot 2024-07-31 at 3 17 21 PM

### Feelings gif (optional)
Fred Willard playing and old timey novelty organ at a funeral from I Think You Should Leave

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -101,7 +101,7 @@ class ApplicationController < ActionController::Base
 
   def store_referring_location
     if request.referer && !request.referer.end_with?("users/sign_in")
-      session[:return_to] = request.referer
+      session[:return_to] = request.referer unless params[:ignore_referer]
     end
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -100,8 +100,8 @@ class ApplicationController < ActionController::Base
   end
 
   def store_referring_location
-    if request.referer && !request.referer.end_with?("users/sign_in")
-      session[:return_to] = request.referer unless params[:ignore_referer]
+    if request.referer && !request.referer.end_with?("users/sign_in") && params[:ignore_referer].blank?
+      session[:return_to] = request.referer
     end
   end
 
--- a/app/controllers/case_contacts/form_controller.rb
+++ b/app/controllers/case_contacts/form_controller.rb
@@ -83,7 +83,7 @@ class CaseContacts::FormController < ApplicationController
     update_volunteer_address(@case_contact)
     flash[:notice] = message
     if @case_contact.metadata["create_another"]
-      redirect_to new_case_contact_path(params: {draft_case_ids:})
+      redirect_to new_case_contact_path(params: {draft_case_ids:, ignore_referrer: true})
     else
       redirect_back_to_referer(fallback_location: case_contacts_path(success: true))
     end
--- a/app/controllers/case_contacts/form_controller.rb
+++ b/app/controllers/case_contacts/form_controller.rb
@@ -77,7 +77,7 @@ class CaseContacts::FormController < ApplicationController
     else
       message = "Case #{"contact".pluralize(draft_case_ids.count)} successfully created."
       create_additional_case_contacts(@case_contact)
-      first_casa_case_id = draft_case_ids.slice(0)
+      first_casa_case_id = draft_case_ids.first
       @case_contact.update(status: "active", draft_case_ids: [first_casa_case_id], casa_case_id: first_casa_case_id)
     end
     update_volunteer_address(@case_contact)
--- a/app/controllers/case_contacts/form_controller.rb
+++ b/app/controllers/case_contacts/form_controller.rb
@@ -83,7 +83,7 @@ class CaseContacts::FormController < ApplicationController
     update_volunteer_address(@case_contact)
     flash[:notice] = message
     if @case_contact.metadata["create_another"]
-      redirect_to new_case_contact_path(params: {draft_case_ids:, ignore_referrer: true})
+      redirect_to new_case_contact_path(params: {draft_case_ids:, ignore_referer: true})
     else
       redirect_back_to_referer(fallback_location: case_contacts_path(success: true))
     end
--- a/app/controllers/case_contacts/form_controller.rb
+++ b/app/controllers/case_contacts/form_controller.rb
@@ -71,17 +71,22 @@ class CaseContacts::FormController < ApplicationController
   def finish_editing
     message = ""
     send_reimbursement_email(@case_contact)
+    draft_case_ids = @case_contact.draft_case_ids
     if @case_contact.active?
       message = @case_contact.decorate.form_updated_message
     else
-      message = "Case #{"contact".pluralize(@case_contact.draft_case_ids.count)} successfully created."
+      message = "Case #{"contact".pluralize(draft_case_ids.count)} successfully created."
       create_additional_case_contacts(@case_contact)
-      first_casa_case_id = @case_contact.draft_case_ids.slice(0)
+      first_casa_case_id = draft_case_ids.slice(0)
       @case_contact.update(status: "active", draft_case_ids: [first_casa_case_id], casa_case_id: first_casa_case_id)
     end
     update_volunteer_address(@case_contact)
     flash[:notice] = message
-    redirect_back_to_referer(fallback_location: case_contacts_path(success: true))
+    if @case_contact.metadata["create_another"]
+      redirect_to new_case_contact_path(params: {draft_case_ids:})
+    else
+      redirect_back_to_referer(fallback_location: case_contacts_path(success: true))
+    end
   end
 
   def send_reimbursement_email(case_contact)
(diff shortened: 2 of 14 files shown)


thejonroberts:
reworking this to use the metadata field rather than passing params around (🤮). have the metadata updating and a scope to find records with it.

lib/tasks/data_post_processors/case_contact_populator.rb
```diff
@@ -3,7 +3,7 @@ def self.populate
     CaseContact.find_each do |case_contact|
       # Get rid of drafts
       unless case_contact.casa_case
-        case_contact.destroy
+        case_contact.destroy!
```
thejonroberts, line 6: unrelated, fixed bin/update issue, can drop commit.

app/decorators/contact_type_decorator.rb
```diff
@@ -13,6 +13,6 @@ def hash_for_multi_select_with_cases(casa_case_ids)
   def last_time_used_with_cases(casa_case_ids)
     last_contact = CaseContact.joins(:contact_types).where(casa_case_id: casa_case_ids, contact_types: {id: object.id}).order(occurred_at: :desc).first
 
-    last_contact.nil? ? "never" : "#{time_ago_in_words(last_contact.occurred_at)} ago"
+    last_contact&.occurred_at.blank? ? "never" : "#{time_ago_in_words(last_contact.occurred_at)} ago"
```
thejonroberts, line 16: had to do this to view the form page, it was erroring

elasticspoon: We should add a test for that.

app/views/case_contacts/form/expenses.html.erb
```diff
@@ -97,6 +97,12 @@
     <% end %>
 
     <div class="actions mb-10 case-contacts-form-buttons">
+      <div class="case-contacts-form-options checkbox-style flex-grow-1">
+        <%= form.fields_for :metadata do |metadata_form| %>
+          <%= metadata_form.check_box :create_another, class: 'form-check-input ' %>
+          <%= metadata_form.label :create_another,  "Create Another", class: 'form-check-label ' %>
+        <% end %>
+      </div>
```
elasticspoon, line 105: I am a bit worried users might not notice this button, can we slide it all the way right next to the submit button?

elasticspoon (changes requested):
Ran into an issue testing this locally:

Created a contact with create another toggled =>
Landed back at create page with prefilled fields (all good) =>
Fill out the contact but did not select create another =>
This did not take me back to the index page the case was created but I stayed on the expenses page.

Peek 2024-08-03 16-18

thejonroberts:
redirect issue is probably a consequence of `CaseContractsController#new`:
```ruby
def new
  store_referring_location
```
so the form 3rd step is stored (when create another checked), redirected to next time create another is not checked. Will look into how to skip storing the location if coming from the create another flow...

app/views/case_contacts/form/expenses.html.erb
thejonroberts: Screenshot 2024-08-04 at 12 07 56 AM

thejonroberts:
@elasticspoon changes addressed, ready for re-review.  I refactored some specs in the process, I can revert and put into a separate PR if necessary.

thejonroberts:
Added tooltip (requirement in issue that I forgot about!)
Screenshot 2024-08-08 at 9 38 00 PM

app/controllers/application_controller.rb
```diff
@@ -101,7 +101,7 @@ def sms_acct_creation_notice(resource_name, sms_status)
 
   def store_referring_location
     if request.referer && !request.referer.end_with?("users/sign_in")
-      session[:return_to] = request.referer
+      session[:return_to] = request.referer unless params[:ignore_referer]
```
elasticspoon, line 104: I think the unless condition probably belongs in the if statement.

In the future I think the `users/sign_in` should have that same param rather than using a check for the URL.

app/controllers/case_contacts/form_controller.rb
elasticspoon, line 1: Personal Thoughts: This controller has gotten so complex. At some point we really need to move the code out of it and into something that makes more sense.

app/controllers/case_contacts_controller.rb
```diff
@@ -35,18 +35,8 @@ def new
...
     casa_cases = policy_scope(current_organization.casa_cases)
-    draft_case_ids =
-      if params.dig(:case_contact, :casa_case_id).present?
-        casa_cases.where(id: params.dig(:case_contact, :casa_case_id)).pluck(:id)
-      elsif casa_cases.count == 1
-        [casa_cases.first.id]
-      else
-        []
-      end
+    draft_case_ids = set_draft_case_ids(params, casa_cases)
```
elasticspoon, line 39: Nit: `build_draft_case_ids` maybe? `set` sounds like it is doing the setting via a side effect.

spec/system/case_contacts/edit_spec.rb
elasticspoon, line 1: We should also check that it has the right case pre-selected.

app/views/case_contacts/form/expenses.html.erb
```diff
@@ -97,6 +97,17 @@
...
+        <%= form.fields_for :metadata do |metadata_form| %>
+          <%= metadata_form.check_box :create_another, class: "form-check-input" %>
+          <%= metadata_form.label :create_another, class: "form-check-label" do %>
+            Create Another
+            <i class="lni lni-question-circle" data-toggle="tooltip" data-placement="top"
+              title="Start a new contact for the same case(s) after submitting.">
+            </i>
+          <% end %>
+        <% end %>
+      </div>
```
elasticspoon, line 110: styling looks off on my screen 
image

elasticspoon (approved):
Looks good to me. The volunteers will really appreciate this, bulk filling out contacts is a big pain point for them. 

None of my comments are blocking, mainly future notes and minor nits.

Thanks!

app/controllers/case_contacts_controller.rb
thejonroberts: agree, can change

app/views/case_contacts/form/expenses.html.erb
thejonroberts: dang! yeah, that really bothers me, will try to fix.

app/controllers/application_controller.rb
thejonroberts: good catch.

thejonroberts:
@elasticspoon Agree with the feedback. Display issue really gets me! 

Sounds like an important feature to get out. If you want, I can open a PR against this branch (or main if it has merged already) for easier review. Shouldn't take too long, 🤞🏻 !

spec/system/case_contacts/edit_spec.rb
thejonroberts: That is what `expect(page).to have_text casa_case.case_number` on last line of spec is checking. The form is using a multi-select js component that is a little tricky to grab the value of - I'm not seeing the database id in the DOM select option, so this is the simplest way. -- the case number is shown only visible when it is selected.

thejonroberts:
Updated with [these changes](https://github.com/thejonroberts/casa/pull/3)

Issue #5860, opened 2024-06-25
case contact creation date bug


elasticspoon:
Volunteers report having intermitent bugs with dates when creating case contacts. Sometimes the date they enters into a new contact is saved as the current date. Sometimes they will even edit the contact and the updated date will not save till a third retry. 

This is an intermittent failure that we have been unable to reproduce. The timing loosely aligns with this PR https://github.com/rubyforgood/casa/pull/5609 being merged so it may be a culprit.

thejonroberts:
I would think that using browser native date input would be more consistent, this is a tough one.

I looked at this a bit. Didn't find anything that would obviously lead to the bug --  but I have some thoughts.

1. CaseContact model uses a strange enum implementation for **string column** `status`. Rails may get confused by this (and devs certainly could be -- I was).  This enum setup is how the status `active?`, `started?`, `details?`, etc.. methods used by the wicked wizard form are setup, and don't seem to do anything else.  
     - We can get rid of the enum and define them explicitly, [as the gem suggests](https://github.com/zombocom/wicked/wiki/Building-Partial-Objects-Step-by-Step).
2. CaseContact `occurred_at` has three validations, that run only if status is `active_or_details?` (determined by above). 
    - We can change them to run anytime record is not `started`, because after first form details page, where date selection happens, contact should never go back to 'started' status. This may help validations run when expected, especially during edits -- I'm unsure what status should be during an edit step, do we kick things back from active to the edit form step? What if that edit is abandoned?

I have no idea if above will fix the issue. But it may, and could help give better validation feedback to users? 

We could also log attributes when the validation fails? Get some actual idea of what date/format is sent.

I have done some work already, but based it off of my work on #5957 , I will rebase the work if that PR doesn't go through soon.

thejonroberts:
Can ignore my previous comment.

One idea I have is to change the date input from the RangedDatePicker component to a straight rails `form.date_field`, and let standard browser & rails validations happen. RangedDatePicker does not set standard max_date/min_date for the input but sets them in data attributes, and handles the range validation via js (in app/javascript/src/validated_form.js) to do this:

![Image](https://github.com/user-attachments/assets/bbe91e2e-3953-4271-82ce-7d09fb32df27)

(note also 'minimize' does nothing on it currently (in safari at least), and will cover fields / submit button)

It would also be more consistent with validation for rest of the form. Is changing that an option?

The only other idea I have is to make a tagged log on every `occurred_at` change with relevant info.

[...]

PR #5974, opened 2024-08-07
Bug: CaseContact form date input

thejonroberts:
Addresses #5860

### What changed, and _why_?
Javascript is running (on any page load app wide) that manipulated `#case_contact_occurred_at` field value if it was present. This was probably to ensure it matched previous date picker library format, but app uses browser date pickers now. The manipulation may have been the cause of bug - we have not been able to reproduce the issue. Even if not, this brings date selection in line with standard validation.

- Remove offending javascript
- Swap date range picker component (partial) that worked with it, in favor of standard rails `form.date_field `, plus proper attributes for browser validation, using iso8601 date format standard for browsers
- Change occurred_at 'not in future' validation from custom method to comparison (as we validate minimum date)
- Use MINIMUM_DATE model class constant to ensure form and model use same date.
- Remove the components/dead javascript that aren't used after the change

Unsure how to test as not able to reproduce/browser specific. I think existing system new/edit specs do a lot of date inputting, so I think that's sufficient, but open to suggestions.

### Feelings gif (optional)
alt text


compwron (approved):

app/views/case_contacts/form/details.html.erb
```diff
@@ -53,10 +53,14 @@
...
-          <%= render "layouts/components/ranged_date_picker",
-            form: form, model_field: :occurred_at, initial_value: occurred_at.to_date,
-            max_date: "today", component_name: "occurred on" %>
+          <% min_date = CaseContact::MINIMUM_DATE %>
+          <% current_date = Date.current %>
+          <% initial_value = @case_contact.occurred_at&.to_date || current_date %>
+          <%= form.date_field :occurred_at,
+            value: initial_value.to_fs(:iso8601),
+            max: current_date.to_fs(:iso8601),
+            min: min_date.to_fs(:iso8601),
```
thejonroberts, line 62: Could drop these attributes to handle with model validations only (allow submit & show validation errors). This will prevent selecting bad dates via UI, but user can still type them in, afaik.

elasticspoon (approved):
I good with this as if, thanks. Love removing unnecessary dependencies. 

There is existing issue where the current inputs don't take into account the timezone of the user is not getting used. So if I enter 8/10/2024 that is saved at 8/10/24 UTC and if I later convert that to EST it becomes 8/9/24 EST. 

It would be cool if that could be covered in this PR but I'm totally fine with that being a different PR as well. (Also I'm not 100% if we use the `occurred_at` with a timezone)

thejonroberts:
> I good with this as if, thanks. Love removing unnecessary dependencies.
> 
> There is existing issue where the current inputs don't take into account the timezone of the user is not getting used. So if I enter 8/10/2024 that is saved at 8/10/24 UTC and if I later convert that to EST it becomes 8/9/24 EST.
> 
> It would be cool if that could be covered in this PR but I'm totally fine with that being a different PR as well. (Also I'm not 100% if we use the `occurred_at` with a timezone)

Yes, feels great when a PR is more red than green!

There is some code that grabs time zone from a browser cookie (`browser_time_zone` from Users::TimeZone concern), but I don't think it's being used here (or CaseContact show). It's set by `app/javascript/src/time_zone.js`.
Definitely worth thinking about and testing a little better, but I think it would take some time to dig around and be really consistent.

Time is always tricky, I like to refer to [this thoughtbot article](https://thoughtbot.com/blog/its-about-time-zones) a lot... Might be worth adding a user timezone column for the user to be able to choose? Or having a good fallback.

elasticspoon:
I think [this PR](https://github.com/rubyforgood/casa/pull/5828) is kind of the idea of how we would do it in this code base but I guesstimate there are more places incorrectly using timezones than correctly at the moment.

thejonroberts:
The occurred_at param is a plain date, sent without any tz info, so saving to UTC, which is good. Potential issues if user is on the 'other side' of the date line from UTC. 
- User Behind => form prefilled date is a day later than expected (tomorrow from user perspective). 
- User Ahead => validation fails because they are 'in the future' (tomorrow from server perspective).

America/New York is 4 hours behind UTC, so the first scenario could be happening after 8pm Eastern for users. It may be annoying but not blocking creation, and doesn't sound like this bug reported.

I think incorporating `browser_time_zone` would be beneficial, but perhaps separate from this work?

elasticspoon:
> The occurred_at param is a plain date, sent without any tz info, so saving to UTC, which is good. Potential issues if user is on the 'other side' of the date line from UTC.
> 
>     * User Behind => form prefilled date is a day later than expected (tomorrow from user perspective).
> 
>     * User Ahead => validation fails because they are 'in the future' (tomorrow from server perspective).
> 
> 
> America/New York is 4 hours behind UTC, so the first scenario could be happening after 8pm Eastern for users. It may be annoying but not blocking creation, and doesn't sound like this bug as reported.
> 
> I think incorporating `browser_time_zone` would be beneficial, but perhaps separate from this work?

Yea for sure not part of this. I was just mentioning it in general. And I has been reported with regards to some notification and email tables in the past.

PR #5975, opened 2024-08-08
CaseContact status enum changes

thejonroberts:

No issue: syntax changes only. None of this is necessary.

While working on CaseContacts, I noticed the status enum was defined oddly. I started poking around. The enum is used to define `.active?`/`.started?` and not much else - doesn't use any of the other methods or scopes that come along with an enum definition. It is also a string column, which is atypical. That is for easy compatibility with the wicked wizard form steps - but standard integer column would make for faster queries.

### What changed, and _why_?

The main thing I wanted to do is define the enum in a way that makes sense. But then I kept going to see if we could better use the enum, and set things up for a transition to integer column. That won't be necessary until performance is a real-world concern. It would require migrating existing data, which could be delicate. But these changes would enable it. The definition & wizard form updates would work if the enum column was changed to an integer type.

- Explicitly define the enum (_If this is all you want to do here, I'm happy with that_)
- Use the methods and scopes that are defined by doing so
- Integrate those changes into the wizard step updates
- Integrate the enum into factories

Current specs. ~~Could add/modify specs for each of the enum-created scopes, will probably look at that a bit, especially if people want to do this.~~ I think the scope changes in specs are enough?

### DIFF
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -47,7 +47,12 @@ class CaseContact < ApplicationRecord
   # Corresponds to the steps in the controller, so validations for certain columns can happen at those steps.
   # These steps must be listed in order and have an html template in case_contacts/form.
   FORM_STEPS = %i[details notes expenses].freeze
-  enum status: (%w[started active] + FORM_STEPS.map(&:to_s)).zip((%w[started active] + FORM_STEPS.map(&:to_s))).to_h
+  enum :status,
+    started: "started",
+    active: "active",
+    details: "details",
+    notes: "notes",
+    expenses: "expenses"
 
   def active_or_details?
     status == "details" || active?
--- a/app/models/case_contact.rb
+++ b/app/models/case_contact.rb
@@ -162,7 +162,7 @@ class CaseContact < ApplicationRecord
     where(casa_case_id: case_ids) if case_ids.present?
   }
 
-  scope :no_drafts, ->(checked) { (checked == 1) ? where(status: "active") : all }
+  scope :no_drafts, ->(checked) { (checked == 1) ? active : all }
 
   scope :with_metadata_pair, ->(key, value) { where("metadata -> ? @> ?::jsonb", key.to_s, value.to_s) }
   scope :used_create_another, -> { with_metadata_pair(:create_another, true) }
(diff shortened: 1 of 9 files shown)


app/models/case_contact.rb
```diff
@@ -40,16 +40,28 @@ class CaseContact < ApplicationRecord
   after_save_commit ::CaseContactMetadataCallback.new
 
   # Corresponds to the steps in the controller, so validations for certain columns can happen at those steps.
-  # These steps must be listed in order and have an html template in case_contacts/form.
+  # These steps must be listed in order, have an html template in case_contacts/form, & be listed in the status enum
   FORM_STEPS = %i[details notes expenses].freeze
-  enum status: (%w[started active] + FORM_STEPS.map(&:to_s)).zip((%w[started active] + FORM_STEPS.map(&:to_s))).to_h
```
thejonroberts, line 45: This was written this way to allow adding wizard form steps and have them automatically incorporated into the enum. I don't think that is worth it being so hard to read. I would rather define explicitly. It evaluates to:
```ruby
status: {
  "started"=>"started", 
  "active"=>"active", 
  "details"=>"details", 
  "notes"=>"notes", 
  "expenses"=>"expenses"
}
```

spec/factories/case_contacts.rb
```diff
@@ -1,16 +1,18 @@
 FactoryBot.define do
+  # note: uses enum status column, see:
+  # https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#enum-traits
```
elasticspoon, line 3: I found this a bit hard to follow initially. It made sense that if there is an enum then traits are automagically created for it but I did not realize that `active`, `started` within the factories were calls to tho traits. 

I would appreciate like 2 sentences explaining that because I don't think is a pattern used elsewhere in the project.

This was the piece of context I was missing https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#traits-within-traits.

elasticspoon (approved):
LGTM.

spec/factories/case_contacts.rb
thejonroberts: That makes sense, will explain better!

thejonroberts: updated.

PR #5999, opened 2024-08-19
CaseGroups request spec & custom policy generator

thejonroberts:
Addresses #5563

### What changed, and _why_?
- Created request specs for CaseGroups controller (they were allowed-skipped)
- CaseGroupController did not have a policy to use (but everything was built from current_organization)
- Wanted to add a policy generator since the `pundit:policy` generator is so minimal
- Created a CaseGroup policy (and spec), and refined generator based on that
- Incorporate a policy into the controller, as used in other routes
- Created hooks for rails controller and scaffold generators to ensure a policy is generated along with them

The request spec was created with the controller as-is, and made to pass.

Then the policy changes were made in the controller without changing specs -- with one exception: The update action redirected to new rather than traditional edit redirect. I made that change.

NOTES:
- [This commit](https://github.com/rubyforgood/casa/pull/5999/commits/43cff782e6559bbbc4f02d1fde03d3224fb5b6c6) contains the changes from generated policy to make policy spec pass.
- I also made some custom scaffold and controller request spec templates, but that may be too much to include here, I will open a separate PR for those.

### Screenshots please :)
Policy generator help:
Screenshot 2024-08-19 at 11 24 30 PM

### Feelings gif (optional)
alt text

### DIFF
--- a/app/controllers/case_groups_controller.rb
+++ b/app/controllers/case_groups_controller.rb
@@ -1,21 +1,25 @@
 class CaseGroupsController < ApplicationController
   before_action :require_organization!
-  before_action :authorize_admin_or_supervisor!
+  before_action :set_case_group, only: %i[edit update destroy]
 
   def index
-    @case_groups = current_organization.case_groups.includes(:casa_cases)
+    authorize CaseGroup
+    @case_groups = policy_scope(CaseGroup).includes(:casa_cases)
   end
 
   def new
-    @case_group = CaseGroup.new
+    @case_group = CaseGroup.new(casa_org: current_organization)
+    authorize @case_group
   end
 
   def edit
-    @case_group = current_organization.case_groups.find(params[:id])
+    authorize @case_group
   end
 
   def create
     @case_group = current_organization.case_groups.build(case_group_params)
+    authorize @case_group
+
     if @case_group.save
       redirect_to case_groups_path, notice: "Case group created!"
     else
@@ -24,27 +28,30 @@ class CaseGroupsController < ApplicationController
   end
 
   def update
-    @case_group = current_organization.case_groups.find(params[:id])
+    authorize @case_group
+
     if @case_group.update(case_group_params)
       redirect_to case_groups_path, notice: "Case group updated!"
     else
-      render :new, status: :unprocessable_entity
+      render :edit, status: :unprocessable_entity
     end
   end
 
   def destroy
-    case_group = current_organization.case_groups.find(params[:id])
-    case_group.destroy
+    authorize @case_group
+
+    @case_group.destroy
     redirect_to case_groups_path, notice: "Case group deleted!"
   end
 
   private
 
   def case_group_params
+    params.merge(casa_org: current_organization)
     params.require(:case_group).permit(:name, casa_case_ids: [])
   end
 
-  def authorize_admin_or_supervisor!
-    authorize :application, :admin_or_supervisor?
+  def set_case_group
+    @case_group = CaseGroup.find(params[:id])
   end
 end
--- a/app/controllers/case_groups_controller.rb
+++ b/app/controllers/case_groups_controller.rb
@@ -52,6 +52,6 @@ class CaseGroupsController < ApplicationController
   end
 
   def set_case_group
-    @case_group = CaseGroup.find(params[:id])
+    @case_group = policy_scope(CaseGroup).find(params[:id])
   end
 end
--- a/app/models/all_casa_admin.rb
+++ b/app/models/all_casa_admin.rb
@@ -12,6 +12,10 @@ class AllCasaAdmin < ApplicationRecord
   def supervisor?
     false
   end
+
+  def volunteer?
+    false
+  end
 end
 
 # == Schema Information
--- a/app/models/all_casa_admin.rb
+++ b/app/models/all_casa_admin.rb
@@ -4,6 +4,14 @@ class AllCasaAdmin < ApplicationRecord
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
   devise :database_authenticatable, :invitable, :recoverable, :validatable, :timeoutable, invite_for: 1.weeks
+
+  def casa_admin?
+    false
+  end
+
+  def supervisor?
+    false
+  end
 end
 
 # == Schema Information
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -57,8 +57,6 @@ class ApplicationPolicy
       user.casa_org == record
     when CasaAdmin, CasaCase, Volunteer, Supervisor, HearingType, ContactTypeGroup, ContactTopic
       user.casa_org == record.casa_org
-    when CaseGroup
-      user.casa_org == record.casa_org
     when CourtDate, CaseContact, CaseAssignment
       user.casa_org == record&.casa_case&.casa_org
     when LearningHour
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -48,7 +48,7 @@ class ApplicationPolicy
   end
 
   def is_admin?
-    user.casa_admin?
+    user&.casa_admin?
   end
 
   def same_org?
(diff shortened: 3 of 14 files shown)


app/models/all_casa_admin.rb
```diff
@@ -4,6 +4,14 @@ class AllCasaAdmin < ApplicationRecord
...
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
   devise :database_authenticatable, :invitable, :recoverable, :validatable, :timeoutable, invite_for: 1.weeks
+
+  def casa_admin?
+    false
+  end
+
+  def supervisor?
+    false
+  end
```
thejonroberts, line 14: I don't understand how an AllCasaAdmin is allowed to interact with the app exactly... I don't think Policies take that into consideration... But I tested them as a user in the Policy spec.

My intuition is that they should be superusers/have access to everything, but they seem to be kept separate, so I decided to do this rather than make a lot of changes in ApplicationPolicy.

app/policies/case_group_policy.rb
```diff
@@ -0,0 +1,42 @@
+class CaseGroupPolicy < ApplicationPolicy
+  class Scope < ApplicationPolicy::Scope
+    def resolve
+      case user
+      when CasaAdmin, Supervisor
+        scope.where(casa_org: @user.casa_org)
+      when Volunteer
+        scope.none
+      else
+        scope.none
```
thejonroberts, line 10: Some policies raise "user type unrecognized" or similar here. I think it makes more sense just to not return the records.

I almost added AllCasaAdmin to the cases, but currently they fall here.

spec/policies/case_group_policy_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        it { is_expected.to include(casa_org_case_group) }
+        it { is_expected.not_to include(other_casa_org_case_group) }
+      end
+
+      context "when user is an all_casa_admin" do
+        let(:user) { all_casa_admin }
+
+        it { is_expected.not_to include(casa_org_case_group) }
+        it { is_expected.not_to include(other_casa_org_case_group) }
+      end
```
thejonroberts, line 117: Is this the correct behavior for AllCasaAdmin?

lib/generators/rails/policy/templates/policy_spec.rb.tt
thejonroberts, line 1: These Thor template files are used by the generator to create the files. Not as important to review as the end results, but take a look!

app/controllers/case_groups_controller.rb
```diff
@@ -24,27 +28,30 @@ def create
...
 
   def case_group_params
+    params.merge(casa_org: current_organization)
     params.require(:case_group).permit(:name, casa_case_ids: [])
   end
 
-  def authorize_admin_or_supervisor!
-    authorize :application, :admin_or_supervisor?
+  def set_case_group
+    @case_group = CaseGroup.find(params[:id])
```
thejonroberts, line 55: Some controllers use `policy_scope(CaseGroup).find(params[:id])` here. That makes sense, but raises an ActiveRecord not found exception if an unauthorized user... We handle authorization failures in ApplicationController with a redirect to root_path (`rescue_from Pundit::NotAuthorizedError, with: :not_authorized`). I think I prefer to let that happen.

config/initializers/generators.rb
```diff
@@ -0,0 +1,38 @@
...
+end
+
+module ActiveModel
+  class Railtie < Rails::Railtie
+    generators do |app|
+      Rails::Generators.configure! app.config.generators
+      Rails::Generators::ControllerGenerator.include PolicyGenerator::ControllerGenerator
+      Rails::Generators::ScaffoldControllerGenerator.include PolicyGenerator::ScaffoldControllerGenerator
+    end
+  end
```
thejonroberts, line 37: Feels a little tacky, but need `ActiveModel ` to have ScaffoldControllerGenerator defined. See [this discussion](https://gist.github.com/annikoff/331f785aa7a207a7945b1eca6eff526b#a-hook-to-invoke-the-custom-generator-with-scaffolding-or-with-controllers-generators).

app/policies/application_policy.rb
```diff
@@ -76,16 +76,16 @@ def same_org?
...
 
   def is_supervisor?
-    user.supervisor?
+    user&.supervisor?
   end
 
   def is_supervisor_same_org?
     # eventually everything should use this
-    user.supervisor? && same_org?
+    is_supervisor? && same_org?
```
thejonroberts, line 88: Could move these changes out to CaseGroupPolicy if don't want to tolerate nil user in AppPolicy (currently give NoMethodErrors)

app/controllers/case_groups_controller.rb
thejonroberts: Thinking about this some more and I think probably better to use policy_scope and let the unauthorized user think that the record doesn't exist - don't leak information.

app/models/all_casa_admin.rb
elasticspoon: All casa admins are closer to admin accounts than super users. They are mainly used to either create new casa orgs and create new admin accounts. They can also do feature flags and make patch notes.

spec/factories/case_groups.rb
```diff
@@ -2,12 +2,17 @@
...
     casa_org { CasaOrg.first || create(:casa_org) }
     sequence(:name) { |n| "Family #{n}" }
 
     after(:build) do |case_group, evaluator|
-      if case_group.case_group_memberships.empty?
+      if evaluator.casa_cases.size > 0
+        evaluator.casa_cases.each do |casa_case|
+          case_group.case_group_memberships.build(casa_case: casa_case)
+        end
+      elsif case_group.case_group_memberships.empty?
```
elasticspoon, line 15: This feels excessively complex but I can't think of a better approach at this moment.

elasticspoon (approved):
I played around with this a bit. This is sick. 

Not 100% sure we should prefill all routes off the bat (index, create, etc) as opposed to maybe just index and create to demo how it might look with different perms but I think that could just be ironed out in practice. 

This is hella neat. Thank you!

elasticspoon:
Ahhhhhh. never mind I am silly I can just do `rails g policy some_action`

spec/policies/case_group_policy_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  describe ".scope" do
+    subject { described_class::Scope.new user, CaseGroup.all }
+
+    describe "#resolve" do
+      let!(:casa_org_case_group) { create :case_group, casa_org: }
+      let!(:other_casa_org_case_group) { create :case_group, casa_org: create(:casa_org) }
+
+      subject { super().resolve }
```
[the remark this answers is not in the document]
thejonroberts: `super` in this case is the `subject` in the parent block, on L74, this call its `resolve` method... 

I think it's cool but all flash and no usefulness here - we don't use the "super subject" for any examples.  I will just replace super here with the L74 block.

app/policies/case_group_policy.rb
thejonroberts: Yeah, come to think of it, I don't think I've seen it used outside of a PolicyScope. But it's nice to have an empty AR collection instead of [] here.

Thanks for the AllCasaAdmin explanation, that helps. So they should definitely get `scope.none`'d by default.

lib/generators/rails/policy/templates/policy.rb.tt
```diff
@@ -0,0 +1,52 @@
...
+  # No actions specified, Example usage:
+  # def index?
+  #   <%= user_example %>
+  # end
+  #
+  # def show?
+  #   <%= record_example %>
+  # end
+<%- end -%>
+<%- actions.each do |action| -%>
```
thejonroberts, line 39: @elasticspoon `actions` here is the full CRUD set for a scaffold generator, but if you use the `rails g controller Model index show` it would be ['index', 'show'], so only creates those policies.  if no actions it will make the comment above.

spec/factories/case_groups.rb
thejonroberts: hmm... nice eye. I think I could set casa_cases, count with ORs and scrap the if/else maybe, will try it out.

thejonroberts: a little better, but didn't want to break any other specs.

lib/generators/rails/policy/templates/policy.rb.tt
thejonroberts: oh, I changed the defaults, so that comment isn't happening for `rails g policy Thing `. I did that so it would work for a scaffold. hmm... gonna see if I can get that back...

thejonroberts: updated. 


When no actions, this comment is generated, plus example specs for index & show, with comment to modify/remove.

PR #6004, opened 2024-08-22
Refactor ApplicationPolicy #same_org?

thejonroberts:
No issue. `ApplicationPolicy#same_org?` was just a little too wild, and allowed [misuse of `authorize`](https://github.com/rubyforgood/casa/pull/6004#discussion_r1727838371).

### What changed, and _why_?
`ApplicationPolicy#same_org?` has gotten out of hand... It should not worry about the class of the record it's getting, only what the record's associated CasaOrg is. 

Any variation from that can be left to the sub-policies that deal with specific models. For example, I redefined `same_org?` for `CasaOrgPolicy` because the record is the casa_org itself, so record.casa_org is impossible. Between that and  most of the Models involved not having a casa_org association defined, I can see how it got to where it is slowly over time. Changes in this PR:

- Use model associations for same_org? check.  
- Override in Policy classes as needed (CasaOrg). 
- Change index actions that went through same_org? (ContactTopicPolicy) 
- Change member actions which authorized a class instead of the available `@record`. (ContactTopicController)
- Fix broken specs (only 2)
- Make an instructive comment in same_org? to discourage future abuse.

Current tests only, as no behavior changes. If there is something that may be under tested, I can dig in. ~~I was seeing some failing tests locally but they happened again when I removed the changes. Curious to see what happens on CI.~~

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -52,26 +52,10 @@ class ApplicationPolicy
   end
 
   def same_org?
-    case record
-    when CasaOrg
-      user.casa_org == record
-    when CasaAdmin, CasaCase, Volunteer, Supervisor, HearingType, ContactTypeGroup, ContactTopic
-      user.casa_org == record.casa_org
-    when CourtDate, CaseContact, CaseAssignment
-      user.casa_org == record&.casa_case&.casa_org
-    when LearningHour
-      user.casa_org == record&.user&.casa_org
-    when ChecklistItem
-      user.casa_org == record&.hearing_type&.casa_org
-    when ContactType
-      user.casa_org == record&.contact_type_group&.casa_org
-    when Followup
-      user.casa_org == record&.case_contact&.casa_case&.casa_org
-    when Class # Authorizing against collection, does not belong to org
-      true
-    else # Type not recognized, no auth since we can't verify the record
-      false
-    end
+    # NOTE: must have casa_org association on a Policy's associated Model
+    # that is: `has_one :casa_org, through: :some_association` (may need to define :some_association)
+    # do not use for collection actions (index), check user type & use policy_scope() on the collection
+    user&.casa_org.present? && user.casa_org == record&.casa_org
   end
 
   def is_admin_same_org?
--- a/app/controllers/contact_topics_controller.rb
+++ b/app/controllers/contact_topics_controller.rb
@@ -4,9 +4,8 @@ class ContactTopicsController < ApplicationController
 
   # GET /contact_topics/new
   def new
-    authorize ContactTopic
-    contact_topic = ContactTopic.new(casa_org_id: current_user.casa_org_id)
-    @contact_topic = contact_topic
+    @contact_topic = ContactTopic.new(casa_org_id: current_user.casa_org_id)
+    authorize @contact_topic
   end
 
   # GET /contact_topics/1/edit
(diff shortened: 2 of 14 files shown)


app/policies/application_policy.rb
```diff
@@ -52,26 +52,10 @@ def is_admin?
...
-    when LearningHour
-      user.casa_org == record&.user&.casa_org
-    when ChecklistItem
-      user.casa_org == record&.hearing_type&.casa_org
-    when ContactType
-      user.casa_org == record&.contact_type_group&.casa_org
-    when Followup
-      user.casa_org == record&.case_contact&.casa_case&.casa_org
-    when Class # Authorizing against collection, does not belong to org
-      true
```
thejonroberts, line 71: This allowed thinking that a collection was authorized, when really it should have used a Policy::Scope instead of ending up here!

thejonroberts: I should probably add this case to app policy spec.

elasticspoon (approved):
Makes sense. Honestly, part of me wonders if we should even have the same org in the application policy? Maybe each individual policy should just be responsible for implementing the method. 

Especially with the way you had to do the testing `allow(record).to receive(:casa_org).and_return(casa_org)`

elasticspoon:
@thejonroberts oh but this does need a small lint fix

thejonroberts:
> Makes sense. Honestly, part of me wonders if we should even have the same org in the application policy? Maybe each individual policy should just be responsible for implementing the method.
> 
> Especially with the way you had to do the testing `allow(record).to receive(:casa_org).and_return(casa_org)`

Well, I obviously think otherwise. 😉 It's just going to look the same for all of the classes (14 and counting). Unless they don't have a proper association defined - and if the model doesn't have a relation to an org, what is it doing here in `same_org?`?

I could have used a factory to create a record with an association, like `create :contact_type, casa_org: casa_org/other_casa_org`, but the method doesn't care what type of record it is, only if it returns the correct casa_org. And which model to use? At one point, I thought about making an array picking one at random!  So I just used a double and stubbed the casa_org value it returns.

One thing I didn't test is when there is no casa_org association, I could add that case (would raise a NoMethodError), but seems like that should only matter during development?

elasticspoon:
> > Makes sense. Honestly, part of me wonders if we should even have the same org in the application policy? Maybe each individual policy should just be responsible for implementing the method.
> > Especially with the way you had to do the testing `allow(record).to receive(:casa_org).and_return(casa_org)`
> 
> Well, I obviously think otherwise. 😉 It's just going to look the same for all of the classes (14 and counting). Unless they don't have a proper association defined. But if the model doesn't have a relation to an org, what is it doing here in `same_org?`?
> 
> I could have used a factory to create a record with an org association, like `create :contact_type, casa_org: casa_org/other_casa_org`, but which model to use? At one point, I thought about making an array of factories models and picking one at random! That is a good indication that it really doesn't matter!
> 
> So `same_org?` doesn't care what type of record it is, only if it returns the correct casa_org -- so we just make a `record` [double and stub](https://thoughtbot.com/blog/stubbles-stubbing-and-doubles-innit) the `.casa_org` value it returns.
> 
> One thing I didn't test is when there is no casa_org association. I could add that case (would raise a NoMethodError), but seems like that should only come up during development, and would lead the dev to the comment?

Yea I see your point, I'm not totally sure how else you could approach testing it. It makes sense, it does a much better job expressing that you are basically duck typing that method.

thejonroberts:
Added user.casa_org = nil & user = nil contexts to specs. 

Also, after thinking about the discussion above, added some clarity to the record double setup, differentiating between `org_record` and `other_org_record`. Should make it easier to understand for future devs. 🤞🏻

PR #6006, opened 2024-08-24
fix: flaky volunteers & supervisors system specs

thejonroberts:
No issue -- I have seen this spec cause failures locally and in CI. 

### What changed, and _why_?
`spec/system/volunteers/index_spec.rb` `:55` & `:400` were the most common failures, roughly 90% of my local runs I would guess.  I was running specs in 'defined' order, not random, and I think it was happening more consistently that way.

I ran some rubocop-capybara auto fixes while troubleshooting - figured I may as well leave those changes in for good measure.

It seems like the `travel_to` calls were the major culprit. Adding `travel_back`s helped, but I do not see why they are necessary, so I just removed them.  The other major issue was the sign_out -> visit -> sign_in -> visit sequence, the spec would get kicked out to the sign in page. I changed the sequence to a `refresh` call. I think a page refresh still properly tests the feature?

### DIFF
--- a/spec/system/supervisors/index_spec.rb
+++ b/spec/system/supervisors/index_spec.rb
@@ -3,12 +3,12 @@ require "rails_helper"
 RSpec.describe "supervisors/index", type: :system do
   shared_examples_for "functioning sort buttons" do
     it "sorts table columns" do
-      expect(page).to have_selector("tr:nth-child(1)", text: expected_first_ordered_value)
+      expect(page).to have_css("tr:nth-child(1)", text: expected_first_ordered_value)
 
       find("th", text: column_to_sort).click
 
-      expect(page).to have_selector("th.sorting_asc", text: column_to_sort)
-      expect(page).to have_selector("tr:nth-child(1)", text: expected_last_ordered_value)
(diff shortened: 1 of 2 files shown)


elasticspoon (approved):
Looks reasonable. Should we just add rubocop-capybara to the app? I personally often mess up the `to_not have_x` vs `to have_no_x`

thejonroberts:
> Looks reasonable. Should we just add rubocop-capybara to the app? I personally often mess up the `to_not have_x` vs `to have_no_x`

Yes, I think that would be great to add! I always always have to double check things like that. So I have it as a global gem for my IDE. 

I'm not sure how to hook that up with standardrb, but I I could poke around this week. I do know you can use [standard via rubocop](https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard), so that is an option.

I also think standard-rails/rubocop-rails would be great to add. I actually implemented it earlier this week, but I didn't want to overwhelm you 😆 . Here's a [PR for standard-rails](https://github.com/thejonroberts/casa/pull/4) if you're curious.

PR #6047, opened 2024-09-20
Add extra day to case contact date validation

thejonroberts:
Addresses #5860 

### What changed, and _why_?
Addresses (temporarily?) issue where validation failed when user selected current day in browser. 

Date could now be wrong (tomorrow) for those situations, however. Also changes Date.tomorrow to Time.zone.tomorrow to possibly to mitigate that.

Current tests only.  Cannot find a decent solution for setting capybara browser driver timezones to tests the difference between browser and server. Possibly could add request spec sending date as browser would send... but unsure how to account for TimeZone differences there?


elasticspoon (approved):
LGTM

PR #6054, opened 2024-09-23
Flaky specs: Use travel_back when travel_to is used

thejonroberts:
None. Fixes flaky specs.

### What changed, and _why_?
Added travel_back wherever travel_to is used.

I was seeing a lot of spec failures locally when making a change that were seemingly unrelated to my changes. This fixed them.  A lot of places where travel to was used without a travel back. (Probably an assumption that this was done automatically). Could possibly do `travel_back` in an after(:each) block in spec_helper.rb instead. Would mean doing for every test, not just time-traveling specs, and I errored on the side of being explicit. The benefit of the spec helper approach would be that it won't ever be a problem again...

### DIFF
--- a/spec/helpers/notifications_helper_spec.rb
+++ b/spec/helpers/notifications_helper_spec.rb
@@ -20,6 +20,7 @@ RSpec.describe NotificationsHelper do
       notification_created_before_deploy_a.update_attribute(:created_at, 2.days.ago - 1.hour)
       notification_created_before_deploy_b.update_attribute(:created_at, 3.days.ago)
     end
+    after { travel_back }
 
     describe "#notifications_after_and_including_deploy" do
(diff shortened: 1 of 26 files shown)


spec/requests/health_spec.rb
elasticspoon, line 1: This spec already has `travel_back in after blocks. I don't *think* we need additional ones?

elasticspoon (approved):
1 small comment but overall LGTM

spec/requests/health_spec.rb
thejonroberts: good catch, I removed them.

elasticspoon:
@thejonroberts could you fix the lint on this?

thejonroberts:
updated @elasticspoon

PR #6057, opened 2024-09-24
Update stimulus rails nested-form & switch to namespaced version

thejonroberts:
Preparation for case contact form overhaul

### What changed, and _why_?
Update `stimulus-rails-nested-form` and switch to the namespaced package `@stimulus-components/rails-nested-form `. The package was using the old form of the npm name. That seems to be the reason for the major version change, no breaking changes in the [changelog here](https://github.com/stimulus-components/stimulus-rails-nested-form/blob/master/CHANGELOG.md).

[package info here](https://www.stimulus-components.com/docs/stimulus-rails-nested-form).

No new functionality, existing tests should do.


app/javascript/controllers/index.js
elasticspoon, line 1: `import RailsNestedForm from '@stimulus-components/rails-nested-form'` should this be in alpha order along with the other controllers?

elasticspoon (approved):
LGTM, 1 minor comment

app/javascript/controllers/index.js
thejonroberts: that makes sense, i was trying to differentiate that it is imported from a library

PR #6060, opened 2024-09-25
AdditionalExpenses and ContactTopicAnswer controllers & policies

thejonroberts:
Preparation for Case Contact Form re-design - #6048 

Creates policy and controllers for AdditionalExpenses and ContactTopicAnswers. These resources will be used by frontend to persist/delete nested form items as they are added & removed - which is why only create/destroy actions exist in the controllers.  This is necessary so that autosave of the form during topic answer (note) input does not re-create a new note (or try to delete a previously deleted note) every time it runs. 

Previously, a blank ContactTopicAnswer would be created in the first details step per selected topic... so each answer during notes step would already have a persisted CTA to update during autosave.  Expenses were not affected since autosave only happened during notes step.

Separated this stuff out to make for easier review.

### DIFF
--- a/app/controllers/additional_expenses_controller.rb
+++ b/app/controllers/additional_expenses_controller.rb
@@ -1,0 +1,32 @@
+class AdditionalExpensesController < ApplicationController
+  def create
+    @additional_expense = AdditionalExpense.new(additional_expense_params)
+    authorize @additional_expense
+
+    respond_to do |format|
+      if @additional_expense.save
+        format.json { render json: @additional_expense.as_json, status: :created }
+      else
+        format.json { render json: @additional_expense.errors.as_json, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  def destroy
+    @additional_expense = AdditionalExpense.find(params[:id])
+    authorize @additional_expense
+
+    @additional_expense.destroy!
+
+    respond_to do |format|
+      format.json { head :no_content }
+    end
+  end
+
+  private
+
+  def additional_expense_params
+    params.require(:additional_expense)
+      .permit(:case_contact_id, :other_expense_amount, :other_expenses_describe)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -96,6 +96,9 @@ Rails.application.routes.draw do
     end
   end
 
+  resources :contact_topic_answers, only: %i[create destroy],
+    constraints: lambda { |req| req.format == :json }
+
   resources :reports, only: %i[index]
   get :export_emails, to: "reports#export_emails"
 
@@ -190,6 +193,9 @@ Rails.application.routes.draw do
     end
   end
   resources :case_court_orders, only: %i[destroy]
+
+  resources :additional_expenses, only: %i[create destroy],
+    constraints: lambda { |req| req.format == :json }
 
   namespace :all_casa_admins do
     resources :casa_orgs, only: [:new, :create, :show] do
--- a/app/controllers/contact_topic_answers_controller.rb
+++ b/app/controllers/contact_topic_answers_controller.rb
@@ -1,0 +1,32 @@
+class ContactTopicAnswersController < ApplicationController
+  def create
+    @contact_topic_answer = ContactTopicAnswer.new(contact_topic_answer_params)
+    authorize @contact_topic_answer
+
+    respond_to do |format|
+      if @contact_topic_answer.save
+        format.json { render json: @contact_topic_answer.as_json, status: :created }
+      else
+        format.json { render json: @contact_topic_answer.errors.as_json, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  def destroy
+    @contact_topic_answer = ContactTopicAnswer.find(params[:id])
+    authorize @contact_topic_answer
+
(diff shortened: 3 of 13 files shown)


config/routes.rb
```diff
@@ -191,6 +194,9 @@
   end
   resources :case_court_orders, only: %i[destroy]
 
+  resources :additional_expenses, only: %i[create destroy],
+    constraints: lambda { |req| req.format == :json }
+
```
thejonroberts, line 199: I did these as root-level resources. They could be case_contact member resources (nested), as they can't exist without a contact parent. I tend to error on the side of not nesting resources... and it was easier. Can move if need be.

The constraints for :json was new to me. Made sense for the no-view / api style endpoint.

elasticspoon:
Hey! @thejonroberts 

Trying to review this PR, mainly trying to get my head about the autosave json stuff. Could you explain this part a bit more (from the big PR):

> Maintaining the autosave feature was complicated! Before, notes were the only thing that could change during an autosave, but now everything is in the form that gets sent during the autosave update. Therefore, nested_attributes_for contact_topic_answers and additional_expenses required a stimulus NestedForm controller that could create and destroy records -- otherwise, autosave would make new records for every save, or try to delete records that had already been removed in the prior update. This also required creating controllers & routes for those json requests.

If I understand you correctly you are saying on every form update(?) we create new contact topic answers and new additional expenses.

Doesn't the existing autosave just submit the form in the background with a debounce? It works with both notes and contact topic answers. What is breaking that behavior here?

Did you explore any other ways of retaining the autosave without the additional controllers? What were the issues?

thejonroberts:
@elasticspoon the relevant bits of the old flow were:

1. Select applicable ContactTopics during details step
2. Enter Answers for those 'selected' in 1 (& autosave on text value inputs)
3. Enter Expenses

**Step 2 is the only step with autosave in effect**. Since topics were already selected in step 1, you aren't adding new (non-persisted) ContactTopicAnswers, only changing the value of existing answers. (I don't think removing was an option, you could just sort of ignore the topics you didn't want to elaborate on). Therefore autosave can submit & re-submit the entire form, everything is already in the database, and it's update only.

Expenses/Step 3 doesn't autosave, so you can add/remove as much as you want without hitting the database until submission.

In the new, all-in-one form, you run into issues autosaving the form:
1. Adding an entry (topic answer or expense)    
    -  Add Note -> enter value -> autosave creates an Answer record... 
    -  Enter more info in same field -> **autosave creates another record** because form doesn't have an ID with the field, it was added after page load.
    - Submit or reload the page, you will have multiple records, one for each time autosave happened.
1. Deleting an entry
    - When editing an existing contact that has answers/expenses with IDs
    - Delete an item - form hides in display but marks hidden `_destroy` input true
    - Autosave should work fine once
    - Autosave again will fail `RecordNotFound` because you've deleted the record already, but we tried to destroy it again.

The delete could probably be worked around / rescued somehow in the controller.  But the Add problem requires the form to know IDs before it tries to autosave any databases.

The only other approach I thought of was something like:
```ruby
# in update action method:
@case_contact.contact_topic_answers.destroy_all
@case_contact.additional_expenses.destroy_all
@case_contact.update(case_contact_params)
```
Deleting all the records just seems wrong... especially before an update... seemed too risky to me, if update fails, what then?

So I figured make the Add and Delete buttons a database-persisted action via these controllers, plus javascript (otherPR) to update the form with new IDs or removing deleted entries entirely.

thejonroberts:
@elasticspoon it seems like a good use of a turbostream - just replace the entire form on autosave....  but I worried about the response replacing an input as the user is interacting with it! 
like user input, pauses -> (autosave request), user resumes typing -> that is replaced with autosave response (first input)

app/controllers/additional_expenses_controller.rb
```diff
@@ -0,0 +1,32 @@
+class AdditionalExpensesController < ApplicationController
+  def create
+    @additional_expense = AdditionalExpense.new(additional_expense_params)
+    authorize @additional_expense
+
+    respond_to do |format|
+      if @additional_expense.save
+        format.json { render json: @additional_expense.as_json, status: :created }
+      else
+        format.json { render json: @additional_expense.errors.as_json, status: :unprocessable_entity }
```
thejonroberts, line 10: The `.as_json` shouldn't be necessary here.  `render json: ...` calls `.to_json` behind the curtain, and at some point in the full test suite, `.to_json` stops behaving as expected. I was getting results like `"<AdditionalExpense #OBJECTID ... >"`, a string representation of the object reference rather than the object converted to json.  Also a .to_json call in application.html.erb would create failures. 

It was really weird. I ran rspec bisect for an entire day, but couldn't pinpoint the issue. Maybe OJ or some gem?

Anyway, I don't think it's important, this works fine. Just wanted to document the strangeness.

elasticspoon: its from a rails 7.2 update https://github.com/rails/rails/pull/51640 that PR changed where ActiveSupport got pulled in so it messed up calls to `to_json`. 

I spent way too long debugging it.

thejonroberts:
Closing as #6048 has gotten some feedback and is almost ready for review.

PR #6067, opened 2024-10-03
Modify store_deploy_time.rake 

MclPio:
to not update deploy time if any non recurring afterparty tasks failed

no issue, but discussed with @FireLemons on 2024-10-01

Problem: Sometimes the deploy notes update when the site does not. In the past this is caused by an afterparty task updating the deploy timestamp without considering whether the other afterparty tasks have successfully been run. If an afterparty task fails, heroku (our hosting service) will not allow any new features to be introduced to the site.

### What changed, and _why_?
moved the afterparty task that stores the deploy time to the end of the list alphabetically so it runs last every time run rake after_party:status within the afterparty task to gather information about how the other tasks ran and use this information to determine whether the deploy time should be updated. Added an if statement to stop updating the latest_deploy_time if a recurring task failed. However not sure if the status down should be used as an indicator of a failed after_party rake task. 

Not tested. can you test tasks?
rake after_party:run currently outputs a puts string when a task fails. Also images below.
"failed tasks found, latest_deploy_time will not be updated!"


FireLemons (changes requested):
see discord

MclPio:
Behaviour where store_deploy_time runs first when year is 0000
Screenshot from 2024-10-15 19-43-00

Behaviour where store_deploy_time runs last when year is 9999
Screenshot from 2024-10-15 19-43-00

PR #6096, opened 2024-11-10
RSpec-Rails Configuration

thejonroberts:

None. Spec configuration only. I was looking at trying to speed up specs. This was the first step, making sure things were set up correctly.

### What changed, and _why_?

I updated `rspec-rails` and re-generated the configuration using `rails g rspec:install`. I also removed some things that were set up a little strangely (`upload_file` helper defined globally in FactoryBot initializer, plus moving some other helpers that made more sense per spec file rather than in rails helper).

We do get some benefits other than mere comments & organization:
- **Actually disabled monkey-patch syntax**. This was set in rails_helper, but the `rspec:install` generator put it in spec_helper, and that actually made a difference for some reason!
- **Moved fixture files into `spec/fixtures/files`**, allowing use of `file_fixture`/`file_fixture_upload` helpers. I kind of stumbled into this solution when I was trying to move `upload_file` helper out of the factory bot support initializer. It's much more convenient, worth moving all the file fixtures, imo.
- **Aggregate failures by default**, except in system specs. This was something I added, not generated. It helps to be able to see all the failures from a spec when working on an issue, imo. Also multiple expectations per spec make for less time in database/setup (as long as they are closely related).
- **Improved seed reproducibility while flake-hunting**.  Sets random order without needing the `--order` flag in .rspec file. Changes support file require sorting to be more deterministic. Also sets `Kernel.srand config.seed`. This will make using `--seed` flag more reliable - anything using Kernal.srand (Faker, probably?) will also use the same seed.
- Documentation format when running single file (instead of standard progress dots)

No behavior change, so no spec changes, aside from one flaky spec that I may have helped be less flaky. 🤞🏻


.gitignore
```diff
@@ -48,9 +48,6 @@ vendor/bundle/
 # produced by tests (?)
 .rakeTasks
 
-# produced by tests (?)
-*.csv
```
thejonroberts, line 52: Needed to be able to move csv file fixtures. Anything produced by specs should go into `tmp/`, anyway.

spec/fixtures/files/supervisor_volunteers.csv
```diff
@@ -1,3 +1,3 @@
-﻿email,display_name,supervisor_volunteers,phone_number
-s5@example.com,s5,volunteer1@example.net,11111111111
-s6@example.com,s6,volunteer1@example.net,11111111111
+﻿email,display_name,supervisor_volunteers,phone_number
+s5@example.com,s5,volunteer1@example.net,11111111111
+s6@example.com,s6,volunteer1@example.net,11111111111
```
thejonroberts, line 3: The diffs in the files here are from a CRLF -> LF change to appease `git mv`. No actual changes to any of the .csvs.

spec/rails_helper.rb
```diff
@@ -31,51 +56,61 @@
   config.include PunditHelper, type: :view
   config.include SessionHelper, type: :view
   config.include SessionHelper, type: :request
-  config.include TemplateHelper
-  config.include Warden::Test::Helpers
```
thejonroberts, line 35: redundant with Devise helpers, afaik

spec/rails_helper.rb
```diff
@@ -85,20 +120,31 @@
...
 
-  config.around :each, :disable_bullet do |example|
-    Bullet.raise = false
-    example.run
-    Bullet.raise = true
-  end
-
-  def pre_transition_aged_youth_age
-    Date.current - CasaCase::TRANSITION_AGE.years
-  end
```
thejonroberts, line 96: Weird place for a helper, and only used a handful of times in two files, I just replaced them in the files.

spec/rails_helper.rb
```diff
@@ -31,51 +56,61 @@
...
-  config.after do
-    Warden.test_reset!
-  end
-
-  Shoulda::Matchers.configure do |shoulda_config|
-    shoulda_config.integrate do |with|
-      with.test_framework :rspec
-      with.library :rails
-    end
-  end
```
thejonroberts, line 53: Moved to bottom (out of this `RSpec.config` block)

spec/rails_helper.rb
```diff
@@ -85,20 +120,31 @@
...
+  shoulda_config.integrate do |with|
+    with.test_framework :rspec
+    with.library :rails
+  end
+end
+
+WebMock.disable_net_connect!(
+  allow_localhost: true,
+  allow: "selenium_chrome:4444"
+)
```
thejonroberts, line 150: This was in a spec/support/ file (only this -- deleted the file, see below), I think it makes more sense here.

spec/spec_helper.rb
```diff
@@ -1,7 +1,16 @@
-require "pry"
-require "webmock/rspec"
-require "email_spec"
-require "email_spec/rspec"
```
thejonroberts, line 4: Moved to `rails_helper`, as per the lightweight comment below.

spec/support/factory_bot.rb
```diff
@@ -1,11 +1,4 @@
-def upload_file(file)
-  content = File.read(file)
-  tempfile = Tempfile.open
-  tempfile.write content
-  tempfile.rewind
-
-  Rack::Test::UploadedFile.new(tempfile)
-end
```
thejonroberts, line 8: This is a weird place for this to be defined! Global helper, in an unrelated file.

Rather than put in a helper file, I replaced its usage with [`fixture_file_upload`](https://apidock.com/rails/v7.1.3.4/ActionDispatch/TestProcess/FixtureFile/fixture_file_upload). 

By default, the method looks in `spec/fixtures/files`, so that is why I moved all of the .csv fixtures into that directory. Moving them was easier than figuring out how to config the fixture files directory, and felt like the standard way to do it.

spec/support/factory_bot.rb
```diff
@@ -30,24 +23,22 @@ def upload_file(file)
...
-    ActiveSupport::Notifications.subscribe("factory_bot.run_factory") do |name, start, finish, id, payload|
-      factory_name = payload[:name]
-      strategy_name = payload[:strategy]
-      factory_bot_results[factory_name] ||= {}
-      factory_bot_results[factory_name][:total] ||= 0
-      factory_bot_results[factory_name][:total] += 1
-      factory_bot_results[factory_name][strategy_name] ||= 0
-      factory_bot_results[factory_name][strategy_name] += 1
+  # This will output total database records being created.
+  if ENV.fetch("SPEC_OUTPUT_FACTORY_BOT_OBJECT_CREATION_STATS", false)
```
thejonroberts, line 27: No need to track these results if we don't want to see them, just moved this conditional here rather than for the after(:suite) puts below.

spec/support/shared_examples/shows_court_dates_links.rb
```diff
@@ -1,4 +1,4 @@
-shared_examples_for "shows court dates links" do
+RSpec.shared_examples_for "shows court dates links" do
```
thejonroberts, line 1: necessary now that monkey-patching is actually disabled.

spec/system/case_court_reports/index_spec.rb
```diff
@@ -8,12 +8,13 @@
...
 
+  let(:date) { Date.current }
+  let(:formatted_date) { date.strftime("%B %d, %Y") } # January 01, 2021
+
   before do
-    travel_to Date.new(2021, 1, 1)
     sign_in volunteer
     visit case_court_reports_path
   end
-  after { travel_back }
```
thejonroberts, line 16: This caused a flaky spec. Even thought there was an explicit `travel_back` here, the spec would fail because the user would get logged out... I'm guessing there's a race condition and `before { sign_in user }` happens in 2021, but we get back to present day before request/page load, and we redirect to root?

It's pretty common to use `travel_to` to test dates in ways that are more readable like this. In general, not just in this app. So may be something I'm missing here.

May be worth making a helper for asserting dates in specific formats on the page to avoid this elsewhere. Maybe even using i18n format yaml?

spec/rails_helper.rb
```diff
@@ -85,20 +120,31 @@
...
+  #   Bullet.raise = true
+  # end
 
   config.around do |example|
     Capybara.server_port = 7654 + ENV["TEST_ENV_NUMBER"].to_i
     example.run
   end
 
-  config.filter_run_excluding :ci_only unless ENV["GITHUB_ACTIONS"]
+  config.filter_run_excluding :ci_only unless ci_environment
```
thejonroberts, line 135: We don't use this tag currently, we could just remove it? Only use case I can think of is coverage reports, and we have an env flag for that (spec helper).

.gitignore
compwron: legit

Gemfile
```diff
@@ -96,7 +96,7 @@ group :test do
   gem "rails-controller-testing"
   gem "rake"
   gem "selenium-webdriver"
-  gem "simplecov"
+  gem "simplecov", require: false
```
compwron, line 99: reasonable

app/services/volunteer_birthday_reminder_service.rb
```diff
@@ -7,7 +7,7 @@ def send_reminders
           .deliver(volunteer.supervisor)
       end
     else
-      puts "Volunteer Birthday Reminder Rake task skipped. Today is not the 15th of the month."
+      Rails.logger.info { "Volunteer Birthday Reminder Rake task skipped. Today is not the 15th of the month." }
```
compwron, line 10: good change

spec/datatables/reimbursement_datatable_spec.rb
```diff
@@ -21,7 +21,7 @@
 
   # Requires the following to be defined:
   # - `sorted_case_contacts` = array of reimbursement records ordered in the expected way
-  shared_examples_for "a sorted results set" do
+  RSpec.shared_examples_for "a sorted results set" do
```
compwron, line 24: reasonable

spec/factories/casa_orgs.rb
```diff
@@ -10,7 +10,7 @@
     twilio_phone_number { "+15555555555" }
 
     trait :with_logo do
-      logo { Rack::Test::UploadedFile.new(Rails.root.join("spec", "fixtures", "org_logo.jpeg")) }
+      logo { Rack::Test::UploadedFile.new(Rails.root.join("spec", "fixtures", "files", "org_logo.jpeg")) }
```
compwron, line 13: reasonable

spec/views/notifications/index.html.erb_spec.rb
```diff
@@ -15,26 +14,26 @@
...
-        notification_1_day_ago.update_attribute(:created_at, base_time - 1.day)
-        notification_2_days_ago.update_attribute(:created_at, base_time - 2.days)
-        notification_3_days_ago.update_attribute(:created_at, base_time - 3.days)
+        notification_1_hour_ago.update!(created_at: 1.hour.ago)
+        notification_1_day_ago.update!(created_at: 1.day.ago)
+        notification_2_days_ago.update!(created_at: 2.days.ago)
+        notification_3_days_ago.update!(created_at: 3.days.ago)
 
-        patch_note_1.update_attribute(:patch_note_group, patch_note_group_all_users)
+        patch_note_1.update!(patch_note_group: patch_note_group_all_users)
```
compwron, line 36: these are importantly different and I suspect defeat the point of the test.

compwron:
Oh I seeeee
> Sets random order without needing the --order flag in .rspec file. Changes support file require sorting to be more deterministic. Also sets Kernel.srand config.seed. This will make using --seed flag more reliable - anything using Kernal.srand (Faker, probably?) will also use the same seed.

compwron (approved):
let's do ittt

spec/views/notifications/index.html.erb_spec.rb
thejonroberts: `update!` will fail if validation fails, `udpate_attribute` skips validation.  This is in a before block, so it's setup for the actual tests... so we should want it to fail if the setup is 'invalid'.

That was my thinking, anyway. Probably had rubocop-rspec running when I did this.

PR #6111, opened 2024-11-27
Update GH Workflow TOC Generator

marc:
n/a

### What changed, and _why_?
The TOC Generator step in GitHub actions sometimes tries to run against branches that no longer exist and fails. This config update should have it only check/run against the main branch.

This update also has the TOC Generator create a PR for readme changes instead of pushing them directly.

Need to see how this runs in GitHub actions.

### Feelings gif (optional)
😀 😀 
First time collaborating with @FireLemons! 🎉

### DIFF
--- a/.github/workflows/toc.yml
+++ b/.github/workflows/toc.yml
@@ -6,3 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: technote-space/toc-generator@v4
+        with:
+          CREATE_PR: true
+          GITHUB_TOKEN: ${{ secrets.ACCESS_TOKEN }}
+          CHECK_ONLY_DEFAULT_BRANCH: true
--- a/.github/workflows/toc.yml
+++ b/.github/workflows/toc.yml
@@ -11,5 +11,3 @@ jobs:
       - uses: technote-space/toc-generator@v4
         with:
           CREATE_PR: true
-          GITHUB_TOKEN: ${{ secrets.ACCESS_TOKEN }}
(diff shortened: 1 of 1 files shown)


aplchian:
Hey @marc! 👋 I see this change restricts the TOC generator to only run on the main branch, but I'm curious about a few architectural implications:

**Branch Strategy Questions:**
1. What happens if a long-running feature branch significantly modifies the README structure and then gets merged? Won't the TOC be out of sync during the entire development period of that branch? How do we ensure documentation consistency across branches?

2. By only running on `main`, we're essentially making the TOC a "post-merge" validation rather than a "pre-merge" validation. What's the rollback strategy if the TOC generator creates a problematic PR after a merge that breaks documentation?

**PR Creation Logic:**
3. The `CREATE_PR: true` flag is interesting - but what happens if multiple commits to main happen in rapid succession? Could we end up with conflicting TOC PRs that need manual resolution? How does the action handle PR collision scenarios?

4. Since this creates PRs automatically, who becomes the reviewer? Is there a risk of infinite loops if the TOC PR itself triggers another workflow run?

**Testing Concerns:**
5. You mentioned "Need to see how this runs in GitHub actions" - but how do we test failure scenarios? What if the TOC generator fails to create a PR due to permissions, rate limits, or API issues? Do we have monitoring/alerting for this workflow?

6. Is there a strategy for testing this workflow change without potentially spamming the repo with test PRs during development?

These might seem like edge cases, but they could become real issues at scale! What are your thoughts? 🤔

PR #6086, opened 2024-10-23
Standard.rb linting -> Rubocop and relevant plugins

thejonroberts:
### What github issue is this PR for, if any?
No Issue.

### What changed, and _why_?
This project welcomes and encourages new contributors who may not have a lot of experience with idiomatic ruby or common patterns to avoid in rails/rspec/capybara, etc...  So having good linting in place is important -- both to maintain code quality and to give newer devs a tool to learn about common patterns & good practices. 

It also makes review easier because there are fewer style nits & common gotchas to get hung up on. Maintainers can focus on more important things!  And, to be honest, I love linters because I don't want to have to think about all these things!

It will also create a nice set of TODOs for people that want to start contributing! An Epic issue similar the the xit/skipped specs epic could be created to fix the todos file by file.

This rubocop setup is based on [this evil martians article on adding standard/rubocop to a legacy project](https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard).

- Move from standardrb to rubocop so that we can add rubocop-* extensions for various tooling:
  - rails, rspec, rspec_rails, factory_bot, capybara
- Configure the rules in `.rubocop.yml`
  - Inherit from the standard/standard-rails rules and try to not fuss with defaults too much.
  - Fix some of the more important/performance rules & make sure they aren't ignored in todo.
- Generate TODO file so that extreme change to codebase is avoided & CI can incorporate immediately.
  - Comments in rubocop.yml on how to find todo issues to fix.
- Replace standard for rubocop in `bin/lint`, GitHub CI lint workflow, & git hooks.
- Remove current `rubocop:disable` comments as any offenses will be in the todo file.

I made a PR into this branch with all of the [autocorrections here](https://github.com/thejonroberts/casa/pull/9). If we want to go ahead and include those, we can just merge that into this PR.  May cause some minor merge conflicts for works in progress. Even if not, the PR will give an idea of the impact this will have on the codebase.

Things to consider:
- I tried to minimize the amount of configuration. Generally, I agree with the Standard.rb approach, less config the better. However, I also have some strong opinions! Standard.rb/standard-rails' rulesets are also decided by working ruby/rails devs who know their way around, and don't necessarily line up with the needs of CASA or the current codebase.
[...]

PR #6120, opened 2024-12-05
Linting: Standard rubocop plugins

thejonroberts:
After discussion in #6086 , I wanted to try standard/rubocop plugin linting options with less/zero config. As suggested by @compwron 

See above closed PR for even more context, but my reasoning for adding all the rubocop plugins is:

> This project welcomes and encourages new contributors who may not have a lot of experience with idiomatic ruby or common patterns to avoid in rails/rspec/capybara, etc... So having good linting in place is important -- both to maintain code quality and to give newer devs a tool to learn about common patterns & good practices.
> 
> It also makes review easier because there are fewer style nits & common gotchas to get hung up on. Maintainers can focus on more important things! And, to be honest, I love linters because I don't want to have to think about all these things!
> 
> It will also create a nice set of TODOs for people that want to start contributing! An Epic issue similar the the xit/skipped specs epic could be created to fix the todos file by file.

### What changed, and _why_?
Adds rubocop plugins relevant to the project, standard config extends them. PR applies auto fixes, and generates a standard-todo.yml file.

I can revert auto fixes and add them to the todo if this is too big a change, I think it makes more sense to have fewer todos, and I wanted people to be able to see the differences mentioned below.

NOTE:  We could have a middle ground with some configuration inherited by standard. I have broken up commits to allow dropping some of them to change this PR.
- [This commit removes all configuration](https://github.com/rubyforgood/casa/pull/6120/commits/73be14d82437755d3f8f4a1c1294b9180041c4cd). View the link to see the difference. A lot of changes to current spec patterns will be necessary.
- [This commit switches from rubocop-rails defaults to standard-rails plugin](https://github.com/rubyforgood/casa/pull/6120/commits/78951169921f5181684a4e9cfa7b2bd115c81968). To me, they both have their plusses/minuses. I think standard-rails is more suited to experienced rails devs with regards to most of their decisions (default scopes, for example).

I think my ideal, for this project, would be to drop those two commits. But if wanting to avoid config/bikeshedding is a priority, they make a lot of sense.

Current tests only, as no behavior changes. Only safe-autocorrections have been applied.


compwron (approved):
I am on board. We may see merge conflicts in other contributor PRs and we should be extra helpful to them when they arise.

PR #6146, opened 2024-12-23
fix: delete boilerplate favicon on login page

willmurphyscode:
Previously, network tools show a 404 getting /users/assets/images/favicon.svg on /users/sign_in, and some tests intermittently failed with this 404. Delete the extraneous favicon link from devise boilerplate to let the normal favicon work on the login screen.

Fixes test flake at
https://github.com/rubyforgood/casa/wiki/Flaky-tests#suspected-testing-tool-crash

Resolves [flaky test at ](https://github.com/rubyforgood/casa/wiki/Flaky-tests#suspected-testing-tool-crash). The test reports sometimes failing with `No route matches [GET] "/users/assets/images/favicon.svg"`.

### What changed, and _why_?

The devise-generated view at `app/views/layouts/devise.html.erb` had its own favicon, which would make 404s on the page, and these 404s would fail some tests. Deleting this link tag lets the normal favicon present on other pages work.

Note: I initially tried using `href="<%= asset_path("favicon.svg")%."` instead, but `favicon.svg` is committed directly at `public/assets/images/favicon.svg`, so Rails doesn't seem to actually have an asset path for it. I don't know why this was done, so I don't want to undo it. The file at `public/assets/images/favicon.svg` does not seem to be the right favicon for the project.

_Note: We love [capybara](https://rubydoc.info/github/teamcapybara/capybara) tests! If you are writing both haml/js and ruby, please try to test your work with tests at every level including system tests like https://github.com/rubyforgood/casa/tree/main/spec/system_ 

On `main`, in dev tools, you can see a 404 on `GET http://0.0.0.0:3000/users/assets/images/favicon.svg`. On this branch, the favicon loads and is the expected CASA logo.

I'm happy to add an automated test for this, but I wasn't sure where one belonged.

One important note is that during testing, I was not able to make the failure described at https://github.com/rubyforgood/casa/wiki/Flaky-tests#suspected-testing-tool-crash happen on purpose, so this is a bit of a speculative fix. But it _does_ fix a 404 that was happening during local development, and that currently happens in prod on https://casavolunteertracking.org/users/sign_in

### Screenshots please :)

Before:
- browser tab: image
- network tools: image

After:
- browser tab: image

- network tools: image

### DIFF
--- a/app/views/layouts/devise.html.erb
+++ b/app/views/layouts/devise.html.erb
@@ -4,10 +4,6 @@
   <meta charset="UTF-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <link
-    rel="shortcut icon"
-    href="assets/images/favicon.svg"
-    type="image/x-icon">
   <title>CASA Volunteer Tracker</title>
 
   <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>


willmurphyscode:
Maybe we should change favicon to https://casavolunteertracking.org/favicon.ico instead? That seems to be what's actually used in prod, and seems to be an actual logo of the org. Will update the PR.

willmurphyscode:
Hi friends! I think this is ready for review.

1. I can't reproduce the docker test failure. I think it is probably a flake. (To attempt: I got the docker tests running locally and re-ran with the same seed number that failed in CI, and it didn't fail.)
2. I don't think adding a system test that says something like, "the login in page doesn't add a random extra favicon" is a super useful test. The test will be slow and fiddly (because a JS library actually chooses the icon), and both the impact of a regression (e.g. login page has no favicon) and the likelihood of a regression (someone re-adds copy/paste boilerplate to the devise view) are super low.

Let me know if there's anything else I should do to get this in.

compwron (approved):
Great :)

PR #6162, opened 2025-01-13
fix: NoMethodError users/passwords#create

compwron:
https://app.bugsnag.com/ruby-for-good/casa/errors/65c96d515ff67f00092d7e7b?filters[event.unhandled]=true&filters[event.since]=30d&filters[app.release_stage]=production
Screenshot 2025-01-12 at 8 39 38 PM


elasticspoon (commented):
This fixes the error. That said I think this error is from some botting or malicious requests in some form. All the requests are json which makes no sense, the emails are gibberish and none of them use US ips.

compwron:
Yeah, I just want to stop having errors which are not things we need to fix

PR #6216, opened 2025-02-06
Feature/6207 Endpoint api/v1/sign_in now outputs randomized tokens in response upon sign-in

### LINKED ISSUE #6207: Enhance endpoint security for /api/v1/users/sign_in with randomization
Part of epic #3942

### What type of user does this affect?
 - volunteers
 
### How should it operate? ⚙️🛠️
Endpoint can now output randomized `token` and `refresh_token` on successful `/api/v1/sign_in` request.

Additionally, since tokens now expire,  `user model` has  `token_expires_at` and `refresh_token_expires_at` datetimes.

### Acceptance Criteria
- [x] Expands the `user model` to handle both the `token` and `refresh token`, as well as their expiration times.
- [x] Creates functions to generate `randomized` tokens in `user.rb` and set default expiration to `7 hours` and `30 days`
- [x] Updates `session controller` and `base controller` with new details if applicable

Edit: Made api_credentials table and put randomized functions in new models/api_credential.rb that belongs to user

7riumph:
Resolves #6207

### What changed, and _why_?
Volunteers  api/v1/sign_in endpoint now outputs secured randomized tokens upon sign-in.

- Dropped `token` column from `users` table
- Added `api_credentials` table to be used for `users` model for greater `separation of concerns` (SoC)
- Utilized `SHA-256` hash to ensure `raw tokens` are never stored in `db` ( `api_token_digest` & `refresh_token_digest` )
- Created `models/api_credentials` and `concern/api.rb` for further `api` concern separation and `utility`

_Why_?: To ensure compliance with apples review process, and enhance security with an _applicable_ industry standard.

New tests:

`Authentication Helper Function tests (14 in total)` &rarr; spec/models/api_credential_spec.rb
`Credentials Factory Definition` &rarr; spec/factories/api_credential.rb

Updated tests ( no more token column ):

`BaseController test` &rarr; spec/requests/api/v1/base_spec.rb
`SessionController test` &rarr; spec/requests/api/v1/users/sessions_spec.rb
`Users Factory Definition` &rarr; spec/factories/users.rb

Deployment & Documentation:

`Deployment task` &rarr; lib/tasks/deployment/20230822145532_populate_api_tokens.rake
`Swagger Documentation` &rarr; swagger/v1/swagger.yaml
`More Swagger`&rarr; spec/swagger_helper.rb

### Screenshots please :)
Example output with updated api/v1/sign_in endpoint [blueprint](https://github.com/rubyforgood/casa/blob/778758eb7081a6a883a623a74f18c7e70b1b0b1b/app/blueprints/api/v1/session_blueprint.rb)

example_output_sign_in_success

### Feelings gif (optional)
very strange

### DIFF
--- a/app/controllers/api/v1/users/sessions_controller.rb
+++ b/app/controllers/api/v1/users/sessions_controller.rb
@@ -4,7 +4,7 @@ class Api::V1::Users::SessionsController < Api::V1::BaseController
     if @user
       render json: Api::V1::SessionBlueprint.render(@user), status: 201
     else
-      render json: {message: "Wrong password or email"}, status: 401
+      render json: {message: "Incorrect email or password."}, status: 401
     end
   end
 
--- a/app/models/api_credential.rb
+++ b/app/models/api_credential.rb
@@ -1,0 +1,3 @@
+class ApiCredential < ApplicationRecord
+  belongs_to :user
+end
--- a/app/models/api_credential.rb
+++ b/app/models/api_credential.rb
@@ -1,12 +1,76 @@
+require "digest"
+
 class ApiCredential < ApplicationRecord
   belongs_to :user
 
-  before_save :hash_tokens
+  before_save :generate_api_token
+  before_save :generate_refresh_token
+
+  # Securely confirm/deny that Hash in db is same as current users token Hash
+  def authenticate_api_token(api_token)
+    Digest::SHA256.hexdigest(api_token) == api_token_digest
+  end
+
+  def authenticate_refresh_token(refresh_token)
+    Digest::SHA256.hexdigest(refresh_token) == refresh_token_digest
+  end
+
+  # Securely generate and then return new tokens
+  def return_new_api_token!
+    generate_api_token
+    save!
+    {api_token: @api_token}
+  end
+
+  def return_new_refresh_token!
+    generate_refresh_token
+    save!
+    {refresh_token: @refresh_token}
+  end
+
+  # Verifying token has or has not expired
+  def is_api_token_expired?
+    token_expires_at < Time.current
+  end
+
+  def is_refresh_token_expired?
+    refresh_token_expires_at < Time.current
+  end
 
   private
 
-  def hash_tokens
-    self.api_token_digest = Digest::SHA256.hexdigest(api_token) if api_token_changed?
-    self.refresh_token_digest = Digest::SHA256.hexdigest(refresh_token) if refresh_token_changed?
+  # Generate unique tokens and hashes them for secure db storage
+  def generate_api_token
+    @api_token ||= SecureRandom.hex(18)
+    self.api_token_digest = Digest::SHA256.hexdigest(@api_token) if @api_token.present?
+  end
+
+  def generate_refresh_token
+    @refresh_token ||= SecureRandom.hex(18)
+    self.refresh_token_digest = Digest::SHA256.hexdigest(@refresh_token) if @refresh_token.present?
   end
 end
+
+# == Schema Information
+#
+# Table name: api_credentials
+#
+#  id                       :bigint           not null, primary key
+#  api_token_digest         :string
+#  refresh_token_digest     :string
+#  refresh_token_expires_at :datetime
+#  token_expires_at         :datetime
+#  created_at               :datetime         not null
+#  updated_at               :datetime         not null
+#  user_id                  :bigint           not null
+#
+# Indexes
+#
+#  index_api_credentials_on_api_token_digest      (api_token_digest) UNIQUE WHERE (api_token_digest IS NOT NULL)
+#  index_api_credentials_on_refresh_token_digest  (refresh_token_digest) UNIQUE WHERE (refresh_token_digest IS NOT NULL)
+#  index_api_credentials_on_user_id               (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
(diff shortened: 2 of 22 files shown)


xihai01:
looks good to me so far 🔥

I think instead of storing the token and refresh token as plaintext in the db, we can store it as a hash
like what shen suggested

Then the client would be provided the plaintext versions of the tokens and whenever they call the api, the hash versions will be compared

app/controllers/api/v1/users/sessions_controller.rb
```diff
@@ -2,6 +2,10 @@ class Api::V1::Users::SessionsController < Api::V1::BaseController
   def create
     load_resource
     if @user
+
+      @user.regenerate_session_token!
+      @user.regenerate_refresh_token!
+
       render json: Api::V1::SessionBlueprint.render(@user), status: 201
     else
       render json: {message: "Wrong password or email"}, status: 401
```
xihai01, line 11: optional: let's update the error message to "Incorrect email or password."
similiar to how other apps do it.

image

app/controllers/api/v1/users/sessions_controller.rb
```diff
@@ -2,6 +2,10 @@ class Api::V1::Users::SessionsController < Api::V1::BaseController
   def create
     load_resource
     if @user
+
+      @user.regenerate_session_token!
+      @user.regenerate_refresh_token!
```
compwron, line 7: don't do this twice?

app/models/supervisor.rb
```diff
@@ -81,10 +81,13 @@ def recently_unassigned_volunteers
 #  receive_email_notifications   :boolean          default(TRUE)
 #  receive_reimbursement_email   :boolean          default(FALSE)
 #  receive_sms_notifications     :boolean          default(FALSE), not null
+#  refresh_token                 :string
```
compwron, line 84: maybe make a new type of user, ApiUser?

app/models/user.rb
```diff
@@ -7,10 +7,12 @@ class User < ApplicationRecord
   include DateHelper
 
   before_save :normalize_phone_number
+  before_create :set_token_expiration
   after_create :skip_email_confirmation_upon_creation
   after_create :create_preference_set
   before_update :record_previous_email
   has_secure_token :token, length: 36
+  has_secure_token :refresh_token, length: 36
```
compwron, line 15: 36 what? seconds, minutes, days, weeks?

7riumph: Vague syntax refers to token length, in this case. The token is 36 characters long.

app/controllers/api/v1/users/sessions_controller.rb
7riumph: It is actually for two separate tokens. The plan is the `session_token` expires quickly and is repeatedly replaced for as long as the `refresh_token` is valid to maintain access.

app/models/supervisor.rb
7riumph: Created the [concern](https://github.com/rubyforgood/casa/blob/778758eb7081a6a883a623a74f18c7e70b1b0b1b/app/models/concerns/api.rb)

app/models/api_credential.rb
```diff
@@ -0,0 +1,78 @@
+require "digest"
+
+class ApiCredential < ApplicationRecord
+  belongs_to :user
+
+  before_save :generate_api_token
+  before_save :generate_refresh_token
+
+  # Securely confirm/deny that Hash in db is same as current users token Hash
+  def authenticate_api_token(api_token)
```
xihai01, line 10: does these functions return boolean value?
if so, the naming should be consistent with other similar functions starting with the is_....

db/migrate/20250207080511_create_api_credentials.rb
```diff
@@ -0,0 +1,19 @@
+class CreateApiCredentials < ActiveRecord::Migration[7.2]
+  def change
+    create_table :api_credentials do |t|
+      t.references :user, null: false, foreign_key: true
+      t.string :api_token
```
xihai01, line 5: you're also storing the plain text tokens in the table?

xihai01: nvm I think you created a migration to remove them right?

xihai01 (commented):
lgtm

app/models/api_credential.rb
7riumph: Yes, returns a boolean. How's Something like "is_api_token_valid"?

db/migrate/20250207080511_create_api_credentials.rb
7riumph: Yes, originally added unneeded token columns then created a migration to remove them.

xihai01: cool

app/models/api_credential.rb
xihai01: yes that would be perfect
also follows ruby conventions too

lib/tasks/deployment/20230822145532_populate_api_tokens.rake
```diff
@@ -4,9 +4,9 @@ namespace :after_party do
     puts "Running deploy task 'populate_api_tokens'" unless Rails.env.test?
 
     # Put your task implementation HERE.
-    User.where(token: nil).each do |user|
+    User.find_each do |user|
```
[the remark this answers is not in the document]
xihai01: I see I see
I forgot I originally created this
yea, just leave then

PR #6224, opened 2025-02-13
Attempt to address volunteer#create invalid email error in bot-bugsag

7riumph:
Resolves #NaN

Noticed an error with `volunteers#create` through the discord. 

Likely due to an invalid or non standard email.

I refactored it to ensure in the volunteers_controller `create` function....
- Invitations to Invalid emails throw an alert

### DIFF
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -34,12 +34,11 @@ class VolunteersController < ApplicationController
     if @volunteer.save && @volunteer.email.match?(URI::MailTo::EMAIL_REGEXP)
       @volunteer.invite!(current_user)
       # call short io api here
-
-      short_url_key_value_pairs = @volunteer.phone_number.blank? ? {0 => nil, 1 => nil} : handle_short_url([Rails.application.routes.url_helpers.accept_user_invitation_url(
-        invitation_token: @volunteer.raw_invitation_token, host: request.base_url
-      ), request.base_url + "/users/edit"])
-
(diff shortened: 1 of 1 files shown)


app/controllers/volunteers_controller.rb
```diff
@@ -31,14 +31,17 @@ def create
     @volunteer = current_organization.volunteers.new(create_volunteer_params)
     authorize @volunteer
 
-    if @volunteer.save
+    if @volunteer.save && @volunteer.email.match?(URI::MailTo::EMAIL_REGEXP)
```
compwron, line 34: good

compwron (commented):
Needs specs but a good start!

FireLemons:
The regex addition didn't solve the real life error case of an email address with a close parentheses at the end `xxx@gmail.com)` and I think it's because it's technically a correct email format according to the RFC `URI::MailTo::EMAIL_PATTERN` is based off of so instead I think it's better to error handle email send attempts.

for
```
    if @volunteer.save
      @volunteer.invite!(current_user)
      # call short io api here
      invitation_url = Rails.application.routes.url_helpers.accept_user_invitation_url(invitation_token: @volunteer.raw_invitation_token, host: request.base_url)
```
wrap @volunteer.invite!(current_user) in a try catch so not only does it catch emails with invalid formats i.e. xxx@gmail.com)(from bugsnag) but it will also catch domains and email addresses that don't exist

compwron:
how's it going?

7riumph:
> how's it going?

Will get back to it now.

PR #6285, opened 2025-03-30
feat: use rspec retry to deflake ci

elasticspoon:
Adds `rspec-retry` to compensate for some flakiness by retrying failures.

The basic issue in common with the failures is a pattern like:

- perform action on page
- assert some database state

The issue is that the tests are often faster than the server, thus, the check it the action worked happens before the action finished. Ex:

- click submit to create case contact
- controller gets the request and starts working
- check that count of contacts went up => fails controller not done yet
- controller finishes

Thus, we can use capybara waits to fix this. Basically we want to check that the front end shows the action finished (we see an alert) then check the database state.

- perform action on page
- check for front end flash message (capybara will wait on these)
- assert some database state

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -98,6 +98,7 @@ end
 group :test do
   gem "brakeman" # security inspection
   gem "capybara"
+  gem "rspec-retry" # for retrying flaky tests
   gem "capybara-screenshot"
   gem "database_cleaner-active_record"
   gem "docx"
--- a/spec/system/casa_cases/edit_spec.rb
+++ b/spec/system/casa_cases/edit_spec.rb
@@ -238,17 +238,20 @@ RSpec.describe "Edit CASA Case", type: :system do
       expect(page).to have_text("Set Implementation Status")
 
       find(".ts-control").click
-      ts_checkboxes = page.all(".ts-dropdown-content input")
 
       select_all_el = page.find("span[data-test=select-all-input]")
       # uncheck all contact type options
       select_all_el.click
-      ts_checkboxes.each do |el|
-        expect(el).not_to be_checked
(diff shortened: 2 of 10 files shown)


spec/system/casa_cases/edit_spec.rb
```diff
@@ -44,17 +44,21 @@
...
+        expect(page).not_to have_css(".form-check-input--checked")
+        expect(page).to have_css(".form-check-input--unchecked", count: 3)
       end
       # check all contact type options
       select_all_el.click
-      expect(ts_checkboxes).to all(be_checked)
+      within ".ts-dropdown-content" do
+        expect(page).not_to have_css("input.form-check-input--unchecked")
+        expect(page).to have_css("input.form-check-input--checked", count: 3)
+      end
```
elasticspoon, line 61: This is an ugly fix but the issue with the initial code is that when you run `checked?` on the nodes rspec won't use any capybara waiting. It checks right at the moment and either passes or fails. 

Thus, if the test runs that check before the javascript on the page runs the test will fail. When we use `have_` selectors we can avoid this issue. 

I would like to have done something more like `have_checked("id_of_input") but tomselect does not give us something like that. instead I added a css class for checked and unchecked inputs in tomselect and assert against that.

compwron:
:)

Gemfile
```diff
@@ -98,6 +98,7 @@ end
 group :test do
   gem "brakeman" # security inspection
   gem "capybara"
+  gem "rspec-retry" # for retrying flaky tests
```
compwron, line 101: https://rubygems.org/gems/rspec-retry seems reasonable, last update 2019 but simple

PR #6405, opened 2025-05-28
Fix #6404: Trim whitespace from custom link text before saving

nizam12khan:
Resolves #6404 

 - Added a `before_save` callback in `CustomOrgLink` to trim whitespace from the `text` field before saving.
- Added a migration to trim whitespace from existing `CustomOrgLink` records (please review if this is desired). Used `up`/`down` methods since this is a data migration and not reversible. Tested locally to confirm it works.

db_before
DB_after

Issue #6404, opened 2025-05-28
Trim whitespace before saving custom link name in database


FireLemons:
## Description
Trim the whitespace before  saving the displayed text of the link in the database
See: https://github.com/rubyforgood/casa/pull/6350

No tests required. Just prove it works.

## Screenshots of current behavior, if any
![Image](https://github.com/user-attachments/assets/7846cc57-011b-4daf-bfe5-43c0ea4d1d46)
![Image](https://github.com/user-attachments/assets/d568964c-177c-44b5-a7bd-4e9e49d04b91)

## Logging in to test environments
_Login Details:_  

Login Emails: 
- volunteer1@example.com  view site as a volunteer
- supervisor1@example.com view site as a supervisor
- casa_admin1@example.com view site as an admin
- all_casa_admin1@example.com view site as an all casa admin
  - go to `/all_casa_admins/sign_in`  

password for all users: 12345678  

### Questions? Join Slack!

We highly recommend that you join us in [slack](https:https://join.slack.com/t/rubyforgood/shared_invite/zt-35218k86r-vlIiWqig54c9t~_LkGpQ7Q) #casa channel to ask questions quickly. And [discord](https://discord.gg/qJcw2RZH8Q) for office hours (currently Tuesday 5-7pm Pacific), stakeholder news, and upcoming new issues.

nizam12khan:
I'd like to work on this issue. Can I take it?

nizam12khan:
Hi! I noticed the issue wasn’t assigned when I started working on it, so I went ahead and submitted a PR. Hope that’s okay! Let me know if I should follow a different process next time. 😊

compwron:
> Hi! I noticed the issue wasn’t assigned when I started working on it, so I went ahead and submitted a PR. Hope that’s okay! Let me know if I should follow a different process next time. 😊

all good :)

PR #6407, opened 2025-05-29
Add spec for trim_name method in CustomOrgLink model 

nizam12khan:
**Related Issue**: #6404

**Changes**:
- Added an RSpec test for the `trim_name` method in `CustomOrgLink` model, which was implemented in #6405 
- The spec verifies that `trim_name` trims whitespace from the `text` attribute when present 
- Used `save` to trigger the `before_save` callback.

**Testing**:
- Ran `rspec spec/models/custom_org_link_spec.rb` and confirmed the tests pass.


compwron (approved):
Nice, thanks!! :)

PR #6499, opened 2025-09-13
Un-flaking tests pt 2

h-m-m:

@gvt posted some tests that were flaky on his local machine in Slack at https://rubyforgood.slack.com/archives/C09ERPYTSVB/p1757699757608479?thread_ts=1757697989.705529&cid=C09ERPYTSVB

This addresses most of those to make them less flaky.

A few I skipped because
* It was a chromdriver crash and @gvt has an install I have no experience with
* The fix was less obvious to me than the rest

### What changed, and _why_?

Most of these changes are

* Don't `visit` another path after hitting a button as you can cause the browser to interrupt an existing submit action before Rails has processed it
  * Match on `has_current_path` when the page should change
  * Match on content in the page that should be different from before if the path does not change — some of these `visit` calls were visiting a page we were already on
* This change also removes the `x` to no longer skipped tests that also seem to have the same problem.

I've done a few test runs and it works okay so far. I'd like @gvt to try some of these locally and see if it helps

### Screenshots please :)
n/a

### Feelings gif (optional)
alt text


gvt (approved):
The code changes look good to me. When i checkout this branch and run the system specs, I still get some failures, but fewer and faster (just under 5 minutes) which def is an improvement.

[iTerm2 Session Sep 13, 2025 at 3:08:38 PM.txt](https://github.com/user-attachments/files/22314035/iTerm2.Session.Sep.13.2025.at.3.08.38.PM.txt)

compwron (approved):

PR #6491, opened 2025-09-12
Add Placement Types management functionality

abachman:
### What github issue is this PR for, if any?

Resolves #5778, which will resolve #4695.

## What changed, and _why_?

Added control of PlacementType records for organization admins via the existing /casa_org/:id/edit page / flow.

Admins can create new and update existing placement types belonging to their organization.

### Additional Changes

- removed the standardrb config line from `.rubocop.yml` to prevent errors caused by the rubocop + standard upgrades in #6488 
- updated `bundle exec erblint` to `bundle exec erb_lint` in response to a deprecation wraning

### How is this **tested**? (please write rspec and jest tests!) 💖💪

- added policy specs, view specs, and request specs covering the changed code

### Screenshots

Placement Types listed on `/casa_org/:id/edit`:

image

Editing an existing Placement Type form:

image

Creating a new Placement Type form:

image

PR #6506, opened 2025-09-13
chore: remove superfluous test code

abachman:
### What changed, and _why_?

In the process of working on #6491, I noticed that there was extra view spec setup code that I had copy-pasted into my new view specs.

Turned out to be superfluous, so I tried take the same code out everywhere else it appeared. These are the tests where that could be done safely!

### How is this **tested**?
 
Deleted code, Rspec suite still passes: ✅ 

### Screenshots please :)

image


gvt (dismissed):

compwron (dismissed):
lgtm if CI goes green

PR #6669, opened 2026-01-22
Remove global span tag CSS

### LINKED ISSUE #5377: Update <span> tag CSS globally
**Description**
- We are globally overriding the `<span>`   tag's inline display behavior to the block as per below:
- https://github.com/rubyforgood/casa/blob/7dc1c6bd7aaf587e1ee146fd50218d470076375d/public/assets/css/main.css#L56-L59
- IMO, this kills the purpose of the span tag in general
- We have other alternatives to achieve the display block behavior e.g. `<p>, <div>` etc
- We need to remove the code chunk shown above (_surpizingly there's a syntax error in the current main.css as well at line#59 and it's not breaking_:sweat_smile:)
- Along with that we need to update the whole application where the current span styling is required or breaks the UI in any way. For example, removing this chunk breaks the styles on the Supervisors index page view

**Existing global span style needs deleting**
- https://github.com/rubyforgood/casa/blob/7dc1c6bd7aaf587e1ee146fd50218d470076375d/public/assets/css/main.css#L56-L59

**New style(s) needs creating**
- New styles or alternatives needs to be created/replaced throughout the app to ensure that nothing breaks after the removal of above chunk

Sergio-Mira:
Resolves #5377

### What changed, and _why_?
Remove global span CSS and instead adds new CSS class for volunteer info in the Supervisors table where it was previously using `status-btn` (its not a button!).

It changes a bit the style in the Admin edit page (see screenshot) but I prefeer the status not taking the whole line but I can work on making sure that also looks the same as prod.

Tested comparing `main` and this branch in all pages to check styles match.

### Screenshots please :)
Screenshot 2026-01-22 at 12 34 52
Screenshot 2026-01-22 at 12 40 13
(note current status does no longer take the whole line)

### Feelings gif (optional)
https://giphy.com/gifs/XIqCQx02E1U9W


compwron (approved):
lgtm! :)

PR #6708, opened 2026-02-10
Upgrade node to v24

### LINKED ISSUE #6688: Update node to v24
node v24 "Krypton" is the latest LTS node version. If we don't update it eventually the deploy will break because heroku will stop supporting v22.

See https://github.com/rubyforgood/casa/wiki/Nodejs-update-checklist

stefannibrasil:

Closes #6688

### What changed, and _why_?

- upgrade node to 24.x since 22.x is not being supported anymore (only security patches), and heroku will require newer versions soon
- Dockerfile: the image couldn't be built without explicitly downloading this node's version. Maybe on Ruby's 3.3.10 alpine, it will be available and we don't need these lines.

My local env generated some `peer: true` changes in the `package.json` file. However, this seems to be a bug related to npm. I removed them from this PR to isolate the node changes.

### How is this **tested**?

Tests are passing! I also built js and css, and checked the app locally and didn't find any errors, and no console's errors either:

https://github.com/user-attachments/assets/c48ceb43-b6f8-4ec8-ad07-4084cc4e7e76

I am also happy to QA this branch in the QA env, if needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -82,7 +82,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.28.6.tgz",
       "integrity": "sha512-H3mcG6ZDLTlYfaSNi0iOKkigqMFvkTKlGUYlD8GW7nNOYRrevuA46iTypPyv+06V3fEmvvazfntkBU34L0azAw==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.28.6",
         "@babel/generator": "^7.28.6",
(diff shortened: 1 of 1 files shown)


doc/LINUX_SETUP.md
```diff
@@ -54,9 +54,9 @@ sudo -u postgres psql -c "CREATE USER $USER WITH CREATEDB"
 ```
 # Install NVM and Node JS
 #   you can use curl
-curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh | bash
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.4/install.sh | bash
```
stefannibrasil, line 57: Extracted from https://github.com/nvm-sh/nvm?tab=readme-ov-file#install--update-script

I do not use Linux, so I couldn't verify this works, though.

FireLemons:
:medal_sports:

PR #6712, opened 2026-02-12
Correctly Configure Docker Node Pinning

FireLemons:
Relying on the ruby alpine image for node does not allow us to specify which major version we use. It was the solution for a while because the version of node we wanted just happened to be the same as the node from that image.

Currently we are using an old solution that does allow for custom node version in our image. However the repo we are getting node from is the latest alpine image which has a version of node that updates somewhat frequently which makes the method of pinning in this solution cause a conflict as soon as it updates.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,10 @@
 ARG ROOT=/usr/src/app/
 
-FROM ruby:3.3.8-alpine AS builder
+# available alpine packages: https://pkgs.alpinelinux.org/packages
+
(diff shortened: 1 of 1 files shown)


FireLemons:
@stefannibrasil we had an incorrect docker config. Just wanted to ping you in case you picked up some bad info from working on it earlier

stefannibrasil:
oh, interesting, thanks for the heads up. What error were you getting?

PR #6722, opened 2026-02-16
Fix race conditions in spec/system/all_casa_admins/

### LINKED ISSUE #6694: Race Condition Check: spec/system/all_casa_admins/
## Overview

Examples:
#### Good `expect`s
```ruby
  
  wait_for_download
```

#### Bad/Suspicious `expect`s
```ruby
  
```

## Things to do

 - Check these files:
    - [ ] spec/system/all_casa_admins/all_casa_admin_spec.rb
    - [ ] spec/system/all_casa_admins/edit_spec.rb
    - [ ] spec/system/all_casa_admins/password_change_spec.rb
    - [ ] spec/system/all_casa_admins/patch_notes/index_spec.rb
    - [ ] spec/system/all_casa_admins/sessions/new_spec.rb

vishaltps:
Fixes #6694

## Summary
- Replace all non-webpage `expect`s with page-level assertions in `spec/system/all_casa_admins/` to fix race conditions in CI
- Replace `CasaOrg.find_by` DB query with regex path match on the page
- Remove `ActionMailer::Base.deliveries` checks (email delivery content is not visible on the webpage; existing page assertions already confirm success)
- Remove `CasaAdmin.find_by` invitation check (already confirmed by page assertion "New admin created successfully")
- Remove "admin invitations expire" test (pure model test with zero browser interaction — does not belong in a system spec)
- Rewrite `expect { visit "/flipper" }.to raise_error(ActionController::RoutingError)` to check the webpage by temporarily enabling exception rendering

## Files checked

All 5 files from the issue checklist were reviewed:

| File | Result |
|------|--------|
| `all_casa_admin_spec.rb` | Fixed — 5 non-webpage expects replaced/removed |
| `sessions/new_spec.rb` | Fixed — 1 non-webpage expect rewritten |
| `edit_spec.rb` | Clean — all expects already check the page |
| `password_change_spec.rb` | Clean — all expects already check the page |
| `patch_notes/index_spec.rb` | Clean — all expects already check the page |

## Follow-up
The "admin invitations expire" test was removed from the system spec. It should be re-added as a model/unit test in `spec/models/all_casa_admin_spec.rb` (which does not currently exist).

## Test plan
- [ ] `bundle exec rspec spec/system/all_casa_admins/` — all 17 examples pass, 0 failures

### DIFF
--- a/spec/system/all_casa_admins/all_casa_admin_spec.rb
+++ b/spec/system/all_casa_admins/all_casa_admin_spec.rb
@@ -52,6 +52,7 @@ RSpec.describe "all_casa_admins/casa_orgs/casa_admins/new", type: :system do
     fill_in "Address", with: "123 Main St"
     click_on "Create CASA Organization"
     expect(page).to have_text "CASA Organization was successfully created."
+    expect(page).to have_text "Cool Org Name"
     expect(page).to have_current_path(%r{/all_casa_admins/casa_orgs/\d+}, ignore_query: true)
     expect(page).to have_content "Administrators"
     expect(page).to have_content "Details"
@@ -91,6 +92,8 @@ RSpec.describe "all_casa_admins/casa_orgs/casa_admins/new", type: :system do
     fill_in "Display name", with: "Freddy Valid"
     click_button "Submit"
     expect(page).to have_content "Email has already been taken"
+
+    expect(CasaAdmin.find_by(email: "valid@example.com").invitation_created_at).not_to be_nil
   end
 
   it "edits all casa admins" do
@@ -109,6 +112,7 @@ RSpec.describe "all_casa_admins/casa_orgs/casa_admins/new", type: :system do
     fill_in "all_casa_admin_email", with: "newemail@example.com"
     click_on "Update Profile"
     expect(page).to have_text "successfully updated"
+    expect(ActionMailer::Base.deliveries.last.body.encoded).to match(">Your CASA account's email has been updated to newemail@example.com")
 
     # change password
     click_on "Change Password"
@@ -122,5 +126,21 @@ RSpec.describe "all_casa_admins/casa_orgs/casa_admins/new", type: :system do
     fill_in "all_casa_admin_password_confirmation", with: "newpassword"
     click_on "Update Password"
     expect(page).to have_text "Password was successfully updated."
+    expect(ActionMailer::Base.deliveries.last.body.encoded).to match("Your CASA password has been changed.")
+  end
+
+  it "admin invitations expire" do
+    all_casa_admin = AllCasaAdmin.invite!(email: "valid@email.com")
+    raw_token = all_casa_admin.raw_invitation_token
+
+    # Invitation is valid within 1 week
+    travel 2.days
+    visit accept_all_casa_admin_invitation_path(invitation_token: raw_token)
+    expect(page).to have_text "Set my password"
+
+    # Invitation expires after 1 week
+    travel 8.days
+    visit accept_all_casa_admin_invitation_path(invitation_token: raw_token)
+    expect(page).to have_text "The invitation token provided is not valid!"
   end
 end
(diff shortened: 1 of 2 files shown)


spec/system/all_casa_admins/all_casa_admin_spec.rb
```diff
@@ -126,14 +122,5 @@
     fill_in "all_casa_admin_password_confirmation", with: "newpassword"
     click_on "Update Password"
     expect(page).to have_text "Password was successfully updated."
-    expect(ActionMailer::Base.deliveries.last.body.encoded).to match("Your CASA password has been changed.")
```
FireLemons, line 129: Can you revert this? Checking email activity is part of system tests but since capybara lacks tools to check email content, we have to check the backend. It's ok here because the email check is preceded by `expect(page).to have_text "Password was successfully updated."` which waits for the page to load.

spec/system/all_casa_admins/all_casa_admin_spec.rb
```diff
@@ -112,7 +109,6 @@
     fill_in "all_casa_admin_email", with: "newemail@example.com"
     click_on "Update Profile"
     expect(page).to have_text "successfully updated"
-    expect(ActionMailer::Base.deliveries.last.body.encoded).to match(">Your CASA account's email has been updated to newemail@example.com")
```
FireLemons, line 115: Please revert. See comment for line 129.

spec/system/all_casa_admins/all_casa_admin_spec.rb
```diff
@@ -92,8 +91,6 @@
     fill_in "Display name", with: "Freddy Valid"
     click_button "Submit"
     expect(page).to have_content "Email has already been taken"
-
-    expect(CasaAdmin.find_by(email: "valid@example.com").invitation_created_at).not_to be_nil
```
FireLemons, line 96: Please revert. See comment for line 129.

spec/system/all_casa_admins/all_casa_admin_spec.rb
```diff
@@ -52,8 +52,7 @@
     fill_in "Address", with: "123 Main St"
     click_on "Create CASA Organization"
     expect(page).to have_text "CASA Organization was successfully created."
-    organization = CasaOrg.find_by(name: "Cool Org Name")
```
FireLemons, line 55: Can you add an expect checking if the name of the new casa org is on the page?

spec/system/all_casa_admins/all_casa_admin_spec.rb
```diff
@@ -126,14 +122,5 @@
     fill_in "all_casa_admin_password_confirmation", with: "newpassword"
     click_on "Update Password"
     expect(page).to have_text "Password was successfully updated."
-    expect(ActionMailer::Base.deliveries.last.body.encoded).to match("Your CASA password has been changed.")
-  end
-
-  it "admin invitations expire" do
```
FireLemons, line 132: The only way to create an all casa admin is via direct db access. Can you convert this test into a system test by having the user click on the invitation links?

FireLemons (changes requested):

spec/system/all_casa_admins/all_casa_admin_spec.rb
vishaltps: Added expect(page).to have_text "Cool Org Name" after the org creation success message. This also
  replaces the previous CasaOrg.find_by lookup that was only used to build the exact path, now using a
  regex path match instead.

spec/system/all_casa_admins/all_casa_admin_spec.rb
vishaltps: Converted to a system test that visits the invitation acceptance link via accept_all_casa_admin_invitation_path. It verifies that the "Set my password" form loads within 1 week, and
  shows "The invitation token provided is not valid!" after expiry.

PR #6733, opened 2026-02-24
Race Condition Check: spec/system/court_dates/edit_spec.rb

### LINKED ISSUE #6329: Race Condition Check: spec/system/court_dates/edit_spec.rb
## Overview

Examples:
#### Good `expect`s
```ruby
  
  wait_for_download
```

#### Bad/Suspicious `expect`s
```ruby
  
```

## Things to do

stefannibrasil:
Closes https://github.com/rubyforgood/casa/issues/6329

- fixes one factory creation (it was creating an admin twice)
- removes unused context
- avoids checking the database for changes, and checks for what the user sees when interacting with the page

### DIFF
--- a/spec/system/court_dates/edit_spec.rb
+++ b/spec/system/court_dates/edit_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe "court_dates/edit", type: :system do
   let(:admin) { create(:casa_admin, casa_org: organization) }
   let(:volunteer) { create(:volunteer) }
   let(:supervisor) { create(:supervisor, casa_org: organization) }
-  let!(:casa_case) { create(:casa_case, case_number: 'CINA-08-1001', casa_org: organization) }
+  let!(:casa_case) { create(:casa_case, case_number: "CINA-08-1001", casa_org: organization) }
   let!(:past_court_date) { create(:court_date, :with_court_details, casa_case: casa_case, date: Date.new(2020, 12, 25)) }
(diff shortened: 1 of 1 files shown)


spec/system/court_dates/edit_spec.rb
```diff
@@ -111,11 +117,13 @@
 
       visit root_path
       click_on "Cases"
-      click_on casa_case.case_number
+      click_on "CINA-08-1001"
```
FireLemons: Can you DRY all these hard coded values? If multiple different but equal values are meant to refer to the same value, they should be referenced using the same variable.

FireLemons (changes requested):

spec/system/court_dates/edit_spec.rb
stefannibrasil: I can do that, sure. I agree with that for code outside of tests but not so much for tests. I am on the side that, especially for system tests, having explicit values like this is best for understanding the business logic of the app and debugging, when needed. By having explicit values, one does not need to evaluate the variable and jump straight to what the test is covering.

For example, checking for `court_date.date.strftime("%B %-d, %Y")` requires the developer to go and evaluate that when debugging a test. It's small but having the explicit date saves time, and keeps the test close to what the app renders.

Reverted it to your request in [39ee510](https://github.com/rubyforgood/casa/pull/6733/commits/39ee5106d121a25f98a7384b4a480cdedccd2bf0)

compwron (approved):

PR #6756, opened 2026-03-09
Remove race conditions from casa_org edit system tests

### LINKED ISSUE #6702: Race Condition Check: spec/system/casa_org/edit_spec.rb
## Overview

Examples:
#### Good `expect`s
```ruby
  
  wait_for_download
```

#### Bad/Suspicious `expect`s
```ruby
  
```

## Things to do

costajohnt:
## Issue

Closes #6702

## What changed and why

Three fixes in `spec/system/casa_org/edit_spec.rb`:

1. **Silent failure fix:** `has_no_checked_field?` and `has_checked_field?` are Capybara query methods that return booleans — they were never asserting anything. Replaced with proper `expect(page).not_to have_checked_field` / `expect(page).to have_checked_field`.

2. **Logic fix:** The factory default for `show_driving_reimbursement` is `true` (already checked). The original test called `check` on an already-checked field, which was a no-op. Changed the first action to `uncheck` so the test actually verifies toggling the flag off and on.

3. **Race condition fix:** Replaced `expect(organization.reload.logo).to be_attached` (DB query that can race against server processing) with `expect(page).to have_content("CASA organization was successfully updated.")` which uses Capybara's built-in waiting.

## How I tested

- `bundle exec rspec spec/system/casa_org/edit_spec.rb` — 5 examples, 0 failures

### DIFF
--- a/spec/system/casa_org/edit_spec.rb
+++ b/spec/system/casa_org/edit_spec.rb
@@ -8,13 +8,13 @@ RSpec.describe "casa_org/edit", type: :system do
     sign_in admin
     visit edit_casa_org_path(organization)
 
-    check "Show driving reimbursement"
+    uncheck "Show driving reimbursement"
     click_on "Submit"
-    has_no_checked_field? "Show driving reimbursement"
+    expect(page).not_to have_checked_field("Show driving reimbursement")
 
     check "Show driving reimbursement"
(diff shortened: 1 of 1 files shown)


compwron (approved):
Nice! :)

PR #6757, opened 2026-03-09
Race Condition Check: spec/system/case_court_reports/index_spec.rb

### LINKED ISSUE #6700: Race Condition Check: spec/system/case_court_reports/index_spec.rb
## Overview

Examples:
#### Good `expect`s
```ruby
  
  wait_for_download
```

#### Bad/Suspicious `expect`s
```ruby
  
```

## Things to do

costajohnt:
Resolves #6700

## Summary

- **spec/system/case_court_reports/index_spec.rb**: Replaced 4 database-level assertions (`attached?`, `reload`, `wait_for_report_attachment`) with page-based Capybara checks in the "generates and attaches a report on success" test. The button re-enabling, download icon visibility, and spinner state serve as reliable page-level signals that the report was generated successfully.
- **spec/support/case_court_report_helpers.rb**: Removed the `wait_for_report_attachment` helper method (only used in this one test). The two remaining helpers (`open_court_report_modal`, `open_case_select2_dropdown`) are unchanged.

All other tests in the file already used `expect(page)` exclusively — no changes needed.

## Test plan

- [x] `bundle exec rspec spec/system/case_court_reports/index_spec.rb` passes with 0 failures
- [x] No `reload`, `attached?`, or `wait_for_report_attachment` calls remain in the spec

### DIFF
--- a/spec/support/case_court_report_helpers.rb
+++ b/spec/support/case_court_report_helpers.rb
@@ -21,17 +21,4 @@ module CaseCourtReportHelpers
     # Wait for the dropdown to appear
     expect(page).to have_css(".select2-dropdown", visible: :visible)
   end
-
-  # Polls the database until the casa_case has an ActiveStorage court_report attached.
-  # This is used after clicking 'Generate Report' to wait for the background job to complete.
-  def wait_for_report_attachment(casa_case, timeout: 5)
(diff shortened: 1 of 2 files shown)


compwron (approved):
Nice

PR #6766, opened 2026-03-09
Race condition check: invite volunteers system spec

stefannibrasil:
Related https://github.com/rubyforgood/casa/issues/6698

- removes unused setup
- removes test that does not have any user interaction and is already covered by a unit test
- checks for what the user sees in the page when inviting volunteers, especially before checking the DB

### DIFF
--- a/spec/system/volunteers/invite_spec.rb
+++ b/spec/system/volunteers/invite_spec.rb
@@ -16,9 +16,9 @@ RSpec.describe "Inviting volunteers", type: :system do
       fill_in "Display name", with: "Jane Doe"
       fill_in "Date of birth", with: Date.new(1995, 5, 15)
 
-      expect {
-        click_on "Create Volunteer"
-      }.to change(Volunteer, :count).by(1)
+      click_on "Create Volunteer"
+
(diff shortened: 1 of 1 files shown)


app/views/devise/invitations/edit.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="password-box px-4 pb-3">
-  <h2 class="my-3">Send invitation</h2>
+  <h2 class="my-3">Set password</h2>
```
stefannibrasil: Let me know if you have other ideas. This looks more relevant in this context 👀

spec/system/volunteers/invite_spec.rb
```diff
@@ -208,15 +200,4 @@
       expect(page).to have_selector(".alert", text: "Sorry, you are not authorized to perform this action.")
     end
   end
-
-  describe "invitation expiration" do
```
stefannibrasil, line 212: This test is not an integration one because there is no UI interaction. It's also covered by this unit test already: https://github.com/rubyforgood/casa/blob/main/spec/models/volunteer_spec.rb#L388

compwron (approved):
we could argue about tests forever but I am ok with merging this

PR #6764, opened 2026-03-09
Replace date string with Date object

jyeharry:
### What github issue is this PR for, if any?

Resolves #6763

### What changed, and _why_?

Replaced a date string `"01/02/2020"` with a `Date` object to ensure consistent date rendering across timezones.

### How is this **tested**? (please write rspec and jest tests!) 💖💪

Updated flaky test.

### Screenshots please :)

N/A

PR #6785, opened 2026-03-18
Race Condition Check: spec/system/case_contacts/

### LINKED ISSUE #6701: Race Condition Check: spec/system/case_contacts/
Our system tests run on a mock browser. Sometimes Github's servers are strained so the browser performs very slowly. For some tests this causes a race condition between entering data using a the browser and checking the result of entering data directly from the database. The database check happens before the browser actions are completed if the browser's performance is crippled enough.

Examples:
#### Good `expect`s
```ruby
  
  wait_for_download
```

#### Bad/Suspicious `expect`s
```ruby
  
```

## Things to do

 - Check these files:
    - [ ] spec/system/case_contacts/additional_expenses_spec.rb
    - [ ] spec/system/case_contacts/contact_topic_answers_spec.rb
    - [ ] spec/system/case_contacts/drafts_spec.rb
    - [ ] spec/system/case_contacts/edit_spec.rb
    - [ ] spec/system/case_contacts/followups/create_spec.rb
    - [ ] spec/system/case_contacts/followups/resolve_spec.rb
    - [ ] spec/system/case_contacts/index_spec.rb
    - [ ] spec/system/case_contacts/new_spec.rb

stefannibrasil:

Resolves https://github.com/rubyforgood/casa/issues/6701

### What changed, and _why_?

When github's free CI servers are under heavy load, a race condition between the page loading and checking the database causes tests to flake. This is caused by a system test inputting data into a form then immediately checking the database without waiting for the form to finish submitting.

For every database check in the system files, this ensures it's preceded by a capybara matcher with automatic waiting, such as checking for assertions after the page loads before checking for DB values, or replaced the database
check with a check for something to appear on the page.

I also took the opportunity to refactor some small things to save up test setup time:

- build factories instead of creating as much as possible
- require `:js` only when necessary
- remove unused variables and factories
- keep necessary factories close to where they are used to avoid creating extra factories when not needed
- updates Date string to Date objects (these were flaky locally) - similar to https://github.com/rubyforgood/casa/pull/6764

### DIFF
--- a/spec/system/case_contacts/index_spec.rb
+++ b/spec/system/case_contacts/index_spec.rb
@@ -1,10 +1,10 @@
 require "rails_helper"
 
-RSpec.describe "case_contacts/index", :js, type: :system do
+RSpec.describe "case_contacts/index", type: :system do
   subject { visit case_contacts_path }
 
   let(:volunteer) { build(:volunteer, display_name: "Bob Loblaw", casa_org: organization) }
-  let(:organization) { create(:casa_org) }
+  let(:organization) { build(:casa_org) }
 
   before { sign_in volunteer }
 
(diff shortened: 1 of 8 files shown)


spec/system/case_contacts/index_spec.rb
```diff
@@ -14,19 +14,9 @@
     let!(:case_assignment) { create(:case_assignment, volunteer: volunteer, casa_case: casa_case) }
 
     context "without filter" do
-      let(:case_contacts) do
```
stefannibrasil, line 17: Instead of creating 4 factories for all scenarios, only create the necessary ones for each scenario.

compwron (approved):
nice :)

PR #6804, opened 2026-03-27
Address HTML markup issues reported by Herb

marcoroth:

 N/A

### What changed, and _why_?

This pull request addresses the HTML markup issues reported by the [Herb](https://github.com/marcoroth/herb) gem. I was testing my gem against real applications and thought I'd send a pull request addressing the markup issues I found in return 🙈 

The issues were mismatched opening/closing tags, block elements (`<h6>`, `<div>`) incorrectly nested inside `<p>` tags, ERB output tags (`<%= %>`) used in HTML attribute position, and a few typos like `</class>` instead of `</p>` in the public error pages. No logic changes, just structural HTML fixes to have the server render valid HTML.

### How is this **tested**? 

With the `herb` gem installed, you can run `herb analyze`. It now reports 240/240 files clean. All changes should preserve the same rendered output so existing tests should be unaffected, but let me know if you want me to add some specific tests.

I also saw you are runnig Better HTML and ERB Lint, so I didn't want to just add Herb. These markup fixes are also working without Herb so I left it out for now. But more than happy to add it as a dependency and set it up to run as part of CI, either in this PR or in a follow up 🙌🏼 

### Screenshots

CleanShot 2026-03-27 at 13 40 59@2x

### Feelings gif

giphy

### DIFF
--- a/app/views/emancipations/show.html.erb
+++ b/app/views/emancipations/show.html.erb
@@ -22,7 +22,7 @@
           type: "checkbox",
           class: "emancipation-category-check-box form-check-input",
           value: category.id,
-          disabled: emancipation_category_checkbox_checked(@current_case, category)) %>
+          checked: emancipation_category_checkbox_checked(@current_case, category)) %>
 
         <div class="pt-10"><%= category.name %></div>
       </div>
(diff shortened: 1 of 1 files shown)


app/views/devise/mailer/reset_password_instructions.html.erb
```diff
@@ -52,7 +52,7 @@
                       <tbody style="box-sizing: border-box;">
                         <tr style="box-sizing: border-box;">
                           <td class="cell c1776" style="box-sizing: border-box; width: 70%; vertical-align: middle;text-align:center;" width="70%" valign="middle">
-                            <img style="max-width: 100%;" src="https://user-images.githubusercontent.com/8918762/120879374-7c813a00-c588-11eb-92d7-efa6fdfdfbf0.png" alt="Logo" class="c926" style="box-sizing: border-box; color: rgb(158, 83, 129); font-size: 50px;padding-top:10px;background:#00447c;">
+                            <img src="https://user-images.githubusercontent.com/8918762/120879374-7c813a00-c588-11eb-92d7-efa6fdfdfbf0.png" alt="Logo" class="c926" style="max-width: 100%; box-sizing: border-box; color: rgb(158, 83, 129); font-size: 50px;padding-top:10px;background:#00447c;">
```
marcoroth, line 55: Previously, there were two `style` attributes. I now merged the first one into the second one. 

Technically this could have some side-effects as browsers (and email clients) usually ignore the second attribute with the same name.

PR #6826, opened 2026-04-05
Fix Jest test failures caused by Node 24 / jsdom incompatibilities

### LINKED ISSUE #6825: Fix Jest test failures caused by Node 24 / jsdom incompatibilities
Two Jest test suites are currently failing in CI due to incompatibilities with newer versions of Node and jsdom. The production code is correct in both cases — only the tests need updating. [See this CI run for an example.](https://github.com/rubyforgood/casa/actions/runs/24001622298/job/69998503863?pr=6823)

### `validated_form.test.js` (2 failures)

The tests assert that a CSS border value equals `'2px solid #ffc107'`. Newer versions of jsdom normalize CSS color values to `rgb()` notation when reading them back via jQuery's `.css()`, so the actual value returned is `'2px solid rgb(255, 193, 7)'`. The fix is to update the two assertions to match the normalized format.

### `two_minute_warning_session_timeout.test.js` (1 failure)

The test mocks `window.location` using `Object.defineProperty(window, 'location', { writable: true, value: ... })`. Newer versions of jsdom define `location` as non-configurable, causing this to throw `TypeError: Cannot redefine property: location`. The fix is to replace that pattern with `delete window.location` followed by a plain assignment, which is the modern jsdom-compatible approach.

### Steps to reproduce

```shell
npx jest app/javascript/__tests__/validated_form.test.js app/javascript/__tests__/two_minute_warning_session_timeout.test.js
```

Both suites fail on `main` with Node 24.

cliftonmcintosh:
Resolves #6825

### What changed, and _why_?

Updated two Jest test files to be compatible with Node 24 and newer versions of jsdom. No production code was changed.

**`app/javascript/__tests__/two_minute_warning_session_timeout.test.js`**

Replaced `Object.defineProperty(window, 'location', { writable: true, value: ... })` with `delete window.location` followed by a plain assignment. Newer versions of jsdom define `location` as non-configurable, causing the previous approach to throw `TypeError: Cannot redefine property: location`.

**`app/javascript/__tests__/validated_form.test.js`**

Updated assertions comparing CSS border values from `'2px solid #ffc107'` to `'2px solid rgb(255, 193, 7)'`. Newer versions of jsdom normalize CSS color values to `rgb()` notation when reading them back via jQuery's `.css()`.

The two previously failing test suites now pass:

```shell
npx jest app/javascript/__tests__/validated_form.test.js app/javascript/__tests__/two_minute_warning_session_timeout.test.js

### DIFF
--- a/app/javascript/__tests__/two_minute_warning_session_timeout.test.js
+++ b/app/javascript/__tests__/two_minute_warning_session_timeout.test.js
@@ -13,10 +13,8 @@ describe('session_timeout_poller', () => {
     mockAlert = jest.fn()
     mockReload = jest.fn()
     global.alert = mockAlert
-    Object.defineProperty(window, 'location', {
-      writable: true,
-      value: { reload: mockReload }
-    })
+    delete window.location
+    window.location = { reload: mockReload }
 
     // Set a short timeout for testing (in minutes)
     originalTimeout = global.window.timeout
--- a/app/javascript/__tests__/validated_form.test.js
+++ b/app/javascript/__tests__/validated_form.test.js
@@ -146,12 +146,12 @@ describe('NonDrivingContactMediumWarning', () => {
       $(() => {
         try {
           expect(checkboxContainer.css('background-color')).not.toBe('rgb(255, 248, 225)')
-          expect(milesDrivenInput.css('border')).not.toBe('2px solid #ffc107')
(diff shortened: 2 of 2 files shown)


compwron (approved):

compwron:
:)

PR #6888, opened 2026-04-25
fix: scope ContactTypeGroup lookup to current organization to prevent cross-org access

PhutiCee:
## Summary
Fixes #6351

## Problem
An admin from Organization A could visit a direct URL containing the ID
of a ContactTypeGroup belonging to Organization B and successfully edit
or update it. This is a multi-tenancy data isolation bug.

## Root Cause
`set_contact_type_group` used `ContactTypeGroup.find(params[:id])`
which searches all records globally, ignoring the current user's org.

## Fix
Scoped the lookup to the current organization:

```ruby
# Before
@contact_type_group = ContactTypeGroup.find(params[:id])

# After
@contact_type_group = current_organization.contact_type_groups.find(params[:id])
```

Also added `rescue_from ActiveRecord::RecordNotFound` in
ApplicationController to return a proper 404 response instead of
raising an unhandled exception.

## Testing
- Added cross-org request spec for GET edit (expect 404)
- Added cross-org request spec for PUT update (expect 404, data unchanged)
- All 14 examples pass, 0 failures

## Notes
As @FireLemons noted, similar patterns may exist in HearingTypes and
Judge controllers and would be worth a follow-up fix.

PR #6894, opened 2026-04-29
refactor: centralize controller rescues, enable index policy scoping, fix Docker mount perms

compwron:
N/A

### What changed, and _why_?

**`refactor: centralize controller rescues and enable index policy scoping`**

- Move `rescue_from ActiveRecord::RecordNotFound` from `ContactTypeGroupsController` up to `ApplicationController` so every controller renders a consistent 404 (HTML or JSON) without each controller needing its own handler.
- Drop the manual `rescue_from StandardError` + `KNOWN_ERRORS` + `log_and_reraise` trio. Bugsnag's Rack middleware already auto-reports unhandled exceptions, so the manual `notify`-then-`raise` was double-reporting (and `KNOWN_ERRORS` was effectively dead code, since more specific `rescue_from` handlers always matched first).
- Enable `after_action :verify_policy_scoped, only: :index` to surface multi-tenancy regressions like the recent `ContactTypeGroup` cross-org bug (#6888). Add `skip_after_action :verify_policy_scoped` with a `# TODO` comment to controllers whose index does not currently call `policy_scope` (CSV reports, public endpoints, admin-scope controllers, `OtherDutiesController`, etc.) so behavior is preserved while each gap is explicit and trackable.
- Update specs that asserted `raise_error(ActiveRecord::RecordNotFound)` to assert a 404 response instead, since the error is now rescued centrally rather than bubbling up.

- Added two new examples in `spec/controllers/application_controller_spec.rb` covering the `record_not_found` HTML and JSON paths.
- Updated 5 existing specs (`bulk_court_dates`, `case_contacts/followups` x2, `users`, `emancipations_controller`) that asserted `raise_error(ActiveRecord::RecordNotFound)` to assert a 404 response.
- Full `spec/requests` + `spec/controllers` suite passes locally: **996 examples, 0 failures**.

### Screenshots please :)

N/A — no UI changes.

### Feelings gif (optional)

🧹

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -33,7 +33,10 @@ jobs:
           mkdir -p tmp/downloads
           chmod 777 tmp tmp/downloads
       - name: docker UP
-        run: docker compose up -d
+        run: |
+          docker compose up -d database selenium_chrome
+          docker compose run --rm --no-deps --user 0:0 --entrypoint sh web -c "chown -R app:app /usr/src/app /usr/local/bundle/gems"
+          docker compose up -d web
       - name: db:setup
         run: docker compose exec -T web rails db:setup
       - name: compile assets
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -33,10 +33,7 @@ jobs:
           mkdir -p tmp/downloads
           chmod 777 tmp tmp/downloads
       - name: docker UP
-        run: |
-          docker compose up -d database selenium_chrome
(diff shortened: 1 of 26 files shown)


compwron:
@copilot fix: standard: Use Ruby Standard Style (https://github.com/standardrb/standard)
  spec/controllers/application_controller_spec.rb:99:35: Rails/RootPathnameMethods: `Rails.public_path` is a `Pathname`, so you can use `Rails.public_path.join("404.html").read`.

PR #6942, opened 2026-05-08
feat(#6502): Filter for new case contact table

### LINKED ISSUE #6502: New Case Contact Table: implement "filter" button/functionality
 - volunteers

## Description
This ticket is a sub-section of a bigger ticket: https://github.com/rubyforgood/casa/issues/6462.
Some foundational work for the bigger ticket has been started in this PR: https://github.com/rubyforgood/casa/pull/6467.

This ticket is to add the "filter" button and its functionality to actually filter the table by its constraints. The existing version of the case contacts table (old design) includes the logic for filtering, so it should be a case of refactoring so that it works on this new design.

Image

## How to access the QA site
_Login Details:_  

Login Emails: 

### Questions? Join Slack!

We highly recommend that you join us in [slack](https:https://join.slack.com/t/rubyforgood/shared_invite/zt-35218k86r-vlIiWqig54c9t~_LkGpQ7Q) #casa channel to ask questions quickly. And [discord](https://discord.gg/qJcw2RZH8Q) for office hours (currently Tuesday 5-7pm Pacific), stakeholder news, and upcoming new issues.

cliftonmcintosh:
Resolves #6502

### What changed, and _why_?

Adds a filter button and collapsible filter panel to the new case contacts table (`/case_contacts/new_design`).

**Backend**
- `app/datatables/case_contact_datatable.rb`

Added a new `apply_additional_filters` method that pipes the datatable query through existing `CaseContact` model scopes before returning results. Each scope already handles nil/blank values gracefully, so only active filters affect the query. Filters combine with AND logic — a record must satisfy every active filter simultaneously.

Filters supported:
- Date range (occurred starting at / ending at)
- Case (one or more case IDs)
- Relationship / contact type (one or more contact type IDs)
- Medium (single-select dropdown)
- Contacted / contact made (single-select dropdown)
- Hide drafts (checkbox)

**Frontend** 
- `app/views/case_contacts/case_contacts_new_design/index.html.erb` 
- `app/javascript/src/dashboard.js`

Restructured the page header into a filter toolbar row with the Filter button on the left and New Case Contact on the right. 
- Clicking Filter toggles a collapsible panel containing all filter inputs.
- Apply Filters triggers a DataTables reload with the active filters attached to the request
- Reset clears all inputs and reloads.

The filter values are collected by a `collectCaseContactFilters()` function and attached to every DataTables POST request via the `ajax.data` hook, so filters remain in effect when the user sorts or pages through results.

**Request specs** (`spec/requests/case_contacts/case_contacts_new_design_spec.rb`)

Nine tests covering each filter param individually and in combination:
- `occurred_starting_at` — returns only contacts on or after the date
- `occurred_ending_at` — returns only contacts on or before the date
- `casa_case_ids` — returns only contacts for the specified cases
- `contact_type_ids` — returns only contacts with the specified contact types
- `contact_medium` — returns only contacts with the specified medium
- `contact_made: true` — returns only contacts where contact was made
- `contact_made: false` — returns only contacts where contact was not made
- `no_drafts` — excludes draft contacts
- Multiple filters combined — all active filters apply together (AND)

**System specs** (`spec/system/case_contacts/case_contacts_new_design_spec.rb`)

Seven tests covering the filter UI:
- Filter button is visible on the page
- Filter panel is hidden by default
- Clicking Filter opens the panel; clicking again closes it
- Applying a medium filter shows only contacts with that medium
- Applying a date range filter shows only contacts within the range
- Checking Hide Drafts excludes draft rows
- Reset clears all filters and shows all contacts again

### Screen recordings
<details>
<summary>Opening and closing the filter panel</summary>

The Filter button lives in the toolbar row to the left of the New Case Contact button. Clicking it toggles a collapsible panel containing all filter inputs. The panel is hidden by default and does not affect the table until a filter is applied.

https://github.com/user-attachments/assets/f42c5270-c4f2-4cd6-b360-1d584c888726

</details>

<details>
<summary>Applying a single filter</summary>

Selecting a value and clicking Apply Filters triggers a DataTables reload with the filter attached to the server request. Only rows matching the filter are returned; the table updates in place without a full page refresh.

https://github.com/user-attachments/assets/171aa1f5-b383-4425-9df4-cd45c6c4db98

</details>

<details>
<summary>Combining multiple filters and resetting</summary>

All active filters are applied together with AND logic — a contact must satisfy every active filter to appear in the results. Clicking Reset clears all inputs and reloads the table with no filters applied, restoring the full result set.

https://github.com/user-attachments/assets/de5e4e24-5d78-43b2-aede-ecce85a3b81b

</details>

<details>
<summary>Mobile view</summary>

The filter toolbar and panel are built with Bootstrap's responsive grid. At phone-sized widths:

- The toolbar buttons wrap to separate lines rather than overflowing
- Each filter section (date inputs, case checkboxes, relationship checkboxes, dropdowns) stacks to full width
- All inputs remain usable without horizontal scrolling

https://github.com/user-attachments/assets/715ab8ae-ae03-47e6-a829-17d1aaed3e12

</details>

### Differences from the Figma design

The [Figma mockup](https://www.figma.com/design/fVC7VShMu4P2FK6FexQKXI/Case-Contact-Table?node-id=1818-12386&t=fMwnPMXlmY22xNsr-1) informed this implementation but was not followed exactly. Questions were raised with the designer via [this Slack thread](https://rubyforgood.slack.com/archives/CVB0QJGVD/p1776431126093069) but went unanswered. A fellow developer advised:

> Since we haven't had any feedback from them in a couple weeks, I think you can make your best guess as to what should be done. We can always improve it later.

The deviations are intentional and documented below.

#### Flat panel instead of drill-down overlay

The Figma shows a two-level drill-down overlay: clicking Filter opens a category list, and clicking a category slides into a sub-panel with that filter's options. A single flat panel was implemented instead, showing all filters at once. The flat panel is simpler to build, simpler to test, and easier to reason about. The backend `additional_filters` param contract is identical regardless of UI shape, so replacing the flat panel with the drill-down overlay in a future ticket is a pure frontend change with no backend work required.

#### Medium, Contacted, and Draft — single-select instead of multi-select

The Figma shows these three as checkbox-based filters supporting multiple simultaneous selections. All three were implemented as single-select dropdowns instead. The existing model scopes for these filters accept a single value; supporting multi-select would require updating the scopes and the frontend. The single-select versions deliver useful functionality now, and the upgrade path is straightforward when prioritised.

#### Created By and Topics filters omitted

The Figma includes Created By and Topics filters. Both were skipped. Created By has an existing backend scope but raises an open question about which users to list in the UI. Topics has no backend scope at all. Both are self-contained additions that can be done in follow-up tickets without changing anything built here.

#### Want Driving Reimbursement filter dropped

The old design included a Want Driving Reimbursement filter. The Figma mockup for the new design does not include it, so it was omitted.

### Feelings gif (optional)

filtering like a pro

### DIFF
--- a/app/datatables/case_contact_datatable.rb
+++ b/app/datatables/case_contact_datatable.rb
@@ -55,7 +55,18 @@ class CaseContactDatatable < ApplicationDatatable
   end
 
   def filtered_records
-    raw_records.where(search_filter)
+    apply_additional_filters(raw_records.where(search_filter))
+  end
+
+  def apply_additional_filters(records)
+    records = records.occurred_starting_at(additional_filters[:occurred_starting_at])
+    records = records.occurred_ending_at(additional_filters[:occurred_ending_at])
+    records = records.with_casa_case(Array(additional_filters[:casa_case_ids])) if additional_filters[:casa_case_ids].present?
+    records = records.contact_type(Array(additional_filters[:contact_type_ids])) if additional_filters[:contact_type_ids].present?
+    records = records.contact_medium(additional_filters[:contact_medium])
+    records = records.contact_made(additional_filters[:contact_made])
+    records = records.no_drafts(additional_filters[:no_drafts].to_i) if additional_filters[:no_drafts].present?
+    records
   end
 
   def raw_records
--- a/app/javascript/src/dashboard.js
+++ b/app/javascript/src/dashboard.js
@@ -44,6 +44,13 @@ const defineCaseContactsTable = function () {
     ajax: {
       url: $('table#case_contacts').data('source'),
       type: 'POST',
+      data: function (d) {
+        const filters = collectCaseContactFilters()
+        if (Object.keys(filters).length > 0) {
+          d.additional_filters = filters
+        }
+        return d
+      },
       error: function (xhr, error, code) {
         console.error('DataTable error:', error, code)
       },
@@ -248,6 +255,43 @@ const defineCaseContactsTable = function () {
       success: () => table.ajax.reload(null, false)
     })
   })
+
+  $('#cc-filter-toggle').on('click', function () {
+    $('#cc-filter-panel').toggle()
+  })
+
+  $('#cc-filter-apply').on('click', function () {
+    table.ajax.reload(null, false)
+    $('#cc-filter-panel').hide()
+  })
+
+  $('#cc-filter-reset').on('click', function () {
+    $('#cc-filter-occurred-starting-at, #cc-filter-occurred-ending-at').val('')
+    $('#cc-filter-medium, #cc-filter-contact-made').val('')
+    $('.cc-filter-casa-case, .cc-filter-contact-type, #cc-filter-no-drafts').prop('checked', false)
+    table.ajax.reload(null, false)
+  })
+}
+
+function collectCaseContactFilters () {
+  const filters = {}
+  const startDate = $('#cc-filter-occurred-starting-at').val()
+  const endDate = $('#cc-filter-occurred-ending-at').val()
+  const casaIds = $('.cc-filter-casa-case:checked').map((_, el) => el.value).get()
+  const contactTypeIds = $('.cc-filter-contact-type:checked').map((_, el) => el.value).get()
+  const medium = $('#cc-filter-medium').val()
+  const contactMade = $('#cc-filter-contact-made').val()
+  const noDrafts = $('#cc-filter-no-drafts').is(':checked')
+
+  if (startDate) filters.occurred_starting_at = startDate
+  if (endDate) filters.occurred_ending_at = endDate
+  if (casaIds.length > 0) filters.casa_case_ids = casaIds
+  if (contactTypeIds.length > 0) filters.contact_type_ids = contactTypeIds
+  if (medium) filters.contact_medium = medium
+  if (contactMade !== '') filters.contact_made = contactMade
+  if (noDrafts) filters.no_drafts = '1'
+
+  return filters
 }
 
 $(() => { // JQuery's callback for the DOM loading
--- a/app/views/case_contacts/case_contacts_new_design/index.html.erb
+++ b/app/views/case_contacts/case_contacts_new_design/index.html.erb
@@ -19,11 +19,11 @@
         <div class="col-12"><h5>Date of contact</h5></div>
         <div class="col-sm-6 input-style-1">
           <label for="cc-filter-occurred-starting-at">Starting from</label>
-          <input type="date" id="cc-filter-occurred-starting-at" class="form-control">
+          <input type="date" id="cc-filter-occurred-starting-at" class="form-control" autocomplete="off">
         </div>
         <div class="col-sm-6 input-style-1">
           <label for="cc-filter-occurred-ending-at">Ending at</label>
-          <input type="date" id="cc-filter-occurred-ending-at" class="form-control">
+          <input type="date" id="cc-filter-occurred-ending-at" class="form-control" autocomplete="off">
         </div>
       </div>
 
--- a/app/views/case_contacts/case_contacts_new_design/index.html.erb
+++ b/app/views/case_contacts/case_contacts_new_design/index.html.erb
@@ -1,10 +1,104 @@
 <div class="title-wrapper pt-30">
-  <div class="d-flex justify-content-between align-items-center flex-wrap gap-2 mb-30">
-    <h1>Case Contacts</h1>
+  <h1 class="mb-20">Case Contacts</h1>
+
+  <div class="d-flex justify-content-between align-items-center flex-wrap gap-2 mb-3" id="cc-filter-toolbar">
+    <button type="button" id="cc-filter-toggle" class="main-btn secondary-btn btn-sm btn-hover">
+      <i class="lni lni-funnel mr-5" aria-hidden="true"></i>
+      Filter
+    </button>
+
     <%= link_to new_case_contact_path, class: "main-btn primary-btn btn-sm btn-hover" do %>
       <i class="lni lni-plus mr-10" aria-hidden="true"></i>
       New Case Contact
     <% end %>
+  </div>
+
+  <div id="cc-filter-panel" class="card-style mb-3" style="display: none;">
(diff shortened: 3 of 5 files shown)


app/datatables/case_contact_datatable.rb
```diff
@@ -55,7 +55,18 @@ def data
...
+  def apply_additional_filters(records)
+    records = records.occurred_starting_at(additional_filters[:occurred_starting_at])
+    records = records.occurred_ending_at(additional_filters[:occurred_ending_at])
+    records = records.with_casa_case(Array(additional_filters[:casa_case_ids])) if additional_filters[:casa_case_ids].present?
+    records = records.contact_type(Array(additional_filters[:contact_type_ids])) if additional_filters[:contact_type_ids].present?
+    records = records.contact_medium(additional_filters[:contact_medium])
+    records = records.contact_made(additional_filters[:contact_made])
+    records = records.no_drafts(additional_filters[:no_drafts].to_i) if additional_filters[:no_drafts].present?
+    records
   end
```
cliftonmcintosh, line 70: A new method rather than extending `search_filter` because the two are fundamentally different in structure. `search_filter` builds a raw SQL `WHERE` clause that ORs conditions together against a single free-text string — merging AND-logic filter conditions into it would break both features. `apply_additional_filters` chains named ActiveRecord scopes instead, each adding an ANDed `WHERE` condition, which is exactly the semantics filters need.

app/views/case_contacts/case_contacts_new_design/index.html.erb
```diff
@@ -1,11 +1,105 @@
...
+      Filter
+    </button>
+
     <%= link_to new_case_contact_path, class: "main-btn primary-btn btn-sm btn-hover" do %>
       <i class="lni lni-plus mr-10" aria-hidden="true"></i>
       New Case Contact
     <% end %>
   </div>
+
+  <div id="cc-filter-panel" class="card-style mb-3" style="display: none;">
```
cliftonmcintosh, line 16: Inline style rather than a CSS class is intentional here. The inline style takes effect immediately as the browser parses the HTML, before any JavaScript loads. A CSS class would leave a window where the panel is briefly visible while the JS bundle is still loading, causing a flash of unwanted content on page load.

app/javascript/src/dashboard.js
```diff
@@ -248,6 +255,43 @@ const defineCaseContactsTable = function () {
...
+  const medium = $('#cc-filter-medium').val()
+  const contactMade = $('#cc-filter-contact-made').val()
+  const noDrafts = $('#cc-filter-no-drafts').is(':checked')
+
+  if (startDate) filters.occurred_starting_at = startDate
+  if (endDate) filters.occurred_ending_at = endDate
+  if (casaIds.length > 0) filters.casa_case_ids = casaIds
+  if (contactTypeIds.length > 0) filters.contact_type_ids = contactTypeIds
+  if (medium) filters.contact_medium = medium
+  if (contactMade !== '') filters.contact_made = contactMade
```
cliftonmcintosh, line 291: `!== ''` rather than a plain truthy check is intentional. The "Not Reached" option has the value `"false"`, which is a non-empty string but would fail a truthy check, silently dropping the filter. The empty string `""` is the only value that means "no filter selected" (the default "Display all" option), so checking against that specifically is a deliberate choice.

spec/system/case_contacts/case_contacts_new_design_spec.rb
```diff
@@ -25,6 +25,94 @@
...
+      expect(page).to have_text(in_person_date)
+      expect(page).not_to have_text(video_date)
+    end
+
+    it "filters by date range" do
+      old_date = I18n.l(in_person_contact.occurred_at, format: :full)
+      recent_date = I18n.l(video_contact.occurred_at, format: :full)
+
+      click_button "Filter"
+      execute_script("document.getElementById('cc-filter-occurred-ending-at').value = '#{4.days.ago.to_date}'")
```
cliftonmcintosh, line 82: `execute_script` is used here instead of Capybara's `fill_in` or `find.set` because those can be unreliable for `type="date"` inputs in headless Chrome — the browser's native date picker intercepts the interaction and jQuery does not see the resulting value. Setting the value directly on the DOM element via JavaScript bypasses the picker and gives jQuery a value it can read with `.val()` when the filter is applied.

app/views/case_contacts/case_contacts_new_design/index.html.erb
```diff
@@ -1,11 +1,105 @@
 <div class="title-wrapper pt-30">
-  <div class="d-flex justify-content-between align-items-center flex-wrap gap-2 mb-30">
-    <h1>Case Contacts</h1>
+  <h1 class="mb-20">Case Contacts</h1>
+
+  <div class="d-flex justify-content-between align-items-center flex-wrap gap-2 mb-3" id="cc-filter-toolbar">
+    <button type="button" id="cc-filter-toggle" class="main-btn secondary-btn btn-sm btn-hover">
+      <i class="lni lni-funnel mr-5" aria-hidden="true"></i>
+      Filter
+    </button>
```
cliftonmcintosh, line 8: Should the color of this button be the same as the color of the New Case Contact button? The Figma has it as white with a dark blue icon and dark blue text.

As coded:

Image


In the Figma:

Image

Here's my rationale for the current color:

"New Case Contact" is the primary action on the page. The Filter button is a utility action, used less frequently and by a subset of users. Giving the primary action a filled, prominent style and the secondary action a less prominent style guides the eye to the most important action without hiding the secondary one.

If they were the same colour, the two buttons would compete for attention equally, which would be visually noisy and would undermine the distinction between "the thing you're here to do" and "a tool to help you find things."

cliftonmcintosh: Here is what it would look like with the `dark-btn-outline` version. I think this makes it hard to see until hover.


https://github.com/user-attachments/assets/8e1e5374-72f5-4153-835e-c3a0549186b2

compwron: I think that your choice is reasonable :)

compwron (approved):

PR #6955, opened 2026-05-20
Upgrade Rails from 7.2 to 8.0

stefannibrasil:
### What github issue is this PR for, if any?

Closes https://github.com/rubyforgood/casa/issues/6925

### What changed, and _why_?

Rails 8.0 was released in late 2024 and brings Solid Queue / Solid Cache / Solid Cable as defaults, Propshaft, Kamal 2, Authentication generator, and a number of framework deprecations.

Most notable changes:
- It's not possible to use `only` and `except` for non-RESTful routes: https://github.com/rubyforgood/casa/pull/6955/changes/c64228a2db65bbb005c32af2665b9af7a1dcaa63
- I removed comments and configs from previous upgrades, and updated `config/environments` with the generated files as much as possible
- I couldn't find out exactly where this changed in the release notes for this Rails version but I imagine it's a good change since it shows the new 404 page instead of raising a server error: https://github.com/rubyforgood/casa/pull/6955/changes/c5af9329fceb779de78cf99b59cae383b5eff0cc
- The new defaults from `config/initializers/new_framework_defaults_8_0.rb` look okay to be enabled altogether but if you prefer to enable them individually, let me know.
- The 406 and 422 error pages have a new generated UI:
Screenshot 2026-05-15 at 12 11 29 PM

Screenshot 2026-05-15 at 12 07 35 PM

### How is this **tested**? (please write rspec and jest tests!) 💖💪

- tested the app locally
- running tests
- CI is passing

### Followups

- Once Rails 8 is up and running, we can later update the `config.load_defaults` to 8.0. It should fix this deprecation warning: "DEPRECATION WARNING: `to_time` will always preserve the full timezone rather than offset of the receiver in Rails 8.1. To opt in to the new behavior, set `config.active_support.to_time_preserves_timezone = :zone`. (called from <top (required)> at /home/runner/work/casa/casa/config/environment.rb:5)"
- It's possible to upgrade from Sprockets to Propshaft: https://github.com/rails/propshaft/blob/main/UPGRADING.md
- upgrade Rails from 8.0 to 8.1 (should be smooth, no breaking changes)

### Questions

Is it possible to test this on staging/QA first? Please try this out locally and let me know if you run into any issues.

PR #6962, opened 2026-05-25
Migrate annotate to annotaterb

stefannibrasil:

Related to https://github.com/rubyforgood/casa/pull/6955

### What changed, and _why_?

`annotate` is not compatible with Rails > 8. There is a new fork gem that is actively maintained: `annotaterb`. Migrating to the fork allows us to upgrade Rails without regressions.

I updated the generated config file with the previous configuration from `annotate`. After running `bundle exec annotaterb models`, there are no file changes.

The difference from now on is that the commands to generate annotations have changed: `annotaterb` provides a CLI with updated options. For example, previously you would use `bundle exec annotate`, now one would use `bundle exec annotaterb models`. Learn more in https://github.com/drwl/annotaterb/blob/main/MIGRATION_GUIDE.md#command-line-differences


compwron (approved):
nice :)

Issue #6912, opened 2026-05-03
Expand Prosopite N+1 enforcement: 12 newly-surfaced N+1s across 6 directories


compwron:
## Context

Prosopite is wired to raise on N+1 in tests (\`spec/support/prosopite.rb\`), but \`spec/.prosopite_ignore\` currently opts out **every** spec directory — so the gem catches nothing in CI. \`PROSOPITE_TODO.md\` tracks known N+1s but is incomplete.

I tried removing the six \"unflagged\" directories from \`.prosopite_ignore\` (\`helpers\`, \`notifications\`, \`seeds\`, \`mailers\`, \`presenters\`, \`config\`) and running those specs. **12 failures surfaced** across all six — meaning none of them are clean today.

### \`spec/presenters\` — 4 failures
- \`spec/presenters/case_contact_presenter_spec.rb:18\`, \`:25\`, \`:33\`, \`:40\` — \`CaseContactPresenter#display_case_number\`. Every call iterates and hits the DB. Likely fix: preload at the controller level or change the lookup data structure.
- (Note: this is also why issue #6909 — retire presenters in favor of Draper — is worth doing; the rewrite can fix this in passing.)

### \`spec/helpers\` — 2 failures
- \`spec/helpers/ui_helper_spec.rb:12\` — \`UiHelper#grouped_options_for_assigning_case\` doesn't eager-load.
- \`spec/helpers/notifications_helper_spec.rb:56\` — \`NotificationsHelper#patch_notes_as_hash_keyed_by_type_name\`.

### \`spec/mailers\` — 2 failures
- \`spec/mailers/supervisor_mailer_spec.rb:122\`, \`:134\` — \`SupervisorMailer.weekly_digest\` N+1s when iterating volunteers' \`case_contact\`s.

### \`spec/notifications\` — 1 failure
- \`spec/notifications/reimbursement_complete_notifier_spec.rb:17\` — \`ReimbursementCompleteNotifier#message\` queries mileage rate per call.

### \`spec/seeds\` — 1 failure
- \`spec/seeds/seeds_spec.rb:33\` — seeds run once; this may be acceptable to wrap in \`Prosopite.pause\` rather than fix.

### \`spec/config\` — 2 failures
[...]

PR #6988, opened 2026-06-04
Prosopite /helpers fixes

stefannibrasil:

Related to https://github.com/rubyforgood/casa/issues/6912

### What changed, and _why_?

`Prosopite` is wired to raise on N+1 in tests (`spec/support/prosopite.rb`), but `spec/.prosopite_ignore` currently opts out every spec directory — so the gem catches nothing in CI. `PROSOPITE_TODO.md` tracks known N+1s but is incomplete.

This PR is focused on fixing the /spec/helpers Prosopite errors. Please see each individual commit for more details on each file.

Now that these directories are enforced, no future PR can silently introduce a N+1s in them.

Note: I re-organized the test so it didn't use an instance variable as per Rubocop's recommendation. Then, I changed the UiHelper to use the instance given as the param instead of an instance variable. I tested the Volunteers edit page where this is used and it's working as before, but let me know if I am missing something.

### DIFF
--- a/spec/helpers/ui_helper_spec.rb
+++ b/spec/helpers/ui_helper_spec.rb
@@ -3,8 +3,8 @@ require "rails_helper"
 RSpec.describe UiHelper, type: :helper do
   describe "#grouped_options_for_assigning_case" do
     before do
-      @casa_cases = create_list(:casa_case, 4)
-      @volunteer = create(:volunteer, casa_org: @casa_cases[0].casa_org)
+      @casa_case = create(:casa_case)
+      @volunteer = create(:volunteer, casa_org: @casa_case.casa_org)
       current_user = create(:supervisor)
       allow(helper).to receive(:current_user).and_return(current_user)
     end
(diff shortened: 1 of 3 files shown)


spec/helpers/ui_helper_spec.rb
```diff
@@ -3,8 +3,8 @@
 RSpec.describe UiHelper, type: :helper do
   describe "#grouped_options_for_assigning_case" do
     before do
-      @casa_cases = create_list(:casa_case, 4)
```
compwron, line 6: Why are we having fewer?

compwron (approved):

spec/helpers/ui_helper_spec.rb
stefannibrasil: Proposite was complaining here... it looks like proposite raises lots of errors like these in tests. I thought of skipping proposite here but upon review, I noticed that the test did not needed to create that many records, which led me to create fewer records.

PR #6990, opened 2026-06-09
fix: repair broken Heroku Scheduler rake tasks

### LINKED ISSUE #6923: Fix broken Heroku Scheduler rake tasks
## Context

Running each project rake task in this repo (`bundle exec rake <name>`) surfaces a cluster of bugs in the tasks driven by Heroku Scheduler. **Three tasks crash on every invocation; two more crash latently when production data shape triggers them.**

This issue covers the "broken in production" bugs only. Maintainability cleanup (auto_annotate noise, `factory_bot:lint` debt, `deployment/` pruning, in-task schedule-guard anti-pattern) and moving controller specs to request specs are tracked separately.

### 1. `send_case_contact_types_reminder` — `NoMethodError: undefined method 'every' for main`

`lib/tasks/send_case_contact_types_reminder.rake:4` and `lib/tasks/send_no_contact_made_reminder.rake:4` both wrap their work in:

```ruby
every 1.weeks do
  CaseContactTypesReminder.new.send!
end
```

`every` is from the `whenever` gem (intended for `config/schedule.rb`), not a Rake primitive. The block raises immediately when the task runs — meaning the case-contact-types SMS reminder has not actually executed since that block was added.

### 2. `send_no_contact_made_reminder` — task name doesn't exist

`lib/tasks/send_no_contact_made_reminder.rake` declares `task send_case_contact_types_reminder: :environment do …` — the same task name used by the other file. `rake -T` confirms only one task is registered, with both descriptions concatenated:

```
rake send_case_contact_types_reminder    # Send an SMS … past 60 or more days / Send an SMS … two weeks have passed …
```

So Heroku Scheduler invocations of `send_no_contact_made_reminder` fail with "Don't know how to build task", and the no-contact-made SMS reminder has never run. The file also `require_relative`s the wrong service file and instantiates `CaseContactTypesReminder` instead of `NoContactMadeReminder`.

### 3. `recurring:init` — `NameError: uninitialized constant ExampleRecurringTask`

`lib/tasks/recurring_jobs.rake:3` calls `ExampleRecurringTask.schedule!`, but `ExampleRecurringTask` is defined in `lib/tasks/example_recurring_task.rb` which Rails does not autoload. The class is also dead example code:

```ruby
def perform
  Bugsnag.notify("This is just ExampleRecurringTask saying hi in #{Rails.env}")
end
```

— likely safe to delete entirely.

### 4. `youth_birthday_reminder` — `NoMethodError: undefined method 'volunteer_id' for nil`

`lib/tasks/youth_birthday_reminder.rake:6`:

```ruby
.deliver(Volunteer.find_by(id: casa_case.case_assignments.first.volunteer_id))
```

When a `CasaCase.birthday_next_month` case has no `case_assignments`, `.first` is `nil`. Confirmed via isolated repro against dev DB:

```
$ rails runner 'cc = CasaCase.birthday_next_month.first; cc.case_assignments.destroy_all; cc.case_assignments.first.volunteer_id'
RAISED: NoMethodError: undefined method 'volunteer_id' for nil
```

Seed data happens to have ≥1 assignment per case, so the bug never fires in test/dev — but it will fire in production whenever a case is briefly without an active assignment.

`.first` also ignores active vs. inactive — should be `case_assignments.active`.

### 5. `emancipation_checklist_reminder_notifier` (in `development/notifications.rake`) — same bug as #4

`lib/tasks/development/notifications.rake:6` has the identical `casa_case.case_assignments.first.volunteer_id` shape and will crash the same way. The whole `development/notifications.rake` file is also missing an `if Rails.env.development?` guard — so the dev tasks `followup_notifier` (`Followup.all.first`) and `reimbursement_complete_notifier` (`CaseContact.all.first`) could be triggered against prod data.

## Proposed PR scope

Single PR titled "Fix broken Heroku Scheduler rake tasks":

- [ ] Drop the `every 1.weeks do … end` and `every 1.days do … end` blocks in both reminder rake files (#1)
- [ ] Rename the task in `send_no_contact_made_reminder.rake` to `send_no_contact_made_reminder` and wire it to `NoContactMadeReminder` (#2)
- [ ] Move `lib/tasks/case_contact_types_reminder.rb`, `no_contact_made_reminder.rb`, `supervisor_weekly_digest.rb` to `app/services/`; drop the `require_relative` lines from their rake files and specs so Rails autoload picks them up (cleanup tied to #1 / #2)
- [ ] Delete `lib/tasks/recurring_jobs.rake` and `lib/tasks/example_recurring_task.rb` (or move the class to `app/jobs/` and gut the Bugsnag noise) (#3)
- [ ] Add nil-safety + switch to `case_assignments.active.find_each` in `youth_birthday_reminder.rake` (#4)
- [ ] Wrap `lib/tasks/development/notifications.rake` in `if Rails.env.development?` and apply the same nil-safety to the emancipation notifier (#5)

## How verified

Ran every project-defined rake task against current dev DB (35 cases, 7 volunteers, 2 birthday-next-month cases, 0 followups; today is Saturday May 2 so `monday?` / `day == 1` guards short-circuit the digest tasks):

| Task | Result |
|---|---|
| `send_case_contact_types_reminder` | exit 1 — `NoMethodError: undefined method 'every'` |
| `recurring:init` | exit 1 — `NameError: uninitialized constant ExampleRecurringTask` |
| `youth_birthday_reminder` (with `case_assignments` empty) | confirmed `NoMethodError` via isolated repro |
| `send_no_contact_made_reminder` | task name not registered (collides with `send_case_contact_types_reminder`) |
| `clear_passed_dates`, `court_report_due_reminder`, `volunteer_birthday_reminder`, `send_learning_hour_reports`, `factory_bot:lint`, `test_checker` | exit 0, no crash |

`post_gc_stat_to_discord` was skipped during verification because it HTTP-fetches the production URL — that's a separate concern (P3).

costajohnt:

Resolves #6923

### What changed, and _why_?

Several rake tasks driven by Heroku Scheduler were broken. Three crashed on every invocation and two crashed latently depending on production data shape. This works through all six items in the proposed scope.

- Dropped the `every ... do` blocks in `send_case_contact_types_reminder.rake` and `send_no_contact_made_reminder.rake`. There is no `every` defined at the rake top level (no `whenever` gem or `config/schedule.rb` in the repo), so both tasks raised `NoMethodError` on every run.
- Renamed the task in `send_no_contact_made_reminder.rake` to `send_no_contact_made_reminder` and wired it to `NoContactMadeReminder`. It had been declaring the duplicate name `send_case_contact_types_reminder` and calling the wrong service, so the no-contact reminder never ran and Scheduler invocations of `send_no_contact_made_reminder` failed with "Don't know how to build task".
- Moved `case_contact_types_reminder.rb`, `no_contact_made_reminder.rb`, and `supervisor_weekly_digest.rb` from `lib/tasks/` to `app/services/` so Rails autoloads them, and removed the `require_relative` lines from the rake files and specs. Their specs moved to `spec/services/` to match.
- Deleted `lib/tasks/recurring_jobs.rake` and `lib/tasks/example_recurring_task.rb`. `recurring:init` crashed with `uninitialized constant ExampleRecurringTask`, and the class was dead example code.
- Made `youth_birthday_reminder.rake` nil safe by iterating `case_assignments.active.find_each` instead of `case_assignments.first.volunteer_id`, which raised when a `birthday_next_month` case had no assignments. This also correctly notifies each active volunteer and ignores inactive assignments.
- Wrapped `lib/tasks/development/notifications.rake` in `if Rails.env.development?` so the dev-only notifier tasks are not defined in production, and applied the same nil safety to the emancipation reminder.

One related cleanup worth calling out: moving the three service files brought them under standardrb (they had been outside its scope while in `lib/tasks/`), which surfaced a few pre-existing lint issues in the moved code. I fixed those in the same pass, `update_attributes` to `update` (also required on Rails 8) in two services, and `Time.now` to `Time.current` in the supervisor digest.

Added `spec/lib/tasks/youth_birthday_reminder_spec.rb`, which drives the task through `Rake::Task[...].invoke` and asserts on `Noticed::Notification` count:

- a `birthday_next_month` case with no active assignment does not raise and sends nothing (this fails on the old `.first.volunteer_id` and passes on the fix)
- a single active assignment notifies one volunteer
- two active assignments notify both
- an active plus an inactive assignment notifies only the active one

The three moved service specs continue to pass unchanged (only the `require_relative` lines were removed).

Verified locally:

- `bundle exec rspec spec/services spec/lib/tasks` => 120 examples, 0 failures
- `bundle exec standardrb` => no offenses
- `rake -T` now shows `send_case_contact_types_reminder` and `send_no_contact_made_reminder` as two distinct tasks, `recurring:init` is gone, and the development notifier tasks appear only under `RAILS_ENV=development` (not `production`).

### Screenshots please :)

This is a backend rake-task change with no UI, so terminal output stands in for screenshots:

```
$ RAILS_ENV=development bundle exec rake -T | grep send_
rake send_case_contact_types_reminder   # Send an SMS to volunteers reminding them ...
rake send_no_contact_made_reminder       # Send an SMS to volunteers reminding them ...

$ RAILS_ENV=production bundle exec rake -T | grep -c emancipation_checklist_reminder_notifier
0

$ bundle exec rspec spec/services spec/lib/tasks
120 examples, 0 failures
```


compwron (dismissed):

compwron:
:)

PR #6994, opened 2026-06-09
Scan /spec/notifications for N+1s

### LINKED ISSUE #6912: Expand Prosopite N+1 enforcement: 12 newly-surfaced N+1s across 6 directories

stefannibrasil:

Closes https://github.com/rubyforgood/casa/issues/6912

### What changed, and why?

The error was raised because the query was being made every time it was called twice. By making the query only once, the error is fixed.

I also added more coverage to the notifier, memoized the reimbursement_amount to make sure it's calculated only once and moved some variables to be private methods to keep this class more object-oriented.

I also updated a request spec that uses this notifier to use instance_double instead of double. 
instance_double raises an error if methods are not available in the specified class or arguments have changed, whether a double would still pass if the method is removed.

### DIFF
--- a/app/notifications/reimbursement_complete_notifier.rb
+++ b/app/notifications/reimbursement_complete_notifier.rb
@@ -17,14 +17,25 @@ class ReimbursementCompleteNotifier < BaseNotifier
   end
 
   def message
-    case_contact = params[:case_contact]
     msg = "Volunteer #{case_contact.creator.display_name}'s request for reimbursement for #{case_contact.miles_driven}mi "
-    msg += " for $#{case_contact.reimbursement_amount} " if case_contact.reimbursement_amount
+    msg += " for $#{reimbursement_amount} " if reimbursement_amount
     msg += "on #{case_contact.occurred_at_display} has been processed and is en route."
     msg
   end
 
   def url
-    case_contacts_path(casa_case_id: params[:case_contact].casa_case_id)
+    case_contacts_path(casa_case_id: case_contact.casa_case_id)
+  end
+
+  private
+
+  def reimbursement_amount
+    return @reimbursement_amount if defined?(@reimbursement_amount)
+
+    @reimbursement_amount = case_contact.reimbursement_amount
+  end
+
+  def case_contact
+    params[:case_contact]
   end
 end
--- a/spec/requests/reimbursements_spec.rb
+++ b/spec/requests/reimbursements_spec.rb
@@ -2,14 +2,10 @@ require "rails_helper"
 
 RSpec.describe ReimbursementsController, type: :request do
   let(:admin) { create(:casa_admin) }
-  let(:casa_org) { admin.casa_org }
   let(:case_contact) { create(:case_contact) }
-  let(:notification_double) { instance_double("ReimbursementCompleteNotifier") }
 
   before do
     sign_in(admin)
-    allow(ReimbursementCompleteNotifier).to receive(:with).and_return(notification_double)
-    allow(notification_double).to receive(:deliver)
   end
 
   describe "GET /index" do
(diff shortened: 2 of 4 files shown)


app/notifications/reimbursement_complete_notifier.rb
```diff
@@ -17,14 +17,25 @@ def title
...
-    case_contacts_path(casa_case_id: params[:case_contact].casa_case_id)
+    case_contacts_path(casa_case_id: case_contact.casa_case_id)
+  end
+
+  private
+
+  def reimbursement_amount
+    return @reimbursement_amount if defined?(@reimbursement_amount)
+
+    @reimbursement_amount = case_contact.reimbursement_amount
```
compwron, line 35: How could case_contact have a method reimbursement_amount on it if we got it out of params? It is a string or json, not an object... so this would always come back as an error?

spec/requests/reimbursements_spec.rb
```diff
@@ -4,7 +4,7 @@
   let(:admin) { create(:casa_admin) }
   let(:casa_org) { admin.casa_org }
   let(:case_contact) { create(:case_contact) }
-  let(:notification_double) { double("ReimbursementCompleteNotifier") }
+  let(:notification_double) { instance_double("ReimbursementCompleteNotifier") }
```
compwron: I appreciate the improved double but I also want to avoid doubles and mocks whereever possible

app/notifications/reimbursement_complete_notifier.rb
stefannibrasil: Good question! `case_contact` is an object that can be fetched with `params[:case_contact]` because of the noticed gem. We send an object in https://github.com/rubyforgood/casa/blob/main/app/controllers/reimbursements_controller.rb#L37 I just moved this around; before we were calling `case_contact.reimbursement_amount` as well. The only difference was that we were initializing `case_contact = params[:case_contact]` in the `message` method, and in here I moved it to memoized instance.

I also tested this locally and the notifications get sent:
Screenshot 2026-06-11 at 2 27 20 PM

This notification class also has coverage, which I expanded, so if the param was wrong, the tests would have caught them.

spec/requests/reimbursements_spec.rb
stefannibrasil: Thank you for the nudge. I didn't want to change this too much at first but I updated the test to verify that the notifications get created in [343686c](https://github.com/rubyforgood/casa/pull/6994/commits/343686c4e069435715c2da83bf4abaa0ca4a4459)

compwron:
:)

PR #6517, opened 2025-09-26
create stubs for missing tests

compwron:
create stubs for missing tests

PR #7004, opened 2026-06-12
test: replace ByOrganizationScope and Roles spec stubs with real tests (#5563)

augustocbx:
Part of #5563 (umbrella — replaces 2 of the remaining stub specs; intentionally not using "Resolves" so the umbrella stays open)

### What changed, and _why_?
Replaced two `pending "add some tests for X"` stub specs (introduced in #6517) with real tests, following the definition of done in the issue body. Both files live in `spec/models/concerns/`, so they are bundled as one themed PR:

- `spec/models/concerns/by_organization_scope_spec.rb` — exercises the `.by_organization` scope through `CasaCase`: records from another `CasaOrg` are excluded, and an org with no records gets an empty result. This concern is the core of the app's multi-tenancy, so it seemed worth covering first.
- `spec/models/concerns/roles_spec.rb` — exercises `#role` for all four including user types: `Volunteer`, `Supervisor`, `CasaAdmin`, and the separate `AllCasaAdmin` model.

No `pending`, `skip`, or `xit` remains in either file.

This PR is tests. Commands run locally:

- `bundle exec rspec spec/models/concerns/by_organization_scope_spec.rb spec/models/concerns/roles_spec.rb` → 4 examples, 0 failures
- `bundle exec rspec spec/models/concerns/` → 6 examples, 0 failures, 2 pending (the two remaining stub files in that directory, `api_spec.rb` and `CasaCase/validations_spec.rb`, left for follow-up PRs)
- `bundle exec standardrb spec/models/concerns/` → clean

### Screenshots please :)
No UI changes — test-only PR.

### DIFF
--- a/spec/models/concerns/by_organization_scope_spec.rb
+++ b/spec/models/concerns/by_organization_scope_spec.rb
@@ -1,7 +1,21 @@
 require "rails_helper"
 
 RSpec.describe ByOrganizationScope, type: :model do
-  # TODO: Add tests for ByOrganizationScope
+  describe ".by_organization" do
+    let(:casa_org) { create(:casa_org) }
+    let(:other_casa_org) { create(:casa_org) }
 
-  pending "add some tests for ByOrganizationScope"
+    let!(:casa_case_in_org) { create(:casa_case, casa_org: casa_org) }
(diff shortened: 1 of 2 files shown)


compwron (approved):
yay! :)

PR #7039, opened 2026-07-04
5563 - Replace validator spec stubs with real tests

fuentesjr:
Part of #5563 — intentionally not `Resolves`, since the umbrella has more stubs remaining.

### What changed, and _why_?
Replaces the four `pending "add some tests for X"` stubs in `spec/validators/` with real specs (one commit per file):

- **UrlValidator** — valid http/https, disallowed scheme, missing host, malformed URI, and the custom `scheme:` option (allowed + rejected)
- **UserValidator** — phone number, display name, communication preferences, SMS-requires-phone, date of birth, and email uniqueness branches
- **CasaOrgValidator** — twilio phone number length / format / valid / blank branches
- **CourtReportValidator** — both mismatch error branches plus both valid status/submitted_at combinations

The specs characterize current behavior. Two oddities are signposted with NOTE comments rather than silently ratified: `CasaOrgValidator` adds its error to `:number`, which isn't a real `CasaOrg` attribute (the column is `twilio_phone_number`), and `UserValidator`'s error strings have a leading space baked in (already characterized in `spec/models/user_spec.rb`).

- `bundle exec rspec spec/validators/` → 32 examples, 0 failures, 0 pending
- `bundle exec standardrb spec/validators/` → no offenses
- no `pending` / `skip` / `xit` remains in the four files

### Screenshots please :)
N/A — test-only change, no UI.

---
**Follow-up candidates** (happy to open separate issues if maintainers agree these are bugs):
- `CasaOrgValidator` error key `:number` → should likely be `:twilio_phone_number`
- `CourtReportValidator` is wired up with a `fields:` option it never reads
- `UserValidator` error message strings have a leading space

### DIFF
--- a/spec/validators/casa_org_validator_spec.rb
+++ b/spec/validators/casa_org_validator_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe CasaOrgValidator, type: :validator do
 
       described_class.new.validate(casa_org)
 
-      expect(casa_org.errors.added?(:number, "must be 10 digits or 12 digits including country code (+1)")).to be true
+      expect(casa_org.errors[:number]).to include("must be 10 digits or 12 digits including country code (+1)")
     end
 
     it "adds an error on :number when the twilio phone number contains non-digit characters" do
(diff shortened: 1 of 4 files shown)


compwron (approved):
:)

Issue #7041, opened 2026-07-05
reimbursement requests


compwron:
<div class="gE iv gt" style="font-size: 0.875rem; padding: 20px 0px 0px; cursor: auto; color: rgb(34, 34, 34); font-family: &quot;Google Sans&quot;, Roboto, RobotoDraft, Helvetica, Arial, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><br class="Apple-interchange-newline">
Sarah Bosken <sbosken@pgcasa.org> | Sarah Bosken <sbosken@pgcasa.org> | Fri, May 1, 11:03 AM |   |  
-- | -- | -- | -- | --
Sarah Bosken <sbosken@pgcasa.org>
to me | to me
to me

[...]

PR #7055, opened 2026-07-15
Adding reimbursement status badge

Jberma23:
Addresses #7041 

### What changed, and _why_?
Added a ui feature to let volunteers know the current state of reimbursements to reduce emails to supervisors

Added new unit tests in spec/system/case_contacts/index_spec.rb and spec/decorators/case_contact_decorator_spec.rb to cover the new added content/feature

### Screenshots please :)
Screenshot 2026-07-15 at 12 19 30 PM
Screenshot 2026-07-15 at 12 22 05 PM

### DIFF
--- a/app/views/case_contacts/_case_contact.html.erb
+++ b/app/views/case_contacts/_case_contact.html.erb
@@ -13,6 +13,17 @@
               <% if !contact.active? %>
                 <span class="badge badge-pill light-bg text-black">Draft</span>
               <% end %>
+              <% if contact.want_driving_reimbursement? %>
+                <% decorated_contact = contact.decorate %>
+                <span class="ms-auto">
+                  <%= render BadgeComponent.new(
+                    text: decorated_contact.reimbursement_status_text,
+                    type: decorated_contact.reimbursement_status_badge_type,
+                    rounded: true,
+                    margin: false
+                  ) %>
+                </span>
+              <% end %>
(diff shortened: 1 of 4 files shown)


app/views/case_contacts/_case_contact.html.erb
```diff
@@ -13,6 +13,17 @@
               <% if !contact.active? %>
                 <span class="badge badge-pill light-bg text-black">Draft</span>
               <% end %>
+              <% if contact.want_driving_reimbursement? %>
+                <% decorated_contact = contact.decorate %>
```
arielsalcepk, line 17: Caching `contact.decorate` in a local variable here is actually a nice improvement over how the rest of this partial handles it, worth considering decorating once at the top of this partial and reusing it everywhere, rather than just in this new block. From what I see, right now `contact.decorate` gets called fresh four separate times above this (`medium_icon_classes`, `contact_groups`, `contact_types`, `subheading`), each one allocating a new decorator instance, and this renders once per case contact on the page, so it adds up across however many rows are showing. Not a blocker, but would be nice for anyone touching this file again, pulling the decoration to the top and reusing your variable throughout would clean up an existing inefficiency.

arielsalcepk: Could totally be done in a separate PR dedicated exclusively to improvements, and then testing the corresponding pieces that were touched.

arielsalcepk (approved):
Looks great to me! Thanks for adding this one so quickly 👍

compwron (approved):

Issue #6265, opened 2025-03-19
Bug: CI vulnerable to flaking when Github's servers are under high load


FireLemons:
## Impacted Users
 - contributing developers

## Environment
Github Continuous Integration(CI)

## Problem
When github's free CI servers are under heavy load, a race condition between the page loading and checking the database causes tests to flake. This is caused by a system test inputting data into a form then immediately checking the database without waiting for the form to finish submitting.

For every database check in the system files, make sure it's preceded by a capybara matcher with automatic waiting or replace the database check with a check for something to appear on the page.

Common matchers with built in waiting are `have_selector`, `have_css`, `have_content`, `have_button` ,`have_field`, and `have_link`.

## How to Replicate
It's difficult to replicate. To replicate it consistently you have to slow down how fast the page loads in your test browser.  

## Example failure output
Example 1:
```
Failures:

  1) case_contacts/edit is successful with mileage reimbursement on
     Failure/Error: expect(case_contact.contact_made).to be true

       expected true
            got false

     [Screenshot Image]: /home/runner/work/casa/casa/tmp/screenshots/failures_r_spec_example_groups_case_contacts_edit_is_successful_with_mileage_reimbursement_on_616.png


     # ./spec/system/case_contacts/edit_spec.rb:146:in `block (2 levels) in <top (required)>'
     # ./spec/rails_helper.rb:132:in `block (2 levels) in <top (required)>'
     # ./spec/rails_helper.rb:117:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:116:in `block (2 levels) in <top (required)>'

  2) case_contacts/edit when admin admin successfully edits case contact with mileage reimbursement
     Failure/Error: expect(case_contact.contact_made).to be true

       expected true
            got false

     [Screenshot Image]: /home/runner/work/casa/casa/tmp/screenshots/failures_r_spec_example_groups_case_contacts_edit_when_admin_admin_successfully_edits_case_contact_with_mileage_reimbursement_195.png


     # ./spec/system/case_contacts/edit_spec.rb:53:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:132:in `block (2 levels) in <top (required)>'
     # ./spec/rails_helper.rb:117:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:116:in `block (2 levels) in <top (required)>'

  3) case_court_reports/index as a supervisor when searching for cases selects the correct case
[...]

PR #7056, opened 2026-07-15
Email volunteers when their reimbursement request is marked complete

arielsalcepk:
Addresses #7041 — this covers the email-notification half of the ticket only. The status-visibility half is #7055, opened separately by @Jberma23. Not marking this `Resolves` since #7041 shouldn't auto-close until both land.

### What changed, and _why_?
Wires up email delivery on `ReimbursementCompleteNotifier`. Volunteers now get an email when a supervisor marks their reimbursement request complete, previously this notification was in-app only (the `deliver_by :email` block existed but was commented-out scaffolding).

- Email delivery is scoped to the volunteer recipient only (`recipient.volunteer? && recipient.receive_email_notifications?`), not the supervisor who triggered the change, since they already know they just did it. The existing in-app notification still goes to both; only the new email channel is scoped.
- `VolunteerMailer#reimbursement_complete_email` takes plain positional args (`user, case_contact`) to match this mailer's existing methods, rather than Noticed's default `.with(params)` calling convention.
- Added `DebugPreviewMailer#no_data`, separate from the existing `invalid_user`, for the case where a looked-up preview user is valid but has no related record to render (a volunteer with no case contact wanting reimbursement). Keeps that fallback message accurate instead of reusing `invalid_user`'s "ID didn't resolve" wording for a different situation.

- `bundle exec rspec`: 3625 passing, 1 pre-existing failure unrelated to this change (`spec/system/case_contacts/case_contacts_new_design_spec.rb`, a columns-panel system spec; consistent with the flakiness already tracked in #6265). Will post in `#casa` per the note below if this recurs in the Actions build on this PR.
- `bundle exec standardrb`: clean
- New coverage: `spec/mailers/volunteer_mailer_spec.rb`, `spec/notifications/reimbursement_complete_notifier_spec.rb`, `spec/mailers/previews/volunteer_mailer_preview_spec.rb`
- No new system/Capybara spec, this PR is backend mailer/notifier plumbing with no new UI, the existing reimbursements-page system specs already cover the `mark_as_complete` interaction this hooks into.
- Verified manually end-to-end locally: created a reimbursement-eligible case contact as a volunteer, marked it complete as a supervisor, confirmed the real email fires (not a preview) with the correct recipient, subject, and content via Letter Opener.

### Screenshots please :)

Case before getting marked as complete:
image

Right after, this file was generated to my system with the HTML for the resulting email:
Screenshot 2026-07-15 at 3 01 22 PM

### Feelings gif (optional)


spec/mailers/volunteer_mailer_spec.rb
```diff
@@ -41,6 +41,26 @@
...
+      expect(mail.subject).to eq("Your reimbursement request has been processed")
+      expect(mail.body.encoded).to match("#{case_contact.miles_driven}mi")
+      expect(mail.body.encoded).to match(case_contact.occurred_at_display)
+    end
+
+    context "when the casa org has a mileage rate" do
+      let!(:mileage_rate) { create(:mileage_rate, casa_org: case_contact.casa_case.casa_org, amount: 6.50, effective_date: 3.days.ago) }
+
+      it "includes the reimbursement amount" do
+        expect(mail.body.encoded).to match("$")
```
Jberma23: I think this is going to catch any string consider $ is a regex end-anchor. Think you may need to use something like match(/\$[\d,]+\.\d{2}/)

Jberma23 (changes requested):
So looking at the logic I think it would send this email out even if they uncheck the reimbursement checkbox. So we probably need to check for that as well

spec/mailers/volunteer_mailer_spec.rb
arielsalcepk: @Jberma23 Fixed both, thanks for catching these, and here's what changed for a brief summary (although please do check this further):
Almost used your regex as is, just had to escape $ and . as well, otherwise those are the end-of-line anchor and any-character wildcard, not literal characters. And the test now uses: /\$[\d,]+\.\d{2}/. 👍 

Controller now guards the notifier call with ActiveModel::Type::Boolean.new.cast(reimbursement_complete), so unchecking no longer fires it. Worth noting: that bug predates this PR, was harmless before as an in-app-only notification, but our email delivery is what made it consequential. This PR successfully both exposed and fixed it, specially thanks to your call out!

Added tests covering both directions and value formats, plus the no-mileage-rate case.

Jberma23:
Sounds good thanks for addressing those!

Jberma23 (approved):

compwron (approved):

PR #7068, opened 2026-07-17
5563 - Replace value parameter spec stubs with real tests

fuentesjr:
Part of #5563, intentionally not `Resolves`, since the umbrella has more stubs remaining.

### What changed, and _why_?
Replaces the four `pending "add some tests for X"` stubs in `spec/values/` with real specs (one commit per file), following the patterns in `case_contact_parameters_spec.rb`, `volunteer_parameters_spec.rb`, and `banner_parameters_spec.rb`:

- **UserParameters** carries the bulk of the coverage: the permit allowlist (including nested `address_attributes`), `ParameterMissing` on a missing root key, and every builder method (`with_organization`, `with_password`, `with_type`, `without_type`, `without_active`, `with_only`, `without`)
- **SupervisorParameters / CasaAdminParameters** are thin subclasses that only fix the root param key (`:supervisor` / `:casa_admin`); specs pin the root key, `ParameterMissing`, and one inherited builder each rather than re-testing the parent
- **AllCasaAdminParameters** is a separate `SimpleDelegator`: `email`/`password` passthrough, unpermitted filtering, `ParameterMissing`, and `with_password`

Rebased on current main after #7070 landed. #7070 already added a `user_parameters_spec.rb`; this PR extends that file with the broader value-object coverage above (allowlist, `ParameterMissing`, `with_type`, `with_only`) and adds the three sibling parameter specs it didn't touch. The `#without` specs here assert the corrected behavior from #7070, not the old bug. No app code changed.

- `bundle exec rspec spec/values/user_parameters_spec.rb spec/values/supervisor_parameters_spec.rb spec/values/casa_admin_parameters_spec.rb spec/values/all_casa_admin_parameters_spec.rb` gives 22 examples, 0 failures, 0 pending
- `bundle exec standardrb spec/values/` reports no offenses
- No `pending` / `skip` / `xit` remains in the four files

### Screenshots please :)
N/A, test-only change, no UI.

### DIFF
--- a/spec/values/all_casa_admin_parameters_spec.rb
+++ b/spec/values/all_casa_admin_parameters_spec.rb
@@ -1,7 +1,39 @@
 require "rails_helper"
 
 RSpec.describe AllCasaAdminParameters do
-  # TODO: Add tests for AllCasaAdminParameters
+  subject { described_class.new(params) }
 
-  pending "add some tests for AllCasaAdminParameters"
+  let(:params) {
+    ActionController::Parameters.new(
+      all_casa_admin: ActionController::Parameters.new(
+        email: "all_admin@example.com",
+        password: "password123"
+      )
+    )
+  }
+
+  it "permits the allowed attributes" do
+    expect(subject["email"]).to eq("all_admin@example.com")
+    expect(subject["password"]).to eq("password123")
+  end
+
+  it "filters out attributes that are not permitted" do
+    params[:all_casa_admin][:admin] = true
+    expect(subject["admin"]).to be_nil
+  end
+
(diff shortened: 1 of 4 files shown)


fuentesjr:
Rebased on main now that #7070 has merged. Reconciled `user_parameters_spec.rb` so it extends the coverage from #7070 (adds the permit allowlist, `ParameterMissing`, `with_type`, and `with_only`) and asserts the corrected `#without` behavior rather than the old bug. The other three files (supervisor, casa_admin, all_casa_admin parameter specs) were untouched by #7070. Green locally: 22 examples, 0 failures.

compwron (commented):
:)

PR #7070, opened 2026-07-17
Fix UserParameters#without being a no-op

### LINKED ISSUE #7067: Bug: UserParameters#without is a no-op with symbol keys, leaking active/type into admin creation
## Impacted User Types
- **admins** — can create a CasaAdmin with `active=false` via an extra field in the create request
- **volunteers / supervisors** — can trigger an unhandled 500 on `POST /casa_admins`

## Environment
`main` @ b7b87cd65, Rails 8.0.5. Reproduced locally in both dev and test. Not a UI bug, so no screenshot — the repro below is request-level.

## Current Behavior
`UserParameters#without` (`app/values/user_parameters.rb#L51-L53`) never removes anything when called with symbols:

```ruby
def without(*keys)
  params.reject! { |key| keys.include?(key) }
end
```

`ActionController::Parameters#reject!` yields **string** keys, so `keys.include?(key)` compares `[:active, :type].include?("active")` → always `false`.

Its only caller passes symbols — `CreateCasaAdminService#build` (`app/services/create_casa_admin_service.rb#L10-L17`):

```ruby
CasaAdminParameters.new(@params)
  .with_password(SecureRandom.hex(10))
  .with_organization(@current_organization)
  .without(:active, :type)
```

Both controllers hand the service raw `params`, and `:active`/`:type` are both on the `UserParameters` permit allowlist — so both survive into `CasaAdmin.new`:
- `active=false` persists (DB default is `true`).
- `type=<sibling class>` raises `ActiveRecord::SubclassNotFound`, which nothing rescues → 500. Rails' STI guard is what prevents this being a type-escalation; only `type=CasaAdmin` is accepted.

Separately, `without` is the only builder in the class that doesn't return `self`, so it can't be chained.

## Expected Behavior
`.without(:active, :type)` strips both keys, so a new admin always gets `active` from the DB default and `type` from STI, regardless of the request body.

## How to Replicate
```ruby
# 1. active leaks — as an admin
sign_in create(:casa_admin, casa_org: org)
post casa_admins_path, params: {casa_admin: {email: "leak@example.com", display_name: "L", active: "false"}}
CasaAdmin.find_by(email: "leak@example.com").active  # => false, expected true

# 2. type raises — as a volunteer
sign_in create(:volunteer, casa_org: org)
post casa_admins_path, params: {casa_admin: {email: "x@e.com", display_name: "X", type: "Volunteer"}}
# => ActiveRecord::SubclassNotFound
```
Note (2) surfaces as `SubclassNotFound` rather than `Pundit::NotAuthorizedError` because `CasaAdminsController#create` calls `service.build` before `authorize` (`app/controllers/casa_admins_controller.rb#L41-L43`). Nothing is persisted, but any authenticated user can trigger the error.

## Suggested Fix
```ruby
def without(*keys)
  keys = keys.map(&:to_s)
  params.reject! { |key| keys.include?(key) }
  self
end
```

Happy to open a PR. Found while replacing the `spec/values` stub specs for #5563; those specs currently characterize the buggy behavior with a `NOTE` pointing here.

dylangriffinshub:
## Summary

Fixes #7067 — `UserParameters#without` compared symbol keys against `ActionController::Parameters#reject!`'s string keys, so its condition was always false and the method silently removed nothing. It also didn't return `self`, breaking the builder chain.

The only caller, `CreateCasaAdminService#build`, calls `.without(:active, :type)` expecting both to be stripped before `CasaAdmin.new`. Because the call was a no-op:
- Any authenticated user could set `active=false` on a new admin via an extra field in the create request (DB default is `true`).
- Passing `type=<sibling class>` raised an unrescued `ActiveRecord::SubclassNotFound` (500), since Rails' STI guard is the only thing preventing type escalation.

## Fix

`without` now stringifies the incoming keys before comparing (matching how `reject!` yields keys), and returns `self` so it can still be chained.

Also replaces the pending stub in `spec/values/user_parameters_spec.rb` with real coverage for `without`, `without_type`, `without_active`, `with_organization`, and `with_password`.

## Test plan
- [ ] `bundle exec rspec spec/values/user_parameters_spec.rb`
- [ ] Manually verify: as a volunteer/supervisor, `POST /casa_admins` with an extra `type` or `active` field no longer leaks through

### DIFF
--- a/app/values/user_parameters.rb
+++ b/app/values/user_parameters.rb
@@ -49,7 +49,9 @@ class UserParameters < SimpleDelegator
   end
 
   def without(*keys)
+    keys = keys.map(&:to_s)
     params.reject! { |key| keys.include?(key) }
+    self
   end
 
   private
--- a/spec/values/user_parameters_spec.rb
+++ b/spec/values/user_parameters_spec.rb
@@ -1,7 +1,75 @@
 require "rails_helper"
 
 RSpec.describe UserParameters do
-  # TODO: Add tests for UserParameters
+  def build_params(attrs)
+    ActionController::Parameters.new(user: attrs)
+  end
 
-  pending "add some tests for UserParameters"
+  describe "#without" do
+    it "removes the given keys and is chainable" do
+      params = build_params(display_name: "Jane", active: "false", type: "Volunteer")
+
+      result = described_class.new(params).without(:active, :type)
+
+      expect(result).to be_a(described_class)
+      expect(result.to_h).to eq("display_name" => "Jane")
+    end
+
+    it "removes keys even when the params only has string keys" do
+      params = build_params(display_name: "Jane", active: "false")
+
+      result = described_class.new(params).without(:active)
+
+      expect(result.to_h).to eq("display_name" => "Jane")
+    end
+
+    it "supports further chaining after without" do
+      params = build_params(display_name: "Jane", active: "false", password: "old")
+
+      result = described_class.new(params).without(:active).with_password("new-password")
+
+      expect(result.to_h).to eq("display_name" => "Jane", "password" => "new-password")
+    end
+  end
+
+  describe "#without_type" do
+    it "removes the type key" do
+      params = build_params(display_name: "Jane", type: "Volunteer")
+
+      result = described_class.new(params).without_type
+
+      expect(result.to_h).to eq("display_name" => "Jane")
+    end
+  end
+
+  describe "#without_active" do
+    it "removes the active key" do
+      params = build_params(display_name: "Jane", active: "false")
+
+      result = described_class.new(params).without_active
+
+      expect(result.to_h).to eq("display_name" => "Jane")
+    end
+  end
+
+  describe "#with_organization" do
+    it "sets casa_org_id" do
+      params = build_params(display_name: "Jane")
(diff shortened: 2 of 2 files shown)


fuentesjr:
Nice, thanks for jumping on this. The fix looks right. One heads-up on overlap. I opened this issue while doing the `spec/values` stub replacements for #5563, and that spec work is up as #7068 (green, in review). It rewrites this same `user_parameters_spec.rb` from the stub, and covers more (permit allowlist, `ParameterMissing`, every builder, plus the supervisor/casa_admin/all_casa_admin sibling files). The catch: since that PR is test only, its `#without` specs pin down the current broken behavior (they assert the key is *not* removed and that `self` is *not* returned), so once your fix lands, those two specs will start failing and need updating.

Since the only overlap is this one file, maybe the cleanest split is this: let #7068 land the coverage first, then rebase this PR to keep the `user_parameters.rb` fix and just update those two `#without` specs to assert the corrected behavior (key removed, returns `self`), dropping the spec additions here since they'd already be covered. Your call and the maintainers'. Happy to go whichever way keeps it simplest for you. Either way, the `user_parameters.rb` fix here is the part that matters.

compwron (approved):
:)

Issue #7040, opened 2026-07-04
Deploy log cleanup / hardening (casa-qa v3413)


compwron:
## Deploy log cleanup / hardening

The `casa-qa` Heroku build (v3413, Heroku-26 stack) **succeeds and deploys**, but the build log surfaces several warnings worth cleaning up. Full TODO plan gist: https://gist.github.com/compwron/56813f4c5a6665d6d284c7d50b3d8f07

Items inline below, roughly in priority order.

### High priority (real impact)

- [ ] **Disable on-demand asset compilation in production.** Log warns `config.assets.compile = true` in production. Assets are already precompiled at build time (`rake assets:precompile` runs fine), so on-demand compile only adds request latency and load. → Set `config.assets.compile = false` in `config/environments/production.rb`.
- [ ] **Commit a normalized Gemfile.lock.** Build warns `ffi-1.17.2` platform-specific gem is installed but the lockfile only has the generic ruby version. → Run `bundle lock --normalize-platforms` (or `bundle lock --add-platform x86_64-linux`) and commit the updated `Gemfile.lock`.

### Medium priority (hygiene)

- [ ] **Address npm vulnerabilities.** `4 vulnerabilities (3 moderate, 1 high)` in the full tree; `1 moderate` remains in production deps after pruning. → Run `npm audit`, then `npm audit fix`. Note EOL/vulnerable build tooling: `eslint@8.57.1`, `glob@10.5.0`.
- [ ] **Bump Ruby patch version.** Deploying on `ruby-4.0.3`; Heroku recommends `4.0.5` (security + bug fixes). → Update `.ruby-version` to `4.0.5`, test, redeploy.
- [ ] **Verify the release-phase command actually succeeded.** A release command is declared (`this new release will not be available until the command succeeds`); the log ends before showing its output. → Confirm the release/migration step passed in the release-phase logs.

### Low priority (noise / optional)

- [ ] **Refresh deprecated npm packages.** Transitive deprecations from build/lint tooling: `rimraf@3`, `glob@7`/`glob@10.5.0`, `inflight@1.0.6` (memory leak), `coffee-script`, `lodash.template`, `whatwg-encoding@3.1.1`. Runtime unaffected — candidate for a dependency refresh.
- [ ] **Review npm install-script allowlist.** `esbuild@0.28.1` and `unrs-resolver@1.12.2` have install scripts not yet on `allowScripts`. Build works; review/approve only if hardening the supply chain.
[...]

PR #7071, opened 2026-07-18
chore: disable on-demand asset compilation and normalize Gemfile.lock platforms (#7040)

augustocbx:
Addresses the two **High priority (real impact)** items of #7040. The remaining medium/low-priority checklist items (npm audit, Ruby patch bump, deprecated packages, install-script allowlist) are left as follow-ups, so this does not close the issue.

### What changed, and _why_?

Two deploy-log hardening items from #7040:

1. **Disabled on-demand asset compilation in production.** The `casa-qa` build log warns that `config.assets.compile = true` in production. Assets are already precompiled at build time (`rake assets:precompile` runs during the build), so on-demand compilation only adds per-request latency and load — and it silently masks a missing precompiled asset instead of failing fast. Set `config.assets.compile = false` in `config/environments/production.rb`, which is the standard Rails production default for a precompiled app.

2. **Normalized `Gemfile.lock` platforms.** The build warns that the platform-specific `ffi-1.17.2` gem is installed while the lockfile only carried OS-version-pinned platforms. Ran `bundle lock --normalize-platforms`, which collapses the version-pinned entries (`arm64-darwin-21/22`, `x86_64-darwin-21/23`) to their generic forms (`arm64-darwin`, `x86_64-darwin`). The `ruby` and `x86_64-linux` (Heroku) platforms are retained and the platform-specific gem specs are unchanged. `bundle check` passes after the change.

These are production config / lockfile changes with no application-behavior surface, so there is no meaningful RSpec/Jest test to add (the test environment does not load `config/environments/production.rb`, and there is no existing convention in this repo for asserting environment config). Verification instead:

- `ruby -c config/environments/production.rb` → `Syntax OK`.
- `bundle exec standardrb config/environments/production.rb` → exit 0, no offenses.
- `bundle lock --normalize-platforms` produced a minimal diff touching only the `PLATFORMS` block (2 insertions, 4 deletions); `bundle check` → "The Gemfile's dependencies are satisfied".
- Assets continue to be resolved from the precompiled manifest; the build already runs `rake assets:precompile` successfully, so `compile = false` has all referenced assets available.

### Screenshots please :)

Not applicable — no user-facing UI change (production configuration and dependency lockfile only).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -695,11 +695,9 @@ GEM
     zeitwerk (2.8.2)
 
 PLATFORMS
-  arm64-darwin-21
-  arm64-darwin-22
+  arm64-darwin
   ruby
-  x86_64-darwin-21
-  x86_64-darwin-23
+  x86_64-darwin
   x86_64-linux
 
 DEPENDENCIES
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -42,6 +42,9 @@ Rails.application.configure do
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
   # config.asset_host = "http://assets.example.com"
 
+  # Do not fall back to on-demand asset compilation; assets are precompiled at build time.
+  config.assets.compile = false
+
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :microsoft
 


compwron (approved):
:)

PR #7087, opened 2026-07-29
Part2 of design rebuild - DO NOT MERGE YET

giacoelho:
open part 2 with the follow-ups carried over from #7051

Starting point for the part-2 tracking PR. Five items, each verified still open against main rather than copied forward from notes:

- the 3 xit'd flaky case-contact system specs, which need a tracking issue before they can be re-enabled
- two view specs still raising SystemStackError from the settings rail calling edit_casa_org_path(current_organization) unstubbed (pre-existing, still red)
- sort still living on the filterrific form, which is why Clear filters has to carry sorted_by by hand
- mailers naming the chapter inconsistently (@casa_organization vs @casa_org vs nothing on Devise)
- contact types on the case-contact form staying an exposed checkbox fieldset, with the reason it is deliberate

Dropped one item while checking: the read-notification icon is already slate-500, so its contrast was fixed in the merged work and is not outstanding.

### DIFF
--- a/app/assets/stylesheets/tailwind.css
+++ b/app/assets/stylesheets/tailwind.css
@@ -59,8 +59,10 @@
    TomSelect's opaque `.ts-control` paints over a plain `::after` (that stacking, not a
    missing rule, is why every earlier chevron was invisible). base64 is minifier- and
    parser-safe; a raw `data:` URI and a `content` glyph escape both broke in the build, and
-   the injected icon-font element never painted. Pixel-verified to match the single-selects
-   (identical 8x4 dark extent), not just checked by computed style. */
+   the injected icon-font element never painted. Pixel-verified against a native single-select's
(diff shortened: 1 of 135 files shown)


stefannibrasil:
Hi @giacoelho I just wanted to let you know that this PR introduced a few errors that is impacting other work... is it part of the work to address these errors in a follow up PR? Thank you!

fuentesjr:
@stefannibrasil @seanmarcia I may be missing context, but is there any reason why we can't revert this PR? The title seems to suggest this is not ready for main yet

seanmarcia:
It was merged to let the stakeholders do usability testing on the design.

fuentesjr:
@stefannibrasil @seanmarcia @giacoelho Ah I see. Thank you for that bit of context. Well I definitely don't want to step on any one's toes, but I'm more than happy to help out in any way if needed 🙏🏽

Issue #7093, opened 2026-08-06
Bug: URGENT all court report downloads are broken


compwron:
## Reported by stakeholder

Nancy King, Office Manager at **Voices for Children Montgomery**, reports via email:

> Volunteers, Case Supervisors, and myself are getting the following error when trying to generate a report on the tracking form. I have tried Microsoft Edge and Google Chrome.

**Note:** the actual error text/screenshot did not come through in the email we received, so the specific failure is not yet known. We have asked for a follow-up with the exact message.

## Impact

- Affects **all roles** at this org (volunteers, case supervisors, and org staff), not one user.
- Reproduces in both Edge and Chrome, so it is unlikely to be browser-specific.

## Info still needed

- [ ] Screenshot or exact text of the error
- [ ] Which page/button — "tracking form" is stakeholder language, not a term in the app. Most likely candidates:
  - **Court Reports** page (`/case_court_reports`) → "Generate Report" (produces the .docx)
  - **Reports / Export Data** page (`/reports`) → one of the CSV exports
- [ ] The case number (or date range) being used, so we can reproduce with their data
- [ ] Approximate timestamp of a failure, to line it up with production logs

## Triage pointers

If it is the court report .docx generation, the relevant path is
`CaseCourtReportsController#generate` (`app/controllers/case_court_reports_controller.rb`),
which renders JSON and surfaces `e.to_s` to the user for any unhandled error, plus a
"Template is not found" message on `Zip::Error`. Voices for Children Montgomery uses a
custom court report template (see closed #1826 / #2265 / #3698), so a template/placeholder
mismatch is worth ruling out — `app/documents/templates/montgomery_report_template.docx`.

If it is a CSV export, the relevant services are under `app/services/*_export_csv_service.rb`.

## Related history (all closed, for context)

- #1826 / #2265 / #3698 — Montgomery-specific court report template and letterhead
- #1039 — "Generate Report is broken"
- #1253 — exception when downloading report with options
- #1556 — contact_type group filter caused an unhandled error when generating reports

compwron:
## Root cause found: Ruby 4.0's stripped `cgi` stdlib + Active Storage Azure

The stakeholder error text is `undefined method 'parse' for class CGI`. This is **not** a Montgomery template problem — it is every Active Storage read/write in production.

### Mechanism

[...]

PR #7108, opened 2026-08-08
fix(storage): replace retired azure-storage-blob with maintained azure-blob

### LINKED ISSUE #7094: [important!] Rewrite/update Azure integration
https://azure.microsoft.com/en-us/updates?id=retirement-notice-the-azure-storage-ruby-client-libraries-will-be-retired-on-13-september-2024

https://github.com/Azure/azure-storage-ruby

 This project will be in Community Support until 13 September 2024. After this date the project and associated client libraries will be retired permanently. For more details on the retirement and alternatives to using this project, visit Retirement notice: The Azure Storage Ruby client libraries will be retired on 13 September 2024.

 Azure Storage Ruby client libraries will be retired on 13 September 2024. Before that date, you’ll need to transition to calling the Azure Storage REST API directly from your Ruby applications.
The Azure Storage Ruby client libraries were previously in Community Support and will remain in Community Support until their retirement date. Community Support is defined as:
“SDK will no longer receive updates from Microsoft unless otherwise specified in the separate customer agreement. The package will remain available via public package managers and the GitHub repo, which can be maintained by the community.” (See Azure SDK lifecycle and support policy for more information)
After 13 September 2024, the older client libraries can still be used, but any associated updates will no longer be issued by Microsoft. If you choose not to switch to the Azure Storage REST API directly, you can find the source code for the current Azure Storage Ruby client libraries on GitHub as an open-source resource. You can continue to code and apply your own fixes as required.

tacoda:
Resolves #7094

### What changed, and _why_?

The azure-storage-blob / azure-storage-common gems were retired by Microsoft on 13 Sep 2024 and are unmaintained. They also forced a manual `cgi ~> 0.5.1` pin under Ruby 4.0 (azure-storage-common needs CGI.parse, which Ruby 4.0's stripped stdlib omits and Rails 8 no longer pulls in transitively).

Swap to testdouble/azure-blob, a maintained drop-in Active Storage adapter:

- `Gemfile`: `azure-storage-blob` -> `azure-blob`; drop the now-redundant cgi pin (azure-blob declares cgi as a dependency, so the full library resolves again).
- `config/storage.yml`: microsoft service `AzureStorage` -> `AzureBlob`.
- Rewrite the signing regression guard to exercise the `AzureBlob` Active Storage service (offline URL signing for download + direct upload), replacing the old azure-storage-common internal-class assertions.

Same Azure account / container / keys, so no data migration or infra change.

_Note: We love [capybara](https://rubydoc.info/github/teamcapybara/capybara) tests! If you are writing both haml/js and ruby, please try to test your work with tests at every level including system tests like https://github.com/rubyforgood/casa/tree/main/spec/system_ 

One file touched: `spec/lib/azure_storage_signing_spec.rb` (rewritten). Dropped tests coupled to methods that are gone with `azure-storage-blob`. Same intent, re-pointed at the real Active Storage adapter. Builds `ActiveStorage::Service::AzureBlobService` with dummy credentials.

- It has the full CGI library, not Ruby 4.0's escape-only stdlib.
- It signs a private download URL locally, without hitting Azure.
- It signs a direct-upload URL locally, without hitting Azure.
- Offline: signing is pure-local (SharedKey), no network/live Azure. Runs in CI.
- Higher-level than before: tests the Active Storage service contract, not gem internals → survives future gem/Ruby bumps.
- Result: 3 examples, 0 failures.

Gap (unchanged, flagged earlier)

Still no real round-trip test (upload→store→download against live Azure). This path needs credentials. Covered only by staging smoke-test before merge. App test env uses Disk storage, so the Azure path is exercised nowhere else. That blind spot is exactly why #7093 stayed green.

Needs to go through integration testing through a real Heroku environment to verify a connection to the real Azure account in an environment that is production-like.

### Screenshots please :)

N/A - Backend only change

### Feelings gif (optional)
_What gif best describes your feeling working on this issue?_
Indiana Jones

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,16 +8,11 @@ gem "rails", "~> 8.0"
 gem "after_party" # Post-deployment tasks
 gem "amazing_print" # Easier console reading
 gem "authtrail" # Track Devise login activity
-gem "azure-storage-blob", require: false
+gem "azure-blob", require: false # Active Storage adapter for Azure (maintained replacement for retired azure-storage-blob)
 gem "blueprinter" # JSON serialization
 gem "bugsnag" # Error tracking in production
 gem "caxlsx", "~> 4.5" # Excel spreadsheets - TODO can we remove this version restriction?
 gem "caxlsx_rails", "~> 0.7.1" # Excel spreadsheets - TODO can we remove this version restriction?
-# Ruby 4.0 ships a cgi stdlib with only escape/unescape; azure-storage-common needs the
-# full library's CGI.parse to sign blob URLs. Rails 8 no longer pulls cgi in transitively,
-# so without this the stripped stdlib wins and every Active Storage read/write 500s.
-# Guarded by spec/lib/azure_storage_signing_spec.rb. Remove when azure-storage-blob is dropped.
-gem "cgi", "~> 0.5.1"
 gem "cssbundling-rails", "~> 1.4" # CSS compilation
 gem "delayed_job_active_record" # Background job processing
 gem "devise" # Authentication
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -92,14 +92,9 @@ GEM
       dumb_delegator
     axe-core-rspec (4.12.0)
       axe-core-api (= 4.12.0)
-    azure-storage-blob (2.0.3)
-      azure-storage-common (~> 2.0)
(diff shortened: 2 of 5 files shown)


compwron (approved):
I love it. We will test in staging :)

PR #7109, opened 2026-08-09
Fix binary court report downloads

### LINKED ISSUE #7099: Use File.binread and an explicit filename when sending court report .docx
## What

`CaseCourtReportsController#show` sends the generated .docx with two small problems:

```ruby
format.docx do
  @casa_case.latest_court_report.open do |file|
    # TODO test this .read being present, we've broken it twice now
    send_data File.read(file.path), type: :docx, disposition: "attachment", status: :ok
  end
end
```

**1. `File.read` should be `File.binread`.** `File.read` reads in text mode, so binary docx content comes back tagged UTF-8:

```
File.read     encoding=UTF-8      length=365389  bytesize=379838  valid_encoding?=false
File.binread  encoding=ASCII-8BIT length=379838  bytesize=379838
```

Note `length` and `bytesize` disagree by ~14KB. Nothing in the current stack misuses `length` over `bytesize` — verified that `Content-Length` comes out correct and the download is byte-exact — so this is **not** currently causing a bug. But handing `send_data` a UTF-8-tagged string full of invalid UTF-8 is asking for trouble, and that line already carries a `we've broken it twice now` comment (see #1253, and 17883f49a "fix: court report document show").

**2. `send_data` has no `filename:`.** With only `disposition: "attachment"` and no filename, the browser derives the download name from the URL path. That mostly works, but it is incidental rather than intentional, and it breaks down for case numbers containing characters that get escaped in a URL. A stakeholder-reported file in #7093 arrived named `_C-15-JV-24-191.docx` with a leading underscore that does not appear in the case number.

## Suggested change

```ruby
send_data File.binread(file.path),
  type: :docx,
  filename: "#{@casa_case.case_number}.docx",
  disposition: "attachment",
  status: :ok
```

## Testing

Worth resolving the existing TODO at the same time — there is currently no assertion that the downloaded bytes match the stored blob. Something like: generate a report, GET the `.docx` link, and assert `response.body.bytesize == blob.byte_size`, that the body starts with `PK\x03\x04`, and that `Content-Disposition` carries the expected filename.

## Context

Found while debugging #7093. `File.read` was briefly a suspect for the corrupt-download bug there and was ruled out — the real cause was zip64 (#7098). Filing separately so it does not get lost.

alloutflo:
Resolves #7099

### What changed, and _why_?

- Read the stored DOCX with `File.binread`, so the response body remains binary (`ASCII-8BIT`) instead of being tagged as invalid UTF-8 text.
- Pass `#{@casa_case.case_number}.docx` explicitly to `send_data`, so escaped characters in the request URL cannot alter the downloaded filename.
- Replace the empty temporary attachment in the request spec with a real DOCX fixture and verify exact response bytes, blob byte size, ZIP magic, and `Content-Disposition`.
- Use a case number containing `#` and a deliberately different stored blob name, so the regression covers the reported filename edge case.

The new example fails against the old controller implementation because the response has the wrong encoding and no explicit filename, then passes with this change.

- `bundle exec rspec spec/requests/case_court_reports_spec.rb:55` — 4 examples, 0 failures
- `bundle exec standardrb app/controllers/case_court_reports_controller.rb spec/requests/case_court_reports_spec.rb` — passed
- `ruby -c` for both changed Ruby files — passed
- `git diff --check` — passed
- Full `spec/requests/case_court_reports_spec.rb`: 27 examples passed; three unchanged index examples could not render because the local test setup did not contain the built `application.js` asset. The focused DOCX context and all changed-code checks pass.

### Screenshots please :)

Not applicable; this changes the binary download response and request coverage, with no visual UI changes.

### Feelings gif (optional)

AI disclosure: I used an AI coding assistant to help inspect the issue, implement the focused change and regression coverage, and run the checks listed above. I reviewed the final diff and test results before submitting.

### DIFF
--- a/app/controllers/case_court_reports_controller.rb
+++ b/app/controllers/case_court_reports_controller.rb
@@ -24,8 +24,11 @@ class CaseCourtReportsController < ApplicationController
     respond_to do |format|
       format.docx do
         @casa_case.latest_court_report.open do |file|
-          # TODO test this .read being present, we've broken it twice now
-          send_data File.read(file.path), type: :docx, disposition: "attachment", status: :ok
+          send_data File.binread(file.path),
+            type: :docx,
+            filename: "#{@casa_case.case_number}.docx",
+            disposition: "attachment",
+            status: :ok
         end
       end
     end
--- a/spec/requests/case_court_reports_spec.rb
+++ b/spec/requests/case_court_reports_spec.rb
@@ -53,13 +53,14 @@ RSpec.describe "/case_court_reports", type: :request do
       end
 
       let(:casa_case) { volunteer.casa_cases.first }
+      let(:case_number) { "#C-15-JV-24-191" }
+      let(:report_bytes) { Rails.root.join("spec/fixtures/files/default_past_court_date_template.docx").binread }
 
       before do
-        Tempfile.create do |t|
-          casa_case.court_reports.attach(
-            io: File.open(t.path), filename: "#{casa_case.case_number}.docx"
-          )
-        end
+        casa_case.update!(case_number: case_number)
+        casa_case.court_reports.attach(
+          io: StringIO.new(report_bytes), filename: "stored-report.docx"
+        )
       end
 
       it "authorizes action" do
@@ -73,6 +74,15 @@ RSpec.describe "/case_court_reports", type: :request do
 
       it "send response with a status :ok" do
         expect(request).to have_http_status(:ok)
+      end
+
+      it "sends the stored report bytes using the case number as the filename", :aggregate_failures do
+        expect(request.body).to eq(report_bytes)
+        expect(request.body.bytesize).to eq(casa_case.latest_court_report.blob.byte_size)
+        expect(request.body).to start_with("PK\x03\x04".b)
+        expect(request.headers["Content-Disposition"]).to include(
+          %(attachment; filename="#{casa_case.case_number}.docx")
+        )
       end
     end
 


alloutflo:
CI note: the Brakeman failure is unrelated to this diff. Its only warning is `EOLRails` for Rails 8.0.5.1 reaching end of support on 2026-10-07, reported against the unchanged `Gemfile.lock`; there are no scan errors or controller findings. The latest Brakeman run on the same base commit (`8e893b17`) passed on 2026-08-07, so this appears to be time-dependent Brakeman/EOL-data drift rather than a regression in the binary download change. I have kept the PR focused instead of adding an unrelated dependency or ignore-file change.

app/controllers/case_court_reports_controller.rb
```diff
@@ -24,8 +24,11 @@ def show
     respond_to do |format|
       format.docx do
         @casa_case.latest_court_report.open do |file|
-          # TODO test this .read being present, we've broken it twice now
-          send_data File.read(file.path), type: :docx, disposition: "attachment", status: :ok
+          send_data File.binread(file.path),
+            type: :docx,
+            filename: "#{@casa_case.case_number}.docx",
```
compwron, line 29: nit: put filename-ok-formatted timestamp in this (time as well as date)

spec/requests/case_court_reports_spec.rb
```diff
@@ -74,6 +75,15 @@
       it "send response with a status :ok" do
         expect(request).to have_http_status(:ok)
       end
+
+      it "sends the stored report bytes using the case number as the filename", :aggregate_failures do
```
compwron, line 79: nice aggregate_failures :)

compwron (approved):
Beautiful work, thanks! :)

PR #7204, opened 2026-09-20
test: abort with a clear message when built assets are missing

### LINKED ISSUE #7100: Specs rendering the casa_app layout fail locally: tailwind.css was not declared to be precompiled
## What happens

On a fresh local checkout, any spec that renders the `casa_app` layout fails:

```
Sprockets::Rails::Helper::AssetNotPrecompiledError:
  Asset `tailwind.css` was not declared to be precompiled in production.
  Declare links to your assets in `app/assets/config/manifest.js`.

    //= link tailwind.css

  and restart your server
  ./app/views/layouts/casa_app.html.erb:15
```

For example, three examples in `spec/requests/case_court_reports_spec.rb` (the index action):

```
rspec ./spec/requests/case_court_reports_spec.rb:14  # as volunteer can view 'Generate Court Report' page
rspec ./spec/requests/case_court_reports_spec.rb:28  # as a supervisor can view the 'Generate Court Report' page
rspec ./spec/requests/case_court_reports_spec.rb:37  # as a supervisor with no cases in the organization
```

**CI is green** — `.github/workflows/rspec.yml` runs `bundle exec rails assets:precompile` in its Build App step, so `app/assets/builds/tailwind.css` exists there. This is purely a local developer experience problem, but it hits anything rendering a `casa_app` page, which is a growing share of the app as the design migration proceeds.

## Why

`app/assets/builds/tailwind.css` does not exist until someone builds it. Before #7071 ("disable on-demand asset compilation") Sprockets would compile on demand and this worked; now it is a hard failure.

The error message is actively misleading — it tells you to declare the asset in `app/assets/config/manifest.js`, but the manifest is already correct:

```js
//= link_directory ../builds .css
```

`link_directory` only picks up files that exist. The real problem is that nothing has written `tailwind.css` yet.

## Two things worth fixing

**1. Document it.** Neither the README nor `bin/setup` mentions that you must build CSS before running specs. `npm run build` only builds JS — CSS is a separate `npm run build:css`. Easy to miss, and the resulting error points you at the wrong file.

**2. `npm run build:css` may not work.** On my machine it fails outright:

```
$ npm run build:css
> tailwindcss -i ./app/assets/stylesheets/tailwind.css -o ./app/assets/builds/tailwind.css --minify
sh: tailwindcss: command not found
```

`package.json` declares both `tailwindcss` and `@tailwindcss/cli` at `^4.3.2`, but after `npm install`, `node_modules/@tailwindcss/` is empty and there is no `node_modules/.bin/tailwindcss`. In Tailwind v4 the CLI binary comes from `@tailwindcss/cli`, so the script may need to be `npx @tailwindcss/cli` — or this is an install/lockfile problem specific to my machine (arm64 macOS). Flagging it as an observation rather than a diagnosis; someone should check whether a clean `npm ci` produces a working `tailwindcss` binary before changing the script.

## Suggested fix

- Have `bin/setup` build CSS alongside JS, and/or add a line to the README's testing section
- Confirm `npm run build:css` works from a clean `npm ci`, and fix the script if not
- Optionally, make the failure self-explanatory — a clearer message than Sprockets' default would save the next person the detour through `manifest.js`

## Context

Found while debugging #7093 — these three failures were initially mistaken for fallout from that work before being traced to missing built assets.

costajohnt:
Resolves #7100

### What changed, and _why_?

On a fresh checkout, `app/assets/builds/` is empty (it is gitignored and only populated by `npm run build` and `npm run build:css`). `manifest.js` links that directory, and `link_directory` only declares files that exist, so a layout that links `tailwind.css` raises `AssetNotPrecompiledError` with advice to edit the manifest, which is already correct. A missing `application.js` fails with "not present in the asset pipeline".

- `spec/rails_helper.rb`: abort at boot when any of the three bundles the layouts link (`application.js`, `all_casa_admin.js`, `tailwind.css`) is missing, naming the commands that build them. Sits next to the existing pending-migration abort and follows the same pattern.
- `README.md`: note under Running Tests that `bin/rails spec` builds the bundles first (it runs `spec:prepare`, which jsbundling/cssbundling hook into), while a bare `bundle exec rspec` does not.

Two things from the issue that did not hold up:

- #7071 only changed `config/environments/production.rb`; the test environment still has `config.assets.compile = true`. The cause is the missing build output, not on-demand compilation being off.
- `npm run build:css` works from a clean `npm ci` here (arm64 macOS, Node 24.18, npm 11): `@tailwindcss/cli` and the `tailwindcss` bin are installed and the CSS is written. So I left `package.json` alone.

`bin/setup` already runs both builds since #7051, so that part of the issue was done.

No spec: the guard runs while `rails_helper` loads and aborts the process, and the repo has no precedent for subprocess specs. Checked by hand:

- `bundle exec rspec spec/requests/case_court_reports_spec.rb:14:28:37` with the bundles built: 3 examples, 0 failures.
- With `app/assets/builds/tailwind.css` removed, `bundle exec rspec spec/requests/case_court_reports_spec.rb:14` exits 1 and prints `Missing built assets: tailwind.css. Run \`npm run build\` and \`npm run build:css\`, or use \`bin/rails spec\`, which builds them.` before anything else.
- With `tailwind.css` removed, `bin/rails spec SPEC=...` rebuilds it and passes, which is why the README wording distinguishes the two.
- `bundle exec standardrb spec/rails_helper.rb` clean.

CI runs `rails assets:precompile` before rspec, so the guard cannot fire there.

### Screenshots please :)

Terminal output with `tailwind.css` removed:

```
$ bundle exec rspec spec/requests/case_court_reports_spec.rb:14
Missing built assets: tailwind.css. Run `npm run build` and `npm run build:css`, or use `bin/rails spec`, which builds them.
```


compwron (approved):
:)

